PR #332, opened 2021-05-13
Add "extra info" field to the records

shanedsnyder:
In GitLab by @roblatham00 on Feb 21, 2018, 14:45

I did this a while back on intel-hpdd's repo, where it seems to have been forgotten.

I have sometimes needed to annotate log fields with extra information.  for example, in ROMIO, it was helpful to know which round of two-phase collective I/O an operation happened in.

I suspect EXTRA_INFO would be used rarely, so adding 64 bytes to a log entry all the time seems wasteful.  Can switch to a more dynamic approach where EXTRA_INFO allocated only if environment variable set if that seems better

### DIFF
--- a/darshan-runtime/configure
+++ b/darshan-runtime/configure
@@ -4674,10 +4674,10 @@ int
 main ()
 {
 
-    MPI_Comm comm;
-    char* filename;
-    int amode;
-    MPI_Info info;
+    MPI_Comm comm=0;
+    char* filename=0;
+    int amode=0;
+    MPI_Info info=0;
     MPI_File fh;
             MPI_File_open(comm, filename, amode, info, &fh);
   ;
--- a/darshan-runtime/configure.in
+++ b/darshan-runtime/configure.in
@@ -392,10 +392,10 @@ AS_IF([test "x$ENABLE_MPI" = "x1"], [
 # determine if the MPI library includes MPI-IO functions or not
 AC_MSG_CHECKING(for MPI-IO support in MPI)
 AC_TRY_LINK([#include <mpi.h>], [
-    MPI_Comm comm;
-    char* filename;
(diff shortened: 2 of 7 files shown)


shanedsnyder:
In GitLab by @carns on Feb 28, 2018, 13:04

I would vote for a dynamic approach for sure.  64 bytes per access is too much for the general case.

How do you use it exactly?  Do you just do putenv() in the code you are annotating to pass information that will show up in DXT?

In that case, maybe a doc update with the pull request too just so we can remember how to use it when needed :)

shanedsnyder:
In GitLab by @roblatham00 on Apr 20, 2020, 10:38

added 152 commits

<ul><li>657de0d7...7dfbddce - 148 commits from branch <code>master</code></li><li>cf60c400 - Initialize arguments for MPI-IO check</li><li>6bd1cfea - fix a shadowed declaration warning</li><li>f9eab31e - add an &#39;offset&#39; to mpi-io traces</li><li>00304584 - update parser to show MPI-IO offset information</li></ul>

[Compare with previous version](/darshan/darshan/merge_requests/19/diffs?diff_id=8953&start_sha=657de0d7db873245c804525c16713c9c6c635c74)

shanedsnyder:
In GitLab by @roblatham00 on May 1, 2020, 15:16

closed

shanedsnyder:
In GitLab by @roblatham00 on May 1, 2020, 15:17

Uh looks like I force-pushed and overwrote these changes.  I'll have to find the old work and dust it off; might not even still be a good idea.

roblatham00:
Not saying the idea is any good, but it wasn't a ton of work to update to main.

roblatham00:
revisit this work in #979

Issue #247, opened 2021-05-13
API to support injection of characterization data from external tools


shanedsnyder:
In GitLab by @carns on Oct 2, 2018, 11:18

The use case for this would be instrumentation from an I/O library that wants to inject it's data into Darshan so that it travels with the Darshan log and uses it's compression and output infrastructure.

This would most likely mean adding an externally visible wrapper to the module register and record id functions, plus an extra function to retrieve a pointer to the memory region allocated during registration.

shanedsnyder:
In GitLab by @glennklockwood on Oct 2, 2018, 11:47

This would be a great feature to have.  There are a few times where I would have also liked the ability to insert new records/modules (or even arbitrary data, like HDF5's user block) into Darshan logs post-hoc.  While this would probably be dangerous to do in production, it would be a more portable way for downstream analysis to attach indices or derived quantities (e.g., `darshan-parser --perf` counters) to existing logs so they don't have to be recalculated.

I'm not sure if use cases are compatible with the scope of this feature request, but hopefully they are.

shanedsnyder:
In GitLab by @carns on Oct 2, 2018, 12:19

@glennklockwood Would you mind opening a separate ticket for the post-hoc injection idea?  I can see how that would be useful, but it's probably a separate development effort from this because it will need to modify existing logs.  I think it's definitely doable though.

shanedsnyder:
In GitLab by @carns on Oct 4, 2018, 09:31

mentioned in merge request !24

shanedsnyder:
In GitLab by @carns on Apr 25, 2020, 07:17

unassigned @carns

PR #337, opened 2021-05-13
WIP: make API used by modules available externally

shanedsnyder:
In GitLab by @carns on Oct 4, 2018, 09:31

use case: modules that are not compiled directly into Darshan but
would like to register modules and store data there

This merge request splits out the module API into a separate header that can be installed in include path to give external users access.  There is no need for a separate function to get memory pointer; the existing register_record() function should work.

See #247


shanedsnyder:
In GitLab by @carns on May 12, 2021, 15:53

deprecated, we'll reopen later if this becomes a priority

shanedsnyder:
In GitLab by @carns on May 12, 2021, 15:53

closed

PR #381, opened 2021-05-13
use automake and libtool

shanedsnyder:
Re-creating @wkliao's PR from our GitLab repo (https://xgitlab.cels.anl.gov/darshan/darshan/-/merge_requests/77) here to continue to track Darshan's transition to automake/libtool.


carns:
Updated this branch just now with the following changes:

* rebased against current origin/main to reflect updates prior to 3.3.0 release
* reverted back to const feature check for mpi-io wrapper prototypes rather than using MPI_VERSION
* split out apmpi's mpi v3 autoconf test, as it does rely on a complete v3 implementation
* removed old `prepare` script and replaced with an updated `prepare.sh` at the top level

This appears to be complete in basic testing, but we need to try on more platforms, with autoperf, etc. before merging.

Will soon post a WIP spack package revision that supports this build system.

carns:
Also need to check if apmpi options are in alignment between darshan-util and darshan-runtime now.  See discussion on xgitlab.

carns:
Here is a spack fork/branch that supports both this branch (as `@main`) and existing releases:

https://github.com/carns/spack/tree/carns/darshan-github-automake

A few things to resolve:

- [ ] the darshan-util package is installing log format headers with a different naming convention than 3.3.0.  This probably nees to be aligned?
- [ ] make sure that autoperf configure arguments are consistent in util and  runtime
- [ ] need to test optional components, in particular autoperf and hdf modules
- [ ] the darshan-util and darshan-runtime package.py files in the above reference spack fork need to refer to the main branch rather than wkliao/darshan-automake_libtool once this Darshan PR has been merged

wkliao:
Based on the autoconf manual, option `--with-package` is the superset of `--enable-package`.
When 'arg' is not used in both options, they behavior the same, i.e. `--with-hdf5` is the same as `--enable-hdf5`, and `--without-hdf5` is the same as `--disable-hdf5`. Option `--with-package` is for Working With External Software, see Section 15.2 of the manual. To me, having both of the options is redundant. When users set `--with-hdf5` without an argument, the user's intent is clear enough that he/she wants to enable the HDF5 feature using the system installed HDF5.
configure.ac should be revised to search for `h5dump` to obtain the path, to as what you did.

If you kept both options, then you have to deal with the possible conflicted usage.
For example, both `--with-hdf5=/path` and `--disable-hdf5` are used at the command line and vice versa.

shanedsnyder:
Yeah, I agree that the redundancy is a little unfortunate, but think there are other tradeoffs, too. The main reasons for the change are:
1. All other Darshan instrumentation modules can be selectively enabled/disabled at configure time using `--enable-MODULE-mod` options. I mostly just want to be consistent on which mechanisms control whether specific instrumentation modules are built or not. Connecting back to how autoconf describes these options, an enableval seems appropriate for controlling whether a particular internal Darshan module is going to be built and included in the library.
2. Our other modules don't really rely on headers/libraries that aren't guaranteed to be found in the system environment, but the HDF5 module does. Given that we may need an install prefix to find HDF5 installs that aren't generally available in the environment, a withval seems like a convenient way for providing the install prefix, ignoring redundancy concerns.

And just to clarify, users can't use either `--enable-hdf5-mod` or `--with-hdf5` to enable the HDF5 instrumenation module, they are _not_ aliases. Users have to specify they want to use HDF5 with `--enable-hdf5-mod`, or else HDF5 module definitely won't be used. If users only set that option, then Darshan will try to autodetect HDF5 using the path to h5dump. If HDF5 is not in the default software environment, then users can additionally use `--with-hdf5=<dir>` to provide the install prefix. To summarize, `--with-hdf5` has no effect, unless a user has also specified `--enable-hdf5-mod`, and it may not even be necessary depending on the users software environment (i.e., are HDF5 binaries available in default paths, as is the case using Spack).

Does that seem reasonable?

shanedsnyder:
One issue I'm running into on Theta in particular related to finding bz2 libraries for building darshan-util:
```
checking bzlib.h usability... yes
checking bzlib.h presence... yes
checking for bzlib.h... yes
checking for library containing BZ2_bzCompressInit... no
configure: error: libbz2 not found
```
I don't see this issue when building off our current main branch and I notice the autoconf logic for these tests changed. In any case, I don't think an inability to find bz2 should result in an error? My understanding is that it did not fail in prior Darshan versions and I see some comments in the old maint/config/check_bzlib.m4 macro that confirm this, though I don't fully grok yet. We should make sure the new bz2 logic can actually autodetect bz2 installs and should make sure if they aren't found that darshan-util configure doesn't fail.

carns:
Right, bz2 support should be optional.  No darshan logs use bz2 by default; that's just an optional feature in darshan-util (mostly for the darshan-convert utility, like if you are going to bulk archive or anonymize s set of logs) to compress them a little further.

shanedsnyder:
I just pushed some more commits, one of which is a bug fix for how Darshan's ld-opts files are being generated. There are a couple more issues I'm hitting on Theta:
1. darshan-runtime's static library is being compiled with `-DDARSHAN_PRELOAD` flag, but this should only be set when building the shared library. This is resulting in some errors when trying to compile statically-linked applications with Darshan loaded. Looking at the source (https://github.com/darshan-hpc/darshan/blob/wkliao/darshan-automake_libtool/darshan-runtime/lib/Makefile.am#L10), the makefile is setting the `-DDARSHAN_PRELOAD` flag indiscriminately if `ENABLE_LDPRELOAD` is set (which is set by default), but really it should only be set when building the shared library target, not the static library target. I can temporarily workaround just by disabling the build of Darshan's shared library 
2. Shared libraries are not built at all for darshan-runtime, so I can not test them on Theta. Is there anything special you have to do in automake to get it to build static and shared libraries? darshan-util builds both library types fine, but I haven't looked close enough to see differences between the two yet. Ultimately, we'll need to build both libraries by default (though users can disable them explicitly), with each library type requiring slightly different CFLAGS definitions -- you should be able to see the full discrepancies in how the shared libraries are built vs the static libraries in our original Makefile we're using on our main branch (https://github.com/darshan-hpc/darshan/blob/main/darshan-runtime/Makefile.in#L213). I can tinker more with this later, but in the chance someone knows how to pull this off, please feel free to take a stab at it.

Another question I have, not necessarily related to Theta: is there a way we can specify different compilers for darshan-runtime/darshan-util? I think that could be useful for building darshan-runtime and darshan-util all in one configure line, but in cases where the user doesn't want to use the same compiler for each (e.g., darshan-runtime compiler is for compute nodes, but darshan-util is for login nodes with different architecture or something). I think the only way to do this now is to configure darshan-runtime and darshan-util indepednently -- would it make sense to support options like this so users have more control over which compilers are using for each component?

I've got to switch gears for a couple of days, but can have a closer look later this week on these.

wkliao:
> Shared libraries are not built at all for darshan-runtime, so I can not test them on Theta.

Did you try adding "LDFLAGS=-dynamic" when configuring on Theta?

> is there a way we can specify different compilers for darshan-runtime/darshan-util? 

I added "MPICC=mpicc CC=gcc" to my configure command. MPICC is used by darshan-runtime
and CC by darshan-util.

carns:
For a conventional darshan-runtime build on a system like Theta, we need to compile and install both static *and* dynamic libraries, in the same path.  This is necessary for Darshan to transparently instrument applications regardless of whether they are statically or dynamically linked (static happens to be the default on Theta, but both are supported).  The two libraries must be compiled with different flags, though, because the Darshan source includes macros that alter the symbol interception method depending on whether static linking or dynamic linking are used.  There is no one symbol interception method that will work for both cases, unfortunately.

The origin/main makefiles intentionally build every .c file twice for this reason, unless one of the library types is explicitly disabled at configure time.  We need to make automake behave similarly: build the library twice, with different options, once for the static library and once for the dynamic library.

wkliao:
I can see both static and shared libraries built on Theta using the following configure command:
```
% ./configure --prefix=$HOME/Darshan/Github/DEV \
                     --with-log-path=$HOME/Darshan/LOG \
                     --with-jobid-env=NONE \
		     cross_compiling=yes \
                     CC=gcc MPICC=cc

% make install

% /bin/ls -lg ../Github/DEV/lib/
total 4872
-rw-r--r-- 1 users 1053610 Jun 14 18:45 libdarshan.a
-rwxr-xr-x 1 users     982 Jun 14 18:45 libdarshan.la
lrwxrwxrwx 1 users      19 Jun 14 18:45 libdarshan.so -> libdarshan.so.0.0.0
lrwxrwxrwx 1 users      19 Jun 14 18:45 libdarshan.so.0 -> libdarshan.so.0.0.0
-rwxr-xr-x 1 users  638752 Jun 14 18:45 libdarshan.so.0.0.0
-rw-r--r-- 1 users  421282 Jun 14 18:45 libdarshan-util.a
-rwxr-xr-x 1 users    1003 Jun 14 18:45 libdarshan-util.la
lrwxrwxrwx 1 users      24 Jun 14 18:45 libdarshan-util.so -> libdarshan-util.so.0.0.0
lrwxrwxrwx 1 users      24 Jun 14 18:45 libdarshan-util.so.0 -> libdarshan-util.so.0.0.0
-rwxr-xr-x 1 users  350160 Jun 14 18:45 libdarshan-util.so.0.0.0
drwx------ 3 users     512 Jun 14 18:45 Number
drwx------ 2 users     512 Jun 14 18:45 pkgconfig
drwx------ 2 users     512 Jun 14 18:45 Pod
drwx------ 2 users     512 Jun 14 18:45 TeX
```

shanedsnyder:
Thanks for the tip on using `cross_compiling=yes` -- that's the only way I can get configure to decide to build both shared and static libraries on Theta. Our previous method of using `--host=x86-64` doesn't seem to work properly, with configure indicating that cross compiling isn't used and that shared libraries don't work.

Still, I can't seem to find a way to provide specific CFLAGS for Darshan's shared library vs its static library. As indicated previously, we really need all sources for Darshan's .so to be built with the `-DDARSHAN_PRELOAD` compile flag, while we don't want this flag applied to Darshan's .a static library. I've looked over the docs and all kind of examples, and don't see anyway to set CFLAGS for static vs shared libraries. You can obviously do things like

`libdarshan_la_CFLAGS = -DDUMMY`

but that applies to the entire libtool library, which means it applies to both .so and .a libraries. I've confirmed this digging through make output. Things like

`libdarshan_a_CFLAGS = -DDUMMY`

have no effect whatsoever, but that's expected when using libtool, I guess.

Pretty frustrating that there doesn't appear to be a way to do this. Unless you guys have ideas, we may need to avoid using libtool? I think we could get this working without libtool library abstractions getting in our way (i.e., by using libdarshan_a_CFLAGS and libdarshan_so_CFLAGS directly).

wkliao:
Maybe ask suggestions or search for answers on autoconf and libtool mailing lists?
https://lists.gnu.org/mailman/listinfo/autoconf
https://lists.gnu.org/mailman/listinfo/libtool

carns:
I just dug around a little and didn't see an obvious solution to make libtool do this.

Maybe it would be easier to use libtool for dynamic libraries *only*, and use an explicit makefile rule to build the libdarshan.a static library?  The static library is not sophisticated; it's just an archive of all of the objects.  There is less need for libtool in that path anyway.

We could leverage the same SRC variable but maybe have it build to different named targets to prevent collision.  The main difficulty is probably in how to prevent libtool from also attempting to create a .a file on some platforms?  Is there a way to explicitly disable that?

shanedsnyder:
Looks like something like this could work:

```
# Turn off shared libraries during beta-testing, since they
# make the build process take too long.
LT_INIT([disable-static])
```

wkliao:
Also, users can add configure command-line options:
`--enable-shared`, `--disable-shared`, `--enable-static`, `--disable-static`, `--with-pic`, and `--without-pic`.
See https://www.gnu.org/software/libtool/manual/libtool.html#LT_005fINIT

carns:
We might actually want to catch --enable-static and --disable-static (if we can?) or throw an error at the end of the configure script if they have been explicitly specified, since those would make libtool interfere with a manual libdarshan.a target?

shanedsnyder:
Just added a couple more fixes, as well as a commit for using libtool only for shared libraries (1e9151b). This all seems to work on Theta for both static and dynamic linking cases.

carns:
Thanks @shanedsnyder . I just added another commit that adds a safety check to make sure the user doesn't specify `--enable-static` on the darshan-runtime configure command line.  If they do, then configuration will now fail with this error:

```
configure: error: please do not use the --enable-static option to override darshan-runtime library settings.  The libdarshan.a static library will be built automatically without it.
```

Just out of curiosity I also checked what would happen without this check in place if someone specified that option.  It would cause libtool to build another (incorrect) libdarshan.a in the .libs subdir and then install it over the top of the manually specified one.  I don't expect anyone to do this (based on our documentation and/or spack package) but this safety check seems like a good idea to be safe since the result would be confusing.

shanedsnyder:
Note that in commit 8d5d79c, I've modified the behavior of `CC`/`MPICC`/etc. for darshan-runtime, darshan-util, and the top-level.

darshan-runtime is not dependent on an MPI compiler, so I just got rid of all uses of `MPICC` in the code base to avoid confusion. On the runtime side, the following logic is used to determine the compiler to use, `CC`:
* if `CC` is not given and MPI-mode is enabled, use macro to find MPI-capable compiler
* if `CC` is given and MPI-mode is enabled, check that `CC` can compile MPI programs
* if MPI-mode is not used, use given `CC` or determine a compiler to use using AC_PROG_CC

Ultimately, the only way to dictate which compiler darshan-runtime/darshan-util use is via the `CC` variable. To support the ability of top-level configure specifying a different compiler for darshan-runtime and darshan-util, I've added `RUNTIME_CC` and `UTIL_CC` variables. E.g., 

`darshan-top-level/configure <configure_args> --prefix=/path/to/install RUNTIME_CC=mpicc UTIL_CC=gcc`

Distinguishing between a runtime and util compiler is more useful than between MPI and C compiler, I think, just because our runtime code is not MPI dependent.

shanedsnyder:
I'm all done with changes and just rebased on latest master and pushed. I'm ready to merge at this point.

Last remaining thing Phil and I will sort out is details related to Spack packaging, which are outside of this repo. I'll hold off on merging until we've got those sorted out, but shouldn't take us much time at all.

Thanks a ton for this contribution @wkliao, I think everything looks great and think this is going to make our lives a lot easier going forward!

shanedsnyder:
FYI, PR is out for Spack changes related to these changes to Darshan's build system: https://github.com/spack/spack/pull/24906

PR #385, opened 2021-05-14
WIP: Python Job Summary

kevin-harms:
Reimplementation of the darshan-job-summary.pl but using python and the pydarshan framework. 

The command is built in the same method as the pydarshan cli commands. The job summary generates a static HTML file which contains all the data and graphs encoded within the HTML. The HTML will still require access to the internet to pull down some of the dependencies.
  - genshi xml namespace definition
  - bootstrap CSS and javascript
  - jquery javascript

The HTML report is built using Bootstrap 4 and is templated with the Genshi templating engine. The job summary will load a log using the DarshanReport object and then generate various plots using matplotlib. Data manipulation within the job summary is done using Pandas.

Main code is found in: darshan-util/pydarshan/darshan/cli/summary.py
HTML template: darshan-util/pydarshan/darshan/templates/summary.html

Current WIP status:
  - generic layout implemented
  - a few graphs are implemented, others refer to a non-existent static plots.
  - a few tables are implemented, others contain static data as place holders.

How to run:
  - python -m darshan.cli summary --output=report.html log.darshan 

The currently implemented graphs and tables have been tested against existing logs and darshan-job-summary output such that both give identical results (so far).


tylerjereddy:
The `diff` doesn't seem right--shouldn't this branch target `main` rather than the other way around? For example, I see my own recent changes in `main` in the diff here.

carns:
@tylerjereddy you are right; we'll get a new WIP PR opened.

PR #387, opened 2021-05-18
WIP: wrapper performance optimizations

carns:
This PR will track potential wrapper changes to reduce overhead for low-latency I/O operations.

Includes a new benchmark (stdio-benchmark) that can be executed sequentially on tmpfs to measure how much overhead the Darshan wrappers introduce for a set of stdio operations (`fprintf`, `fscanf`, and `fread`) that don't consume much time of their won in this scenario.

Will probably start by focusing on any costs incurred when Darshan is disabled at runtime (via `export DARSHAN_DISABLE=1`).


carns:
PR now includes rough draft of experimental dlopen() for auxiliary library containing clock_gettime() wrappers.  This portion still needs the following:

- [ ] timing around dlopen so we can see it's performance at scale
- [ ] compile-time library path (the first cut relies on rpath on libdarshan.so; that won't work for libdarshan.a)
- [ ] test and see if this works on Theta with static linking, and if so what the performance implications are

carns:
This branch now includes an `--enable-rdtscp=<base_clock_rate>` configure option to enable `rdtscp` timing in place of `clock_gettime()`.  On Theta (with the option set to 1300000000) this is the benchmark impact on an adversarial stdio workload to tmpfs for a statically linked executable:

```
## NO DARSHAN static:
50000000 fprintf()s of size 1 each in 12.465365 seconds (4011113.939493 ops/s)
50000000 fscanf()s of size 1 each in 29.004269 seconds (1723884.156082 ops/s)
50000000 fread()s of size 1 each in 3.130983 seconds (15969424.993880 ops/s)

## darshan@main, enabled, static:
50000000 fprintf()s of size 1 each in 215.936336 seconds (231549.728577 ops/s)
50000000 fscanf()s of size 1 each in 263.471802 seconds (189773.628985 ops/s)
50000000 fread()s of size 1 each in 200.895207 seconds (248885.977447 ops/s)

## darshan@perf-opt, enabled, static:
50000000 fprintf()s of size 1 each in 27.554754 seconds (1814568.911264 ops/s)
50000000 fscanf()s of size 1 each in 52.301563 seconds (955994.373946 ops/s)
50000000 fread()s of size 1 each in 18.619044 seconds (2685422.507416 ops/s)
```
Darshan still adds overhead for this workload, but it is significantly lower than it was before.

carns:
The last step to wrap up this branch is to carry through the macro changes to the other modules besides `stdio`.  The macros have been altered to avoid locking overhead when possible, but they require a slight adjustment to their usage so that they can leverage a local variable in each wrapper.  Right now this has only been done for the stdio module.

Issue #384, opened 2021-05-13
MAINT: matplotlib module should be renamed


tylerjereddy:
It is not good to have a module named `matplotlib.py` in our library--it can interfere with the import of the popular third-party plotting library of the same name when developing code.

For example, I'm having to do this locally for some work:

`darshan/experimental/plots/matplotlib.py -> darshan/experimental/plots/matplotlib_mod.py`

when writing modules in that path that leverage `matplotlib` (the external package).

nawtrey:
Looks like this was fixed via gh-404. Closing.

Issue #307, opened 2021-05-13
analysis utility to generate heat maps from Darshan DXT data


shanedsnyder:
In GitLab by @carns on Mar 8, 2021, 14:05

The heat map would show, per process, how I/O intensity is distributed across time and ranks.  Would be helpful for getting an initial intuitive feel for hot spots and burstiness.  Could also be used to calculate throughput over time intervals, and could be annotated with metadata access intensity and file names.

Ideally this would be a Python tool that utilizes the Darshan Python bindings to produce a 2D array (ranks * nbins), where nbins is fixed regardless of the runtime, and access that span time bins are proportionally divided across them.

If we could come up with a reasonably clear/concise graph it could be one of the lead figures in a revised darshan-job-summary.

carns:
Closing this issue; this has been covered by work in the the pydarshan-devel branch and the pydarshan-job-summary-v1 milestone.

PR #388, opened 2021-05-18
Pydarshan: Break up plots into designated source files, to resolve matplotlib.py issue; Add an DXT Heatmap plot variant; 

jakobluettgau:
#384 #307


tylerjereddy:
@nawtrey May want to inspect this PR to look for overlap in the effort to generate one of the report plots requested by @carns.

nawtrey:
Unfortunately there is overlap here. I was working on #388, I didn't know it had been scooped up

tylerjereddy:
I guess that makes you ideal for providing a detailed review and working together to get it just right according to what the team wants.

jakobluettgau:
No worries, feel free change or update the plot with whatever you have ;) I was just pushing this as blueprint I had for that sort of  plot which was lingering in another branch.

nawtrey:
Sounds good! I'm going to open a PR and dump my examples there. Currently I have a function for generating a heatmap with vertical and horizontal histograms (sum across ranks, sum over time) that I've been using random `numpy` data to iterate with. 

Is there a `darshan` log with interesting IO performance (and multiple ranks) that you've used for testing your heatmap outputs? If so, I think it would be helpful to post a comment with that log (or the file name if its in the test suite) along with the heatmap output from it. I could then generate an output for the same log, and we can start drawing comparisons to figure out what we like/dislike. I think we will need @carns input as well, since this is his brainchild.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+import numpy as np
+
+def plot_dxt_heatmap(report, 
+        xbins=10, ybins=None, 
+        group_by="rank", 
+        mods=None, ops=None, 
+        display_values=False, cmap=None, figsize=None, 
+        amplify=False):
+    """
+    Generates a heatmap plot from a report with DXT traces.
```
tylerjereddy, line 17: @nawtrey How does your DXT heatmap compare with the result below using this feature branch? The test case, available in the repo test suite, is so simple that it should provide a fairly obvious comparison point.

```python
import darshan
import darshan.backend.cffi_backend as backend
from darshan.experimental.plots.plot_dxt_heatmap import plot_dxt_heatmap

# use the super-simple DXT trace currently in the pydarshan
# test suite as a first pass on DXT heat map plot appearance
logfile = '/home/tyler/github_projects/darshan/darshan-util/pydarshan/tests/input/sample-dxt-simple.darshan'
report = darshan.DarshanReport(logfile, read_all=True)

fig = plot_dxt_heatmap(report=report,
                 mods=['DXT_POSIX'])

fig.savefig('test_jakob_dxt_heatmap.png', dpi=300)
```

test_jakob_dxt_heatmap

darshan-util/pydarshan/darshan/experimental/plots/plot_access_histogram.py
```diff
@@ -0,0 +1,100 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+def plot_access_histogram(report, mod):
+    """
+    Plots a histogram of access sizes for specified module.
+
+	Args:
+		report (darshan.DarshanReport): report to gneerate plot from
```
tylerjereddy, line 15: Typo `gneerate`

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+                    
+                for event in rec[f'{op}_segments']:
+                    xbin = int((event['start_time'] / runtime) * xbins)
+                    events[ybin][xbin] += 1
+                    if amplify:
+                        rng = amplify
+                        for i in range(rng):
+                            sur = int(rng/2) + i
+                            if sur != 0 and (ybin - sur) > 0:
+                                events[ybin - sur][xbin] += 1
```
tylerjereddy, line 97: I'd probably try to nudge us in the direction of moving processing/number crunching code to functions under test, along with some i.e., DXT files we can use to guard against regressions, etc.

May also be helpful to have a function that describes what the looping is doing (input/expected output, etc.).

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
tylerjereddy: After you do that comparison, here's a more complex DXT-enabled log file (
[dxt-log2](https://github.com/darshan-hpc/darshan/files/6587813/treddy_mpi-io-test_id4373053_6-2-60198-9815401321915095332_1.darshan.gz)
; it has been gzipped, so unzip it first) you can use to compare with Jakob, this time with the plot output below:

test_jakob_dxt_heatmap

tylerjereddy: And 
[an even larger/more complex gzipped/DXT-enabled log file ](https://github.com/darshan-hpc/darshan/files/6587863/treddy_mpi-io-test_id4373096_6-2-61446-92088258725910398_2.darshan.gz) that produced the plot below:

test_jakob_dxt_heatmap

tylerjereddy: Pushing up to 640 MPI ranks with this [gzipped log file](https://github.com/darshan-hpc/darshan/files/6588042/treddy_mpi-io-test_id17541138_6-2-65169-14537227514913022501_15.darshan.gz)
should be useful to compare the dynamic sizing capabilities of the two approaches:

test_jakob_dxt_heatmap

tylerjereddy: And with `2240` MPI ranks on an `aarch64` cluster with this [gzipped log file](https://github.com/darshan-hpc/darshan/files/6593805/treddy_mpi-io-test_id545458_6-3-43062-1247823192538596076_13.darshan.gz):

test_jakob_dxt_heatmap

tylerjereddy: One thing that might be more intuitive is to use the actual time in seconds for the `x` tick labels, instead of the physical bin numbers.

For example,

```diff
@@ -119,6 +131,11 @@ def plot_dxt_heatmap(report,
     ax.set_title(f"DXT Heatmap mods={mods}, ops={ops}")
     ax.set_ylabel(ylabel)
     ax.set_xlabel(f"Time (binsize={runtime/xbins} seconds)")
+    binsize = runtime/xbins
+    xticks = map(str, ax.get_xticks() * binsize)
+    ax.set_xticklabels(xticks)
```

image

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+    events = np.zeros((ybins, xbins))    
+    
+    for mod in mods:
+        for op in ops:
+            for rec in report.records[mod]:
+                
+                if group_by in ['hostname', 'node']:
+                    ybin = hostnames[rec['hostname']]
+                else:
+                    ybin = int((rec['rank'] /ranks) * ybins)
```
tylerjereddy, line 86: For log files with a few thousand ranks you could save a few thousand unnecessary multiplications with this as far as I can tell. I have few more thoughts on this nested for loop I'll add in a separate comment.

```diff
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
@@ -76,14 +76,15 @@ def plot_dxt_heatmap(report,
     # heatmap to be populated with event counts
     events = np.zeros((ybins, xbins))    
     
+    if group_by == "rank":
+        ybin_denom = ranks / ybins
     for mod in mods:
         for op in ops:
             for rec in report.records[mod]:
-                
                 if group_by in ['hostname', 'node']:
                     ybin = hostnames[rec['hostname']]
                 else:
-                    ybin = int((rec['rank'] /ranks) * ybins)
+                    ybin = int(rec['rank'] / ybin_denom)

```

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,124 @@
...
+                    
+                    
+                for event in rec[f'{op}_segments']:
```
tylerjereddy, line 89: Here we are already nested 4 levels deep in a `for` loop, iterating over a `list` of dictionary objects representing individual read or write events. We're perhaps somewhat hamstrung by the generality of `DarshanRecordCollection`, which as its comments suggest has to handle a diverse set of nested data structures for all the instrumentation modules.

Even with more than 2k ranks performance doesn't really seem to be an issue here in my hands though. So, I guess I'm just saying that it is perhaps a bit unfortunate that we can't leverage i.e., a specialized `pandas` dataframe object for each type of darshan instrumentation module, but I'm also not sure it matters too much for the present task.

It is true that we could use `to_df()`, but we're ultimately still dealing with a collection of `DarshanRecordCollection` values which happen to have some internal dataframes in that case, which still leaves us with awkward pure-Python looping instead of vectorized ops in `pandas` or a similar framework.

tylerjereddy: Longer-term, I'm not sure if we need to think about specializing/structuring the exposed DXT data some other way. One challenge might be that the data is (potetenially) "ragged" to some extent as I understand it. So you might have a dataframe a bit like this if you went that route:

```
rank, write_count, write_segment_1_start_time, write_segment_2_start_time, ...
0           1                  0.1                             NaN
1           2                  0.2                             0.4
```

Where you might end up with many rows/ranks filled with `NaN`s, or decide to store a collection in the dataframe, which may bring you back to the same place of nested data structures again.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
jakobluettgau: I agree very much. The pre-processing for plots should be as much as possible decoupled to allow custom analysis on top of the aggregations. 

For this particular snippet, it's really is just a quick (well actually slow) workaround to amplify visibility, e.g., when dealing with high rank numbers. There is certainly better ways to handle these.   Here I would not break it aways from the plot to much because this is purely for visuals, because the matrix is not holding true values for neighboring ranks anymore after applying it.  Much room for different optimizations for sure.  Allowing to switch log scaling for the intensities/transmitted bytes might be one.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
jakobluettgau: I as well would update the DXT representation because currently it very closely mirrors the raw representation in the log. As there is not really a lot of experience on a "most common mode of consumption" that seemed to be sensible for a start, but I agree about it being awkward to access and handle as is.
I find myself often needing read and write activity combined in a sorted list by start_time, for example. 

So actually I am also very much in favor of having all tabular style record types using pandas data frames or xarray internally and rely on all the community optimizations and user familiarity of their APIs in all sorts of analysis use cases.
I did some performance testing with converting everything to pandas on import in the CFFI layer: Currently this does not perform unfortunately, but that can be solved by adding a log_get_n_records on the C side ;)   Another related thing we wanted to address in combination is looking at ways to allow seeking across modules/(maybe records), which due to the sequence of records being compressed requires a little oversight to not break the existing utils.

I think introducing specialized record collections is a good idea, and I do think the current RecordCollection is exactly in that spirit, just that it tries to find some common functionality that helps with the export to_json/to_dict/do_df or merging/filtering when dealing with multiple reports. In principle I would image to_df in many cases to become a noop that just returns the internal pandas data frame. 

From my perspective this is a trade-off in any case that requires to decide how much one would like to preserve Darshans conceptional and raw data model of breaking multi-level instrumentation into <darshan-log>/<module/layer>/records[].

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
shanedsnyder: Yeah, I think it's helpful to retain the exact application timing, rather than in terms of bins, so the duration of I/O activity is more clear (i.e., "there was heavy I/O activity for 15 seconds" vs "there was heavy I/O activity for 4 bins and each bin is 3.5 seconds").

darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
```diff
@@ -0,0 +1,157 @@
...
+    mpicol_vals = [0, 0, 0, 0, 0, 0, 0]
+    stdio_vals = [0, 0, 0, 0, 0, 0, 0]
+
+
+    # TODO: change to report.summary
+    if 'agg_ioops' in dir(report):
+        print("Summarizing... agg_ioops")
+        report.agg_ioops()
+    else:
+        print("Can not create summary, agg_ioops aggregator is not registered with the report clase.")
```
tylerjereddy, line 31: `clase` -> `class`, I think

darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+    mods = report.summary['agg_ioops']
```
tylerjereddy, line 34: I haven't looked at the `agg_ioops` stuff in detail, but if the team prefers to use what is there I don't really mind. I do note that at the moment, this errors out for me even with a simple file and I'm not entirely sure if `DXT` handling is available for the control flow in this function:

```python
import darshan
from darshan.experimental.plots.plot_opcounts import plot_opcounts

logfile = '/home/tyler/github_projects/darshan/darshan-util/pydarshan/tests/input/sample-dxt-simple.darshan'
report = darshan.DarshanReport(logfile, read_all=True)
ret = plot_opcounts(report=report)
```

```python
  File "test_jakob_opcounts.py", line 6, in <module>
    ret = plot_opcounts(report=report)
  File "/home/tyler/python_37_venv/lib/python3.7/site-packages/darshan-3.3.0.3-py3.7.egg/darshan/experimental/plots/plot_opcounts.py", line 34, in plot_opcounts
    mods = report.summary['agg_ioops']
KeyError: 'agg_ioops'
```

Regardless, we'll probably want a clear error message--at the moment something gets printed, but not sure what it means.

`Can not create summary, agg_ioops aggregator is not registered with the report clase.`

For example, if it is having a problem specifically with `DXT` it should say so. If it is something else that is wrong/missing, a suggestion might be helpful for the developer/end user.

darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
```diff
@@ -0,0 +1,157 @@
...
+    autolabel(rects2)
+    autolabel(rects3)
+    autolabel(rects4)
+
+    fig.tight_layout()
+
+    #plt.show()
+    
+
+    return plt
```
tylerjereddy, line 153: As far as I can tell, the `plt` variable here  has no actual association with any of the figure/axis objects in the function, so I believe you're currently just returning `matplotlib.pyplot` generically.

The plots we're targeting here are pretty simple I think. I had started doing some work on my fork feature branch `treddy_data_access_by_filesystem` and the results were a bit like the one below, though I was never really happy with how it was working.

sample-dxt-simple darshan_data_access_by_category

I guess I'll have to dig into the aggregator code you've written if that is what Shane et al. want to use. Once it is clear to me that this function is actually working, we could check the output against what Phil was envisioning.

darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+    x = np.arange(len(labels))  # the label locations
+    width = 0.15  # the width of the bars
+
+    fig, ax = plt.subplots()
+    rects1 = ax.bar(x - width/2 - width, posix_vals, width, label='POSIX')
+    rects2 = ax.bar(x - width/2, mpiind_vals, width, label='MPI-IO Indep.')
+    rects3 = ax.bar(x + width/2, mpicol_vals, width, label='MPI-IO Coll.')
+    rects4 = ax.bar(x + width/2 + width, stdio_vals, width, label='STDIO')
```
tylerjereddy, line 120: My impression from early discussions with Shane is that we may want to focus on a single module to start with for the report--perhaps `DXT_POSIX`. Anyway, establishing which module we want to focus on first probably makes sense.

I haven't checked if your aggregator code has handling to prevent combining duplicates between i.e., `POSIX` and `DXT_POSIX`; that kind of stuff was starting to get tricky to think about.

shanedsnyder: We should probably discuss this in more detail to make sure we're all on the same page. I think we want to use DXT data for the heatmap plots in particular, since it provides finer-grained details that make heatmaps both more informative and easier to develop. For the other plots that we typically provide on darshan-job-summary, DXT is probably not what we want to use as a starting point...plots like the operation count plots, average I/O time per-process, etc. probably should be prototyped using our traditional instrumentation modules that provide summaries (POSIX makes most sense to start with since almost all logs have some POSIX data), rather than using fine-grained DXT traces.

tylerjereddy: I've been using POSIX yeah. Direct access to pandas dataframe data is a bit more natural there if you don't need DXT anyway.

darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
jakobluettgau: This should be moved to https://github.com/darshan-hpc/darshan/pull/397

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
jakobluettgau: I think the discussion of these has since been continued in https://github.com/darshan-hpc/darshan/pull/396

darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
jakobluettgau: The report.summary namespace is something I am using in some of the interactive analysis tools to cache already performed aggregations. We could think about breaking these out to somehwere else, but as far this pull request goes, it just takes the status quo and splits it into multiple files. So I think this should be addressed seperately.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
jakobluettgau: Given that there is the other DXT heatmap I marked this as resolved :)

tylerjereddy:
It is a bit easier for me to review PRs that focus on a single "thing--" in this case an enhancement to add DXT heatmap capability is getting squeezed in along with a module naming/organization bug fix, so it just makes it a bit harder for me to deal with.

In this case, I'll try to deal with it myself, but moving forward things will probably get reviewed/merged much faster if they are surgically focused, with sample plots in the PR description, and an indication of how it helps get us closer to the sketch below, etc.

For example, it is unclear to me how stuff like changing the notebook metadata in `darshan-util/pydarshan/examples/01_darshan-apmpi.ipynb` to a different Python version is really pertinent to the objective of the `pydarshan-devel` feature branch, which is really just to deliver on the script to generate output like below.

Some of the other adjustments here might be better suited to other feature branches if they are aiming to accomplish something else, but I'll try to just merge it in today if I can, in this case.

darshan-summary-straw-man-v2-2021-05-19

image

image

tylerjereddy:
rebased on latest `pydarshan-devel`; checking over a bit more now

tylerjereddy:
Alright, the tests pass locally on this feature branch, so if this helps the experimental notebooks and so on I don't mind merging. If there's anything here in particular you don't want to break moving forward, probably good to add a test for it.

tylerjereddy:
thanks Jakob

PR #389, opened 2021-05-20
add fork support in Darshan's non-MPI mode

shanedsnyder:
This PR contains a number of changes that allow Darshan to properly instrument forked executables. Currently, Darshan does not handle forked processes well, with Darshan log write failures occurring due to log file names not being guaranteed unique among parents/children. Further, fork children inherit all instrumentation state from parents, leading to duplicated instrumentation counters.

To address this, this PR modifies Darshan's log file name generation to guarantee unique file names for parent/children, and uses the `pthread_atfork()` interface for defining handlers that run on fork children that reset all Darshan instrumentation state. 

Detailed changes:
- modified Darshan's log file name generation to include the pid of the writer to guarantee uniqueness, with this pid value appended behind the job ID -- in non-MPI mode, the job ID used in the log file name is the PID of the original parent process for the executable if no job ID env variable is given
- split Darshan's shutdown into separate `output()` and `cleanup()` routines, simplifying the process of resetting modules instrumentation state
- defined `pthread_atfork()` handler for fork children that resets all state for Darshan core and individual instrumentation modules
- added a metadata field to the Darshan job info indicating the parent pid for fork children

WIP tag while we review changes and while we wait to update AutoPerf instrumentation modules to support new changes.


shanedsnyder:
AutoPerf submodule has been updated, and all appears to work fine in some small-scale testing. Merging.

PR #394, opened 2021-06-08
fix undefined var if POSIX module not active

carns:
Reported by Jeff Layton (see https://lists.mcs.anl.gov/pipermail/darshan-users/2021-June/000693.html).  There was a bug in logic added to darshan-job-summary.pl to filter out negative mmap counter values; it did not account for situations in which the counter was not present at all (i.e., the POSIX module was not activated).


carns:
@laytonjbgmail

shanedsnyder (approved):

PR #395, opened 2021-06-08
pydarshan: Include APMPI stats in the pydarshan html report 

sudheer9:
- update the python report generation script to include APMPI data
- update the summary html file to include 3 more tables with APMPI data
- the branch is rebase with the latest darshan main


darshan-runtime/darshan-core.h
```diff
@@ -120,10 +120,11 @@ struct darshan_core_runtime
 #ifdef HAVE_MPI
     MPI_Comm mpi_comm;
 #endif
+    int pid;
 };
 
 void darshan_core_initialize(int argc, char **argv);
-void darshan_core_shutdown(void);
+void darshan_core_shutdown(int write_log);
```
tylerjereddy, line 127: Looks like rebasing has added a bunch of extra C file changes into the diff. Since this is just a PR to the new job summary feature branch PR, it might be better to just push the Python relevant changes there and take a look at some of the review comments I've added there (you can rebase there as needed too--I think that branch is targeting `main` but now there's a desire to have it target `pydarshan-devel` for Shane's workflow I think)

tylerjereddy (commented):
The diff is a bit of a mess here so I'll take a closer look when things are cleaned up.

darshan-runtime/darshan-core.h
shanedsnyder: Yeah, @sudheer9 I'd recommend that as well. In general, I think PRs are mostly necessary for trying to get things merged into `main` or into `pydarshan-devel`. In this case, it's probably easier to make the changes directly to Kevin's existing WIP branch (`python-job-summary`), which is already a PR to our `pydarshan-devel` branch.

I haven't had a close look at the diff either, mostly because it's got a bunch of unrelated commits from me that are cluttering it. Will probably be easier for us to digest by avoiding any rebase onto master and committing directly to Kevin's branch.

sudheer9:
Thanks @tylerjereddy @shanedsnyder, agree with the suggestions, I will add a new PR with `pydarhsan` specific updates.

PR #386, opened 2021-05-17
WIP: Python job summary

carns:
(code contributed by @kevin-harms )

Reimplementation of the darshan-job-summary.pl but using python and the pydarshan framework.

The command is built in the same method as the pydarshan cli commands. The job summary generates a static HTML file which contains all the data and graphs encoded within the HTML. The HTML will still require access to the internet to pull down some of the dependencies.

* genshi xml namespace definition
* bootstrap CSS and javascript
* jquery javascript

The HTML report is built using Bootstrap 4 and is templated with the Genshi templating engine. The job summary will load a log using the DarshanReport object and then generate various plots using matplotlib. Data manipulation within the job summary is done using Pandas.

Main code is found in: darshan-util/pydarshan/darshan/cli/summary.py
HTML template: darshan-util/pydarshan/darshan/templates/summary.html

Current WIP status:

* generic layout implemented
* a few graphs are implemented, others refer to a non-existent static plots.
* a few tables are implemented, others contain static data as place holders.

How to run:

* python -m darshan.cli summary --output=report.html log.darshan

The currently implemented graphs and tables have been tested against existing logs and darshan-job-summary output such that both give identical results (so far).

PR #396, opened 2021-06-09
WIP: PyDarshan: Adds heatmap plotting code for DXT tracing logs

nawtrey:
DXT heatmap plotting code for potential use in updated job summary. Primarily uses `numpy`, `seaborn`, `matplotlib` for plotting, which are all pip installable.  

Summary
---------
At the moment it takes in logs with `DXT_POSIX` data, concatenates all of the read/write segment dataframes together,  then uses that data (along with the respective rank data) to create a 2D histogram, and plot a heat map with marginal bar graphs. Depending on which operations you include, it will plot only the `read` segment data, the `write` segment data, or both. It does not distinguish between operations. Each bin (in the heatmap) is populated with the data sum and/or proportionate data sum for all IO events read/written during the time spanned by the bin.The color bar was selected based on the fact that it is non-white at the minimum value so areas without data are visible.  

How to run
-----------
See: https://github.com/darshan-hpc/darshan/pull/396#issuecomment-879202907

TODO
------
- [x] Fix average time plotting -- need to span from start time to end time for accurate IO representation 
- [ ] ~~Find a way to aggregate data from other modules, if applicable~~
- [ ] ~~Differentiate between read/write data in heat map~~
- [x] Include number of bins in plot somewhere so you can get an idea of bandwidth (i.e. 100 bins w/ 100 s run time = 1 s/bin)
- [ ] Figure out if/how to utilize the figure title. At the moment it just states which modules are being used (may want to remove it entirely for the job summary)
- [x] Finish documentation/comments

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -5,6 +5,7 @@ from typing import Dict, Any, Tuple, Sequence, TypedDict
 
 import pandas as pd
 import numpy as np
+import numpy.typing as npt
 
 
 class SegDict(TypedDict):
@@ -29,7 +30,7 @@ def get_rd_wr_dfs(dict_list: Sequence[SegDict]) -> Tuple[pd.DataFrame, pd.DataFr
     Parameters
     ----------
 
-    dict_list: a list of DXT records, where each record is a
+    dict_list: a sequence of DXT records, where each record is a
     Python dictionary with the following keys: 'id', 'rank',
     'hostname', 'write_count', 'read_count', 'write_segments',
     and 'read_segments'. The read/write data is stored in
@@ -159,10 +160,10 @@ def get_single_df_dict(
 
     report: a ``darshan.DarshanReport``.
 
-    mods: a list of keys designating which Darshan modules to use for
+    mods: a sequence of keys designating which Darshan modules to use for
     data aggregation. Default is ``["DXT_POSIX"]``.
 
-    ops: a list of keys designating which Darshan operations to use for
+    ops: a sequence of keys designating which Darshan operations to use for
     data aggregation. Default is ``["read", "write"]``.
 
     Returns
@@ -228,10 +229,10 @@ def get_aggregate_data(
 
     report: a ``darshan.DarshanReport``.
 
-    mods: a list of keys designating which Darshan modules to use for
+    mods: a sequence of keys designating which Darshan modules to use for
     data aggregation. Default is ``["DXT_POSIX"]``.
 
-    ops: a list of keys designating which Darshan operations to use for
+    ops: a sequence of keys designating which Darshan operations to use for
     data aggregation. Default is ``["read", "write"]``.
 
     Returns
@@ -275,7 +276,7 @@ def get_aggregate_data(
         # if there are dataframes in the list, concatenate them into 1 dataframe
         agg_df = pd.concat(df_list, ignore_index=True)
     else:
-        raise ValueError(f"No data available for selected module(s) and operation(s).")
+        raise ValueError("No data available for selected module(s) and operation(s).")
 
     return agg_df
 
@@ -283,8 +284,8 @@ def get_aggregate_data(
 def calc_prop_data_sum(
     tmin: float,
     tmax: float,
-    total_elapsed: np.ndarray,
-    total_data: np.ndarray,
+    total_elapsed: npt.NDArray[np.float64],
+    total_data: npt.NDArray[np.float64],
 ) -> float:
     """
     Calculates the proportion of data read/written in the
@@ -323,7 +324,7 @@ def calc_prop_data_sum(
     return prop_data_sum
 
 
-def get_heatmap_data(agg_df: pd.DataFrame, xbins: int) -> np.ndarray:
+def get_heatmap_data(agg_df: pd.DataFrame, xbins: int) -> npt.NDArray[np.float64]:
     """
     Builds an array similar to a 2D-histogram, where the y data is the unique
     ranks and the x data is time. Each bin is populated with the data sum
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -283,50 +283,7 @@ def get_aggregate_data(
     return agg_df
 
 
-def calc_prop_data_sum(
-    tmin: float,
-    tmax: float,
-    total_elapsed: npt.NDArray[np.float64],
-    total_data: npt.NDArray[np.float64],
-) -> float:
-    """
-    Calculates the proportion of data read/written in the
-    time interval of a single bin in ``get_heatmap_data``.
-
-    Parameters
-    ----------
-
-    tmin: the lower bound of the time interval for a given bin.
-
-    tmax: the upper bound of the time interval for a given bin.
-
-    total_elapsed: an array of the elapsed times for every event
-    that occurred within the time interval of a given bin.
-
-    total_data: an array of the data totals for every event that
-    occurred within the time interval of a given bin.
-
-    Returns
-    -------
-
-    prop_data_sum: the amount of data read/written in the time
-    interval of a given bin.
-
-    """
-    # calculate the elapsed time
-    partial_elapsed = tmax - tmin
-    # calculate the ratio of the elapsed time
-    # to the total read/write event time
-    proportionate_time = partial_elapsed / total_elapsed
-    # calculate the amount of data read/written in the elapsed
-    # time (assuming a constant read/write rate)
-    proportionate_data = proportionate_time * total_data
-    # sum the data
-    prop_data_sum = proportionate_data.sum()
-    return prop_data_sum
-
-
-def get_heatmap_data(agg_df: pd.DataFrame, xbins: int) -> np.ndarray:
+def get_heatmap_df(agg_df: pd.DataFrame, xbins: int) -> pd.DataFrame:
     """
     Builds an array similar to a 2D-histogram, where the y data is the unique
     ranks and the x data is time. Each bin is populated with the data sum
@@ -344,10 +301,21 @@ def get_heatmap_data(agg_df: pd.DataFrame, xbins: int) -> np.ndarray:
     Returns
     -------
 
-    hmap_data: ``NxM`` array, where ``N`` is the number of unique ranks
-    and ``M`` is the number of x-axis bins. Each element contains the
-    data read/written by the corresponding rank in the x-axis bin time
-    interval.
+    hmap_df: dataframe with time intervals for columns and rank
+    index (0, 1, etc.) for rows, where each element contains the data
+    read/written by the corresponding rank in the given time interval.
+
+    Examples
+    --------
+    The first column/bin for the `hmap_df` generated from
+    "examples/example-logs/ior_hdf5_example.darshan":
+
+               (0.0, 0.09552296002705891]
+        rank
+        0                   8.951484e+05
+        1                   3.746313e+05
+        2                   6.350999e+05
+        3                   1.048576e+06
 
     """
     # generate the bin edges by generating an array of length n_bins+1, then
@@ -420,5 +388,5 @@ def get_heatmap_data(agg_df: pd.DataFrame, xbins: int) -> np.ndarray:
     # the bytes data
     cats = cats.mul(agg_df["length"], axis=0)
     cats["rank"] = agg_df["rank"]
-    grouped_res = cats.groupby("rank").sum()
-    return grouped_res.to_numpy()
+    hmap_df = cats.groupby("rank").sum()
+    return hmap_df
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -1,0 +1,442 @@
+"""
+Module of data pre-processing functions for constructing the heatmap figure.
+"""
+from typing import Dict, Any, Tuple, Sequence, TypedDict
+
+import pandas as pd
+import numpy as np
+
+
+class SegDict(TypedDict):
+    """
+    Custom type hint class for `dict_list` argument in `get_rd_wr_dfs()`.
+    """
+
+    id: int
+    rank: int
+    hostname: str
+    write_count: int
+    read_count: int
+    write_segments: pd.DataFrame
+    read_segments: pd.DataFrame
+
+
+def get_rd_wr_dfs(dict_list: Sequence[SegDict]) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    """
+    Uses the DXT records to construct individual
+    dataframes for both read and write segments.
+
+    Parameters
+    ----------
+
+    dict_list: a list of DXT records, where each record is a
+    Python dictionary with the following keys: 'id', 'rank',
+    'hostname', 'write_count', 'read_count', 'write_segments',
+    and 'read_segments'. The read/write data is stored in
+    ``read_segments`` and ``write_segments``, where each is a
+    ```pd.DataFrame`` containing the following data (columns):
+    'offset', 'length', 'start_time', 'end_time'.
+
+    Returns
+    -------
+
+    Tuple of form ``(read_df, write_df)``,
+    where each tuple element is a ``pd.DataFrame`` object
+    containing all of the read and write events.
+
+    Notes
+    -----
+
+    Used in ``get_single_df_dict()``.
+
+    Examples
+    --------
+
+    ``dict_list`` and ``(read_df, write_df)``
+    generated from ``tests/input/sample-dxt-simple.darshan``:
+
+        dict_list = [
+            {
+                'id': 14388265063268455899,
+                'rank': 0,
+                'hostname': 'sn176.localdomain',
+                'write_count': 1,
+                'read_count': 0,
+                'write_segments':
+                    offset  length  start_time  end_time
+                    0       0      40    0.103379  0.103388,
+                'read_segments':
+                    Empty DataFrame
+                    Columns: []
+                    Index: []
+            },
+            {
+                'id': 9457796068806373448,
+                'rank': 0,
+                'hostname': 'sn176.localdomain',
+                'write_count': 1,
+                'read_count': 0,
+                'write_segments':
+                    offset  length  start_time  end_time
+                    0       0    4000    0.104217  0.104231,
+                'read_segments':
+                    Empty DataFrame
+                    Columns: []
+                    Index: []
+            },
+        ]
+
+        (read_df, write_df) = (
+            Empty DataFrame
+            Columns: []
+            Index: [],
+            length  start_time  end_time  rank
+            0      40    0.103379  0.103388     0
+            1    4000    0.104217  0.104231     0
+        )
+
+    """
+    # columns to drop when accumulating the dataframes.
+    # Currently "offset" data is not utilized
+    drop_columns = ["offset"]
+    # create empty arrays to store
+    # read/write segment dataframes
+    read_df_list = []
+    write_df_list = []
+    # iterate over all records/dictionaries
+    # to pull out the dataframes
+    for _dict in dict_list:
+        # collect the read and write segment dataframes
+        rd_seg_df = _dict["read_segments"]
+        wr_seg_df = _dict["write_segments"]
+
+        if rd_seg_df.size:
+            # drop unused columns from the dataframe
+            rd_seg_df = rd_seg_df.drop(columns=drop_columns)
+            # create new column for the ranks
+            rd_seg_df["rank"] = _dict["rank"]
+            # add the dataframe to the list
+            read_df_list.append(rd_seg_df)
+
+        if wr_seg_df.size:
+            # drop unused columns from the dataframe
+            wr_seg_df = wr_seg_df.drop(columns=drop_columns)
+            # create new column for the ranks
+            wr_seg_df["rank"] = _dict["rank"]
+            # add the dataframe to the list
+            write_df_list.append(wr_seg_df)
+
+    if read_df_list:
+        # concatenate the list of pandas dataframes into
+        # a single one with new row indices
+        read_df = pd.concat(read_df_list, ignore_index=True)
+    else:
+        # if the list is empty assign an empty dataframe
+        read_df = pd.DataFrame()
+
+    if write_df_list:
+        # concatenate the list of pandas dataframes into
+        # a single one with new row indices
+        write_df = pd.concat(write_df_list, ignore_index=True)
+    else:
+        # if the list is empty assign an empty dataframe
+        write_df = pd.DataFrame()
+
+    return read_df, write_df
+
+
+def get_single_df_dict(
+    report: Any,
+    mods: Sequence[str] = ["DXT_POSIX"],
+    ops: Sequence[str] = ["read", "write"],
+) -> Dict[str, Dict[str, pd.DataFrame]]:
+    """
+    Reorganizes segmented read/write data into a single ``pd.DataFrame``
+    and stores them in a dictionary with an entry for each DXT module.
+
+    Parameters
+    ----------
+
+    report: a ``darshan.DarshanReport``.
+
+    mods: a list of keys designating which Darshan modules to use for
+    data aggregation. Default is ``["DXT_POSIX"]``.
+
+    ops: a list of keys designating which Darshan operations to use for
+    data aggregation. Default is ``["read", "write"]``.
+
+    Returns
+    -------
+
+    flat_data_dict: a nested dictionary where the input module
+    keys (i.e. "DXT_POSIX") are the top level keys, which contain
+    an entry for each input operation (i.e. "read"/"write") that
+    map to dataframes containing all events for the specified operation.
+
+    Examples
+    --------
+    `flat_data_dict` generated from `tests/input/sample-dxt-simple.darshan`:
+        {
+            'DXT_POSIX':
+                {
+                    'read':
+                        Empty DataFrame
+                        Columns: []
+                        Index: [],
+                    'write':
+                        length  start_time  end_time  rank
+                        0      40    0.103379  0.103388     0
+                        1    4000    0.104217  0.104231     0
+                }
+        }
+
+    """
+    # initialize an empty dictionary for storing
+    # module and read/write data
+    flat_data_dict = {}  # type: Dict[str, Dict[str, pd.DataFrame]]
+    # iterate over the modules (i.e. DXT_POSIX)
+    for module_key in mods:
+        # read in the module data, update the name records
+        report.mod_read_all_dxt_records(module_key, dtype="pandas")
+        # retrieve the list of records in pd.DataFrame() form
+        dict_list = report.records[module_key].to_df()
+        # retrieve the list of read/write dataframes from the list of records
+        read_df, write_df = get_rd_wr_dfs(dict_list=dict_list)
+        # create empty dictionary for each module
+        flat_data_dict[module_key] = {}
+        if "read" in ops:
+            # add the concatenated dataframe to the flat dictionary
+            flat_data_dict[module_key]["read"] = read_df
+        if "write" in ops:
+            # add the concatenated dataframe to the flat dictionary
+            flat_data_dict[module_key]["write"] = write_df
+
+    return flat_data_dict
+
+
+def get_aggregate_data(
+    report: Any,
+    mods: Sequence[str] = ["DXT_POSIX"],
+    ops: Sequence[str] = ["read", "write"],
+) -> pd.DataFrame:
+    """
+    Aggregates the data based on which
+    modules and operations are selected.
+
+    Parameters
+    ----------
+
+    report: a ``darshan.DarshanReport``.
+
+    mods: a list of keys designating which Darshan modules to use for
+    data aggregation. Default is ``["DXT_POSIX"]``.
+
+    ops: a list of keys designating which Darshan operations to use for
+    data aggregation. Default is ``["read", "write"]``.
+
+    Returns
+    -------
+
+    agg_df: a ``pd.DataFrame`` containing the aggregated data determined
+    by the input modules and operations.
+
+    Raises
+    ------
+
+    ValueError: raised if the selected modules/operations
+    don't contain any data.
+
+    Notes
+    -----
+    Since read and write events are considered unique events, if both are
+    selected their dataframes are simply concatenated.
+
+    Examples
+    --------
+    `agg_df` generated from `tests/input/sample-dxt-simple.darshan`:
+
+            length  start_time  end_time  rank
+        0      40    0.103379  0.103388     0
+        1    4000    0.104217  0.104231     0
+
+    """
+    # collect the concatenated dataframe data from the darshan report
+    df_dict = get_single_df_dict(report=report, mods=mods, ops=ops)
+    # TODO: generalize for all DXT modules, for now manually set `DXT_POSIX`
+    module_key = "DXT_POSIX"
+    # iterate over each dataframe based on which operations are selected
+    df_list = []
+    for op_key, op_df in df_dict[module_key].items():
+        # if the dataframe has data, append it to the list
+        if op_df.size:
+            df_list.append(op_df)
+
+    if df_list:
+        # if there are dataframes in the list, concatenate them into 1 dataframe
+        agg_df = pd.concat(df_list, ignore_index=True)
+    else:
+        raise ValueError(f"No data available for selected module(s) and operation(s).")
+
+    return agg_df
+
+
+def calc_prop_data_sum(
+    tmin: float,
+    tmax: float,
+    total_elapsed: np.ndarray,
+    total_data: np.ndarray,
+) -> float:
+    """
+    Calculates the proportion of data read/written in the
+    time interval of a single bin in ``get_heatmap_data``.
+
+    Parameters
+    ----------
+
+    tmin: the lower bound of the time interval for a given bin.
+
+    tmax: the upper bound of the time interval for a given bin.
+
+    total_elapsed: an array of the elapsed times for every event
+    that occurred within the time interval of a given bin.
+
+    total_data: an array of the data totals for every event that
+    occurred within the time interval of a given bin.
+
+    Returns
+    -------
+
+    prop_data_sum: the amount of data read/written in the time
+    interval of a given bin.
+
+    """
+    # calculate the elapsed time
+    partial_elapsed = tmax - tmin
+    # calculate the ratio of the elapsed time
+    # to the total read/write event time
+    proportionate_time = partial_elapsed / total_elapsed
+    # calculate the amount of data read/written in the elapsed
+    # time (assuming a constant read/write rate)
+    proportionate_data = proportionate_time * total_data
+    # sum the data
+    prop_data_sum = proportionate_data.sum()
+    return prop_data_sum
+
+
+def get_heatmap_data(agg_df: pd.DataFrame, xbins: int) -> np.ndarray:
+    """
+    Builds an array similar to a 2D-histogram, where the y data is the unique
+    ranks and the x data is time. Each bin is populated with the data sum
+    and/or proportionate data sum for all IO events read/written during the
+    time spanned by the bin.
+
+    Parameters
+    ----------
+
+    agg_df: a ``pd.DataFrame`` containing the aggregated data determined
+    by the input modules and operations.
+
+    xbins: the number of x-axis bins to create.
+
+    Returns
+    -------
+
+    hmap_data: ``NxM`` array, where ``N`` is the number of unique ranks
+    and ``M`` is the number of x-axis bins. Each element contains the
+    data read/written by the corresponding rank in the x-axis bin time
+    interval.
+
+    """
+    # get the unique ranks
+    unique_ranks = np.unique(agg_df["rank"].values)
+
+    # generate the bin edges by generating an array of length n_bins+1, then
+    # taking pairs of data points as the min/max bin value
+    min_time = 0.0
+    max_time = agg_df["end_time"].values.max()
+    bin_edge_data = np.linspace(min_time, max_time, xbins + 1)
+
+    # calculate the elapsed time for each data point
+    elapsed_time_data = agg_df["end_time"].values - agg_df["start_time"].values
+
+    # generate an array for the heatmap data
+    hmap_data = np.zeros((unique_ranks.size, xbins), dtype=float)
+
+    # iterate over the unique ranks
+    for i, rank in enumerate(unique_ranks):
+        # for each rank, generate a mask to select only
+        # the data points that correspond to that rank
+        rank_mask = agg_df["rank"].values == rank
+        bytes_data = agg_df["length"].values[rank_mask]
+        start_data = agg_df["start_time"].values[rank_mask]
+        end_data = agg_df["end_time"].values[rank_mask]
+        elapsed_data = elapsed_time_data[rank_mask]
+
+        # iterate over the bins
+        for j, (bmin, bmax) in enumerate(zip(bin_edge_data[:-1], bin_edge_data[1:])):
+            # create a mask for all data with a start time greater than the
+            # bin minimum time
+            start_mask_min = start_data >= bmin
+            # create a mask for all data with an end time less than the
+            # bin maximum time
+            end_mask_max = end_data <= bmax
+            # use the above masks to find the indices of all data with
+            # start/end times that fall within the bin min/max times
+            start_inside_idx = np.nonzero(start_mask_min & (start_data <= bmax))[0]
+            end_inside_idx = np.nonzero((end_data > bmin) & end_mask_max)[0]
+            # using the start/end indices, find all indices that both start
+            # and end within the bin min/max limits
+            inside_idx = np.intersect1d(start_inside_idx, end_inside_idx)
+            # use the original masks to find the indices of data that start
+            # before the bin minimum time and end after the bin maximum time
+            outside_idx = np.nonzero(~start_mask_min & ~end_mask_max)[0]
+
+            if inside_idx.size:
+                # for data that start/end inside the bin limits,
+                # add the sum of the correspondign data to the hmap data
+                hmap_data[i, j] += bytes_data[inside_idx].sum()
+                # now remove any indices from the start/end index arrays
+                # to prevent double counting
+                start_inside_idx = np.setdiff1d(start_inside_idx, inside_idx)
+                end_inside_idx = np.setdiff1d(end_inside_idx, inside_idx)
+
+            if outside_idx.size:
+                # for data that start before the bin min time and end
+                # after the bin max time (run longer than 1 bin time),
+                # calculate the proportionate data used in 1 bin time
+                # and add it to the hmap data
+                prop_data_sum = calc_prop_data_sum(
+                    tmin=bmin,
+                    tmax=bmax,
+                    total_elapsed=elapsed_data[outside_idx],
+                    total_data=bytes_data[outside_idx],
+                )
+                hmap_data[i, j] += prop_data_sum
+
+            if start_inside_idx.size:
+                # for data with only a start time within the bin limits,
+                # calculate the elapsed time (from start to bin max), use the
+                # elapsed time to calculate the proportionate data read/written,
+                # and add the data sum to the hmap data
+                prop_data_sum = calc_prop_data_sum(
+                    tmin=start_data[start_inside_idx],
+                    tmax=bmax,
+                    total_elapsed=elapsed_data[start_inside_idx],
+                    total_data=bytes_data[start_inside_idx],
+                )
+                hmap_data[i, j] += prop_data_sum
+
+            if end_inside_idx.size:
+                # for data with only an end time within the bin limits,
+                # calculate the elapsed time (from bin min to end time), use the
+                # elapsed time to calculate the proportionate data read/written,
+                # and add the data sum to the hmap data
+                prop_data_sum = calc_prop_data_sum(
+                    tmin=bmin,
+                    tmax=end_data[end_inside_idx],
+                    total_elapsed=elapsed_data[end_inside_idx],
+                    total_data=bytes_data[end_inside_idx],
+                )
+                hmap_data[i, j] += prop_data_sum
+
+    return hmap_data
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -3,7 +3,13 @@ Module of data pre-processing functions for constructing the heatmap figure.
 """
 from __future__ import annotations
 
-from typing import Dict, Any, Tuple, Sequence, TypedDict, TYPE_CHECKING
+from typing import Dict, Any, Tuple, Sequence, TYPE_CHECKING
+
+import sys
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
 
 import pandas as pd
 import numpy as np
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -1,11 +1,15 @@
 """
 Module of data pre-processing functions for constructing the heatmap figure.
 """
-from typing import Dict, Any, Tuple, Sequence, TypedDict
+from __future__ import annotations
+
+from typing import Dict, Any, Tuple, Sequence, TypedDict, TYPE_CHECKING
 
 import pandas as pd
 import numpy as np
-import numpy.typing as npt
+
+if TYPE_CHECKING:
+    import numpy.typing as npt
 
 
 class SegDict(TypedDict):
@@ -22,7 +26,9 @@ class SegDict(TypedDict):
     read_segments: pd.DataFrame
 
 
-def get_rd_wr_dfs(dict_list: Sequence[SegDict]) -> Tuple[pd.DataFrame, pd.DataFrame]:
+def get_rd_wr_dfs(
+    dict_list: Sequence[SegDict], ops: Sequence[str] = ["read", "write"]
+) -> Dict[str, pd.DataFrame]:
     """
     Uses the DXT records to construct individual
     dataframes for both read and write segments.
@@ -38,12 +44,15 @@ def get_rd_wr_dfs(dict_list: Sequence[SegDict]) -> Tuple[pd.DataFrame, pd.DataFr
     ```pd.DataFrame`` containing the following data (columns):
     'offset', 'length', 'start_time', 'end_time'.
 
+    ops: a sequence of keys designating which Darshan operations
+    to collect data for. Default is ``["read", "write"]``.
+
     Returns
     -------
 
-    Tuple of form ``(read_df, write_df)``,
-    where each tuple element is a ``pd.DataFrame`` object
-    containing all of the read and write events.
+    rd_wr_dfs: dictionary where each key is an operation from the input
+    ``ops`` parameter (i.e. "read", "write") and each value is a
+    ``pd.DataFrame`` object containing all of the read/write events.
 
     Notes
     -----
@@ -53,7 +62,7 @@ def get_rd_wr_dfs(dict_list: Sequence[SegDict]) -> Tuple[pd.DataFrame, pd.DataFr
     Examples
     --------
 
-    ``dict_list`` and ``(read_df, write_df)``
+    ``dict_list`` and ``rd_wr_dfs``
     generated from ``tests/input/sample-dxt-simple.darshan``:
 
         dict_list = [
@@ -87,63 +96,52 @@ def get_rd_wr_dfs(dict_list: Sequence[SegDict]) -> Tuple[pd.DataFrame, pd.DataFr
             },
         ]
 
-        (read_df, write_df) = (
-            Empty DataFrame
-            Columns: []
-            Index: [],
-            length  start_time  end_time  rank
-            0      40    0.103379  0.103388     0
-            1    4000    0.104217  0.104231     0
-        )
+        rd_wr_dfs = {
+            'read':
+                Empty DataFrame
+                Columns: []
+                Index: [],
+            'write':
+                length  start_time  end_time  rank
+                0      40    0.103379  0.103388     0
+                1    4000    0.104217  0.104231     0
+        }
 
     """
     # columns to drop when accumulating the dataframes.
     # Currently "offset" data is not utilized
     drop_columns = ["offset"]
-    # create empty arrays to store
-    # read/write segment dataframes
-    read_df_list = []
-    write_df_list = []
-    # iterate over all records/dictionaries
-    # to pull out the dataframes
-    for _dict in dict_list:
-        # collect the read and write segment dataframes
-        rd_seg_df = _dict["read_segments"]
-        wr_seg_df = _dict["write_segments"]
+    # create empty dictionary to store
+    # the concatenated read/write dataframes
+    rd_wr_dfs = {}
+    # iterate over each operation
+    for op_key in ops:
+        # mypy can't tell that these keys are in fact
+        # in the ``SegDict``, so just ignore the type
+        seg_key = op_key + "_segments"  # type: ignore
+        # create empty list to store each dataframe
+        df_list = []
+        # iterate over all records/dictionaries
+        for _dict in dict_list:
+            # ignore for the same reason as above
+            seg_df = _dict[seg_key]  # type: ignore
+            if seg_df.size:
+                # drop unused columns from the dataframe
+                seg_df = seg_df.drop(columns=drop_columns)
+                # create new column for the ranks
+                seg_df["rank"] = _dict["rank"]
+                # add the dataframe to the list
+                df_list.append(seg_df)
 
-        if rd_seg_df.size:
-            # drop unused columns from the dataframe
-            rd_seg_df = rd_seg_df.drop(columns=drop_columns)
-            # create new column for the ranks
-            rd_seg_df["rank"] = _dict["rank"]
-            # add the dataframe to the list
-            read_df_list.append(rd_seg_df)
+        if df_list:
+            # concatenate the list of pandas dataframes into
+            # a single one with new row indices
+            rd_wr_dfs[op_key] = pd.concat(df_list, ignore_index=True)
+        else:
+            # if the list is empty assign an empty dataframe
+            rd_wr_dfs[op_key] = pd.DataFrame()
 
-        if wr_seg_df.size:
-            # drop unused columns from the dataframe
-            wr_seg_df = wr_seg_df.drop(columns=drop_columns)
-            # create new column for the ranks
-            wr_seg_df["rank"] = _dict["rank"]
-            # add the dataframe to the list
-            write_df_list.append(wr_seg_df)
-
-    if read_df_list:
-        # concatenate the list of pandas dataframes into
-        # a single one with new row indices
-        read_df = pd.concat(read_df_list, ignore_index=True)
-    else:
-        # if the list is empty assign an empty dataframe
-        read_df = pd.DataFrame()
-
-    if write_df_list:
-        # concatenate the list of pandas dataframes into
-        # a single one with new row indices
-        write_df = pd.concat(write_df_list, ignore_index=True)
-    else:
-        # if the list is empty assign an empty dataframe
-        write_df = pd.DataFrame()
-
-    return read_df, write_df
+    return rd_wr_dfs
 
 
 def get_single_df_dict(
@@ -202,15 +200,12 @@ def get_single_df_dict(
         # retrieve the list of records in pd.DataFrame() form
         dict_list = report.records[module_key].to_df()
         # retrieve the list of read/write dataframes from the list of records
-        read_df, write_df = get_rd_wr_dfs(dict_list=dict_list)
+        rd_wr_dfs = get_rd_wr_dfs(dict_list=dict_list, ops=ops)
         # create empty dictionary for each module
         flat_data_dict[module_key] = {}
-        if "read" in ops:
+        for op_key in ops:
             # add the concatenated dataframe to the flat dictionary
-            flat_data_dict[module_key]["read"] = read_df
-        if "write" in ops:
-            # add the concatenated dataframe to the flat dictionary
-            flat_data_dict[module_key]["write"] = write_df
+            flat_data_dict[module_key][op_key] = rd_wr_dfs[op_key]
 
     return flat_data_dict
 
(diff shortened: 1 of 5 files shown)


nawtrey:
Here is an example I created using `numpy.random()` data. I made some selections to test the marginal bar graphs and make sure the bin ranges were working properly. 

test_IO_plot_random_data

nawtrey:
Here is the output from the example log `dxt.darshan`. Since this one only contains data for a single rank, the horizontal bar graph is turned off. This one is likely being affected by the average time issue, so there is very little IO activity showing up.  
dxt

nawtrey:
This is the output for the example log `ior_hdf5_example.darshan`. This one could also be suffering from poor IO activity visualization as a result of using the average time. Additionally the horizontal bar graph for rank 3 doesn't seem right, I'll have to look into that. 

ior_hdf5_example

carns:
Those look really nice @nawtrey

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,319 @@
+import os
+import sys
+import glob
+import argparse
+
+import darshan
+import numpy as np
+import seaborn as sns
+import matplotlib.pyplot as plt
```
tylerjereddy, line 9: when this script is executed from repo path `darshan/darshan-util/pydarshan` via for example `python darshan/experimental/plots/plot_dxt_heatmap.py /home/tyler/LANL/rough_work/darshan/treddy_mpi-io-test_id545458_6-3-43062-1247823192538596076_13.darshan`

it will error out because of #384 

```
  File "darshan/experimental/plots/plot_dxt_heatmap.py", line 8, in <module>
    import seaborn as sns
  File "/home/tyler/python_37_venv/lib/python3.7/site-packages/seaborn/__init__.py", line 2, in <module>
    from .rcmod import *  # noqa: F401,F403
  File "/home/tyler/python_37_venv/lib/python3.7/site-packages/seaborn/rcmod.py", line 5, in <module>
    import matplotlib as mpl
  File "/home/tyler/github_projects/darshan/darshan-util/pydarshan/darshan/experimental/plots/matplotlib.py", line 8, in <module>
ModuleNotFoundError: No module named 'matplotlib.pyplot'; 'matplotlib' is not a package
```

Either someone should open a separate PR to remove that module or it should be temporarily removed here as it is in Jakob's PR so that the code can be invoked that way.

tylerjereddy (commented):
Here's one of the comparisons with Jakob's code:

treddy_mpi-io-test_id545458_6-3-43062-1247823192538596076_13

vs.

jakob_2k

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
jakobluettgau: So I would suggest to merge #388. I have since renamed the plot_dxt_heatmap.py to plot_dxt_heatmap2.py so there shouldn't be any conflicts. Once this is ready we can drop the other dxt heatmap. I am also fine with removing it from the other pull request completely.

tylerjereddy:
Some things that @carns and @shanedsnyder (and others) could provide feedback on here, without even checking the code, is whether they prefer:

- a particular color scheme for the heatmap? there are various complicated debates about proper contrast and colorblind-sensitive schemes; the comparison I did above should already give you a (fairly striking) taste for the impact the color scheme can make
- whether they want i.e., logscaling for the heatmap (colors), or dynamic adaptation of the scaling based on input, etc. (i.e., linear for small diffs, log for large gap situations, etc.)

carns:
> Some things that @carns and @shanedsnyder (and others) could provide feedback on here, without even checking the code, is whether they prefer:
> 
>     * a particular color scheme for the heatmap? there are various complicated debates about proper contrast and colorblind-sensitive schemes; the comparison I did above should already give you a (fairly striking) taste for the impact the color scheme can make
> 
>     * whether they want i.e., logscaling for the heatmap (colors), or dynamic adaptation of the scaling based on input, etc. (i.e., linear for small diffs, log for large gap situations, etc.)

Re: color scheme, I don't have a strong opinion on what spectrum to use, but I do think it makes more sense for zero to be white to make it more visually obvious which regions have no I/O.

I'm not sure about the scaling question to be honest.  Maybe start with whatever seems simplest and/or most likely to work and we'll just check how it looks on a range of log files?

nawtrey:
> I'm not sure about the scaling question to be honest. Maybe start with whatever seems simplest and/or most likely to work and we'll just check how it looks on a range of log files?

Based on some of the testing I've done with the colormaps, I recommend we automate it and use a cutoff. For example, if the data varies more than 2 orders of magnitude, use the log normalization:

```python
def use_log_norm(data, cutoff=2): 
    min_power = np.log10(data.min()) 
    max_power = np.log10(data.max()) 
    power_diff = max_power - min_power 
    if power_diff > cutoff:
        # use log10 normalization 
        return True  
    else: 
        # use standard normalization
        return False 
```

This should make cases like https://github.com/darshan-hpc/darshan/pull/396#pullrequestreview-680162455 have more color diversity. But I'm fine with setting it now and changing it later if we find it necessary.

carns:
That suggestion sounds good to me @nawtrey

nawtrey:
Out of the tasks listed in the [TODO list](https://github.com/darshan-hpc/darshan/pull/396#issue-665644066), there are 2 that I believe are much higher priority:
  1. getting the read/write data to be plotted with different colors (effectively 2 heat maps overlaid on top of each other, Phil's mock up had red/blue)
  2. fixing the way I use the time data so it doesn't take the average of the start/stop times but instead uses the elapsed time to determine how much data to put in each bin (or something similar). 

For 1.) I think @carns would have to comment on what the thinks about separating read/write data vs. what we have right now. At the moment it concatenates the read/write segment data and sorts them into the same bins. If this method works for now then we can open an issue about it if/when this gets merged and discuss it later. 

For 2.) I need to explain the problem more clearly so a log can be generated to test this issue. The current code uses the average time for each read/write event (`start_time + end_time / 2`) to bin the timing data. To my understanding, this should work well enough if the read/write events elapsed times' are much smaller than the overall run time for the log. But if we have the opposite case, where **read/write events take significant portions of the overall run time**, we will get a very misleading picture with the current method. 

For example, if you had a simple log with only 1 rank and 4 write events, where each write event was 2 seconds long, and the total run time was 10 seconds, you would expect to see the majority of the bins filled since 80% of the total time is spent writing data. But right now you would only see 4 bins (out of 200) filled, which would appear suggest there is barely any IO occurring and would be misleading. Phil originally described filling the bins by the proportion of data that was used in that time interval, so that's what I'll be working towards with this (and should fix the issue).

carns:
> For 1.) I think @carns would have to comment on what the thinks about separating read/write data vs. what we have right now. At the moment it concatenates the read/write segment data and sorts them into the same bins. If this method works for now then we can open an issue about it if/when this gets merged and discuss it later.

I'm fine just going with the merged view for now.  We might have to look at a few examples to see what it looks like.  Might end up being kind of muddled to have two gradients on the same plot.

Assuming the plotter is fairly modular you could always create separate plots that show just the writes or just the reads as an interim option too.

tylerjereddy:
For the "test case" for the second log above, could Nik try simply writing a Python script that writes data to a file endlessly and monitor it with darshan with DXT tracing on? CPython is just using C under the hood, no MPI, but that might be "ok" for generating the test data for DXT-POSIX?

That would mean building the `darshan-runtime` locally or on an HPC machine at LANL, which I'm not sure Nik has done, but probably useful to try anyway and can ping the team as needed if there are issues with that (there likely will be on first try).

I believe darshan does struggle to track forked/multiprocess Python, but we only need one "rank" to reproduce this issue initially if I understand correctly.

tylerjereddy:
(still useful to be able to generate and test new logs even if the issue appears to be resolved)

nawtrey:
> > For 1.) I think @carns would have to comment on what the thinks about separating read/write data vs. what we have right now. At the moment it concatenates the read/write segment data and sorts them into the same bins. If this method works for now then we can open an issue about it if/when this gets merged and discuss it later.
> 
> I'm fine just going with the merged view for now. We might have to look at a few examples to see what it looks like. Might end up being kind of muddled to have two gradients on the same plot.

Okay, I can generate some examples and we can see if we like them or not. I'll probably open an issue when this gets merged and drop them there.  

> For the "test case" for the second log above, could Nik try simply writing a Python script that writes data to a file endlessly and monitor it with darshan with DXT tracing on? CPython is just using C under the hood, no MPI, but that might be "ok" for generating the test data for DXT-POSIX?
> 
> That would mean building the `darshan-runtime` locally or on an HPC machine at LANL, which I'm not sure Nik has done, but probably useful to try anyway and can ping the team as needed if there are issues with that (there likely will be on first try).

I did start down that path a while ago, but I don't think I made it very far. I can circle back, I'm sure that will be useful as we move forward.

nawtrey:
I've pushed some changes that fix issue 2.) in https://github.com/darshan-hpc/darshan/pull/396#issuecomment-862731274, or at least *seem* to fix it. I added a new function, `get_heatmap_data()`, that bins the data manually and checks that the data sum is conserved, but I haven't done any rigorous testing on it. 

What I can say is some of the logs I've ran look different now, such as `ior_hdf5_example.darshan` (original output in https://github.com/darshan-hpc/darshan/pull/396#issuecomment-857466004):
ior_hdf5_example

nawtrey:
I was also able to generate the figures Tyler requested for comparison in #388.

Compare to https://github.com/darshan-hpc/darshan/pull/388#discussion_r644288966:
sample-dxt-simple

nawtrey:
Compare to https://github.com/darshan-hpc/darshan/pull/388#discussion_r644370247:
treddy_mpi-io-test_id4373053_6-2-60198-9815401321915095332_1

nawtrey:
Compare to https://github.com/darshan-hpc/darshan/pull/388#discussion_r644400154:
treddy_mpi-io-test_id17541138_6-2-65169-14537227514913022501_15

nawtrey:
Compare to https://github.com/darshan-hpc/darshan/pull/388#discussion_r645060554:
treddy_mpi-io-test_id545458_6-3-43062-1247823192538596076_13

nawtrey:
The last one (2400 ranks) looks quite a bit different from the output in [this comment](https://github.com/darshan-hpc/darshan/pull/396#pullrequestreview-680162455), again illustrating the issue with using the mean-time of each read/write event.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
nawtrey: I'll open a PR to fix #384 so we can merge those changes and rebase.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,440 @@
...
+    # for the x tick labels, start at 0 and end with the max time (converted to
+    # an integer)
+    max_time = np.max(end_time_data)
+    if max_time < 1:
+        x_ticklabels = np.around(np.linspace(0, max_time, n_xlabels), decimals=2)
+    elif (max_time > 1) & (max_time < 10):
+        x_ticklabels = np.around(np.linspace(0, max_time, n_xlabels), decimals=1)
+    else:
+        x_ticklabels = np.asarray(
+            np.linspace(0, np.ceil(max_time), n_xlabels), dtype=int
```
tylerjereddy, line 303: `linspace` accepts a `dtype` argument and returns an array, so why wrap it in an `asarray` call that coerces dtype and checks if it is an array?

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,440 @@
...
+    if unique_ranks.size > 12:
+        # only use 6 labels if there are many ranks to label. Might want to
+        # do some checks on how many ranks there are earlier on since any
+        # file with 1000's of ranks is too messy to plot this way.
+        n_ylabels = 6
+        sns_yticks = jgrid.ax_joint.get_yticks()
+        yticks = np.linspace(sns_yticks.min(), sns_yticks.max(), n_ylabels)
+        jgrid.ax_joint.set_yticks(yticks)
+        yticklabels = np.asarray(
+            np.linspace(unique_ranks.min(), unique_ranks.max(), n_ylabels), dtype=int
```
tylerjereddy, line 317: same comment here--`linspace` should already produce an array and provide a means to specify dtype?

tylerjereddy:
Starting to think about regression tests for some of the functions may make sense, perhaps starting with the second point mentioned above. Things to sort out may be:

- learning to generate the cases as discussed above
- sorting out if it is suitable to add more darshan logs to this repo (size/binary asset storage concerns?); some tests may be sufficiently covered with synthetic data for a given function
- if the current data munging functions are in conveniently-testable units

I'll try to check the code more carefully when I have some time. Checking some of the other PRs and doing code review in them is also likely useful in the meanwhile--getting `darshan-util` built properly locally to start probing the APMPI stuff, etc.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
jakobluettgau: https://github.com/darshan-hpc/darshan/pull/388 already made the changes that fixed any matplotlib import problems, and takes care of the fallout across examples. That's why I was suggesting, to go ahead to merge 388 that after making sure it does not conflict with any of the files you touched here ;)

carns:
>     * sorting out if it is suitable to add more darshan logs to this repo (size/binary asset storage concerns?); some tests may be sufficiently covered with synthetic data for a given function

I'm open to suggestions on how to organize this, but on paper at least it seems like it would be nice to have example logs (for any purpose, whether its testing, regression, tutorial examples, etc.) make their way into a separate repo that is nothing but logs and descriptions of logs in https://github.com/darshan-hpc/darshan-logs.

Test scripts (e.g., whatever the python take on `make check` is) could clone it.  Github will allow individual files up to 100 MB in size; beyond that we would need another solution, but I think 100 MB will get us a long ways.

How to organize beyond that I don't know.  Some logs will be useful for multiple purposes.  We could (for example) have a subdir containing symlinks to all of the logs that we think are relevant for darshan report testing, for example, so that they can all be in one quick list even though they are scattered in a directory hierarchy for some other organizational reason?

tylerjereddy:
Phil's testing idea sounds sensible. Network connectivity requirements on unit tests have drawbacks of course, but we could also conditionally skip tests in air gapped scenarios and so on I suppose.

I bet we can get a fair bit of mileage by testing individual functions with just subsets of data/small synthetic data as well (i.e., run the control flow with a "problem" log file, print out the values that flank the problem function, and use a  small data structure that mimics that problem).

tylerjereddy:
Also, to nudge the integration/discussion between this work and #386 here is an example of hacking/integrating this DXT plotting infrastructure into the CLI script over there (see sample plot/and code diff below). This is just a hack, but it is meant to catalyze some discussion on the integration mechanics between folks on the team.

The proper integration, instead of this hack, may also require changes to the html, which would be most unfortunate in my view. However, I'm not really going to judge the need for an html report on top of a PDF report--maybe folks can work together a bit to sort out some consensus on their relative importance.

Also, keep in mind that the DXT plotting code requires DXT tracing data, while the CLI script **still** has a hard requirement for APMPI data to avoid errors (hence, the additional hacks...).

image

```diff
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -18,6 +18,7 @@ import plotnine as p9
 from plotnine import *
 import darshan
 import darshan.templates
+from darshan.experimental.plots import plot_dxt_heatmap
 from PIL import Image
 
 import matplotlib.font_manager as fm
@@ -593,6 +594,16 @@ def main(args=None):
     else:
         apmpi_dict = None
 
+    # Plot DXT heatmap
+    # from Nik's PR...
+    heatmap = plot_dxt_heatmap.plot_dxt_heatmap(report=report)
+    buf = io.BytesIO()
+    heatmap.fig.savefig(buf, format='png')
+    buf.seek(0)
+    encoded = base64.b64encode(buf.read())
+    variables['plot_dxt_heatmap'] = encoded
+
+
     #
     # Plot I/O cost
     #
@@ -604,7 +615,13 @@ def main(args=None):
     variables['plot_op_count'] = plot_op_count(posix_df, mpiio_df, stdio_df).decode('utf-8')
 
     variables['fs_data'] = data_transfer_filesystem(report, posix_df, stdio_df)
-    apmpi_encoded = apmpi_process(apmpi_dict)
+    try:
+        # devs need to add appropriate error
+        # handling, do this for now
+        apmpi_encoded = apmpi_process(apmpi_dict)
+    except TypeError:
+        apmpi_encoded = False
+
     if apmpi_encoded:
        variables['apmpi_call_times_variance'] = apmpi_encoded[0].decode('utf-8')
        variables['apmpi_call_time_distribution'] = apmpi_encoded[1].decode('utf-8')
@@ -624,12 +641,17 @@ def main(args=None):
        for row in apmpi_encoded[7].iterrows():
            variables['apmpi_zero_rank'].append(row[1].values)
     else:
-       variables['apmpi_call_time'] = None
-       variables['apmpi_rank_totaltime'] = None
-       variables['apmpi_max_rank'] = None
-       variables['apmpi_min_rank'] = None
-       variables['apmpi_mean_rank'] = None
-       variables['apmpi_zero_rank'] = None
+       # NOTE: this is bad, shouldn't have to do this
+       variables['apmpi_call_times_variance'] = variables['plot_dxt_heatmap'].decode('utf-8')
+       variables['apmpi_call_time_distribution'] = []
+       variables['apmpi_call_bytes_per_op'] = []
+       variables['apmpi_call_rank0_msg_dist'] = []
+       variables['apmpi_call_time'] = []
+       variables['apmpi_rank_totaltime'] = []
+       variables['apmpi_max_rank'] = []
+       variables['apmpi_min_rank'] = []
+       variables['apmpi_mean_rank'] = []
+       variables['apmpi_zero_rank'] = []
     template_path = pkg_resources.path(darshan.templates, '')
     with template_path as path:

```

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
nawtrey: That's a good point -- I'll fix those in the next round of changes.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,440 @@
...
+    )
+    jgrid.ax_marg_y.tick_params(
+        axis="x", bottom=False, labelbottom=False, top=False, labeltop=False
+    )
+
+    # set the dimensions of the figure to 6.5" wide x 4.5" tall
+    jgrid.fig.set_size_inches(6.5, 4.5)
+
+    # set the spacing between subplots
+    hspace, wspace = 0.01, 0.02
```
nawtrey, line 344: Phil recommended I change the spacing between the bar graphs and the heatmap. Should be able to do that here and double check that colorbar is still spaced appropriately.

tylerjereddy:
Another thing that was mentioned during the meeting today is that we may need less testing if there were a ready-made solution for the i.e., binning code. It probably still would be worth testing even then, though perhaps not quite as stringently.

Indeed, part of my plan was to look at the tested code and then start messing with it to see if I can't find a way to use i.e., axis-oriented NumPy operations to start lifting things out of loops and still getting the tests to pass (faster development cycle than "does my plot still look the same?"). Or, more broadly, ways to cut down on complexity/maintenance burden while still having tests pass, etc.

carns:
> The proper integration, instead of this hack, may also require changes to the html, which would be most unfortunate in my view. However, I'm not really going to judge the need for an html report on top of a PDF report--maybe folks can work together a bit to sort out some consensus on their relative importance.

We talked through some of this live yesterday, but I'll try to capture some thoughts here for posterity. Both pdf and html are useful, but for different use cases. The PDF report is a hard early requirement because it's just so helpful for end user interaction (emailing reports around for discussion, etc.). HTML reports, however, are much more convenient for use cases like a facility web dashboard, making it possible for users (or admins) to just click through a web interface and find information associated with their jobs.  

This leads to the underlying technical issue: how do we most efficiently share work between the two objectives?  We don't want to end up in the situation we have with the old tools, in which the darshan-job-summary reports code was mostly unusable for a web dashboard, so you couldn't share work between a web dashboard and the summary.

The ideal solution would be to use a framework that is equally adept at rendering to both at a high level (just toggling --pdf or --html or some such).  Such a framework may not exist, though.  The next best thing would be to modularize the python routines that generate the plots such that the same underlying code for a given figure can be reasonably reused across the pdf report tool and html tools (or other things we haven't thought of yet) that are operating in different frameworks.  Each framework would do it's own layout, sections, etc but they would share the same callouts to make the actual plots.

We are in a good spot if in the future someone thinks of an interesting new plot for the pdf summary tools, and the same thing can be quickly added to html tools.

nawtrey:
I've been working on splitting up functions into smaller bits, writing tests, adding documentation (with some type hints), and changing the pre-processing to work in terms of `pandas` dataframes. I should be able to push some changes in the next few days, but it might be sort of in pieces at first. 

I'll make it clear when it's ready for more review.

nawtrey:
A lot of code has been moved into individual functions, and I moved the data pre-processing code into its own module (`heatmap_handling.py`), keeping the plot-related code in `plot_dxt_heatmap.py`. Both modules have been gone through and each function now has a docstring and type hints (some are just place holders). I tried to stick to the Google documentation style, but I wouldn't be surprised if I made some mistakes there. I have also changed the way the data aggregation works, taking advantage of `pd.concatenate()` to merge the read/write segment dataframes instead of pulling the data arrays out and concatenating them separately. This keeps things in terms of dataframes, which I'm thinking will allow for easier data manipulation in the long run.

Additionally, I've added two testing modules (`test_heatmap_handling.py` and `test_plot_dxt_heatmap.py`) that cover the new modules 100%.

To run tests (from `pydarshan` root):
```bash
pytest --cov-report term-missing --cov=darshan/experimental/plots
```

nawtrey:
I have removed the script that generates the plots since we likely won't be using it that way, but the current version (see below) can be pasted in and executed using the following format:
```bash
python ~/path/to/pasted_script.py ~/path/to/directory/or/log
```

The code currently saves the plots in the current working directory and generates one for each combination of operations (read, write, read+write), so most logs generate 3 separate figures. 

The code:
```python
import argparse
import os
import glob

from darshan.experimental.plots.plot_dxt_heatmap import plot_heatmap

if __name__ == "__main__":
    # get the path to the file(s) to be read in
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "datapath",
        nargs="?",
        help="Specify path to darshan log file or directory with darshan logs",
    )
    args = parser.parse_args()
    data_path = args.datapath

    if os.path.isfile(data_path):
        # individual file case
        files_to_process = [data_path]
    elif os.path.isdir(data_path):
        # directory case
        files_to_process = glob.glob(os.path.join(data_path, "*.darshan"))

    for filepath in files_to_process:
        # get the file name
        filename = os.path.basename(filepath)
        # iterate over each combination of read/write operations
        for ops in [["read"], ["write"], ["read", "write"]]:
            try:
                # generate the heatmap figure
                hmap = plot_heatmap(log_path=filepath, ops=ops)
            except NotImplementedError as err:
                print(err)
                continue
            except ValueError as err:
                print(err)
                continue
            except KeyError as err:
                print(err)
                continue

            # save the figure in the current directory using
            # the filename and the operation(s) used
            op_str = "_" + "_".join(ops)
            save_path = os.path.join(
                os.getcwd(), os.path.splitext(filename)[0] + op_str + ".png"
            )
            print(f"Saving figure at location: {save_path}")
            plt.savefig(save_path, dpi=300)
            plt.close()
```

darshan-util/pydarshan/tests/test_plot_dxt_heatmap.py
```diff
@@ -0,0 +1,458 @@
...
+    [
+        "examples/example-logs/ior_hdf5_example.darshan",
+        "examples/example-logs/dxt.darshan",
+        "tests/input/sample-dxt-simple.darshan",
+    ],
+)
+@pytest.mark.parametrize("mods", [["DXT_POSIX"], ["DXT_MPIIO"]])
+@pytest.mark.parametrize("ops", [["read", "write"], ["read"], ["write"]])
+def test_plot_heatmap(filepath, mods, ops):
+
```
nawtrey, line 410: I had a difficult time coming up with meaningful checks here (for `plot_dxt_heatmap.plot_heatmap()`). I feel like most of these checks are frivolous, outside of testing the proper errors are raised.

tylerjereddy:
Go ahead and rebase/force push and maybe lift the draft status so I can see the CI running as this PR evolves.

tylerjereddy (commented):
Running `pytest` locally fails; I think `setup.py` needs to be adjusted to include `seaborn`:

`requirements = ["cffi", "numpy", "pandas", "matplotlib"]`

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -0,0 +1,442 @@
...
+    for op_key, op_df in df_dict[module_key].items():
+        # if the dataframe has data, append it to the list
+        if op_df.size:
+            df_list.append(op_df)
+
+    if df_list:
+        # if there are dataframes in the list, concatenate them into 1 dataframe
+        agg_df = pd.concat(df_list, ignore_index=True)
+    else:
+        raise ValueError(f"No data available for selected module(s) and operation(s).")
```
tylerjereddy, line 278: What's the purpose of the f-string prefix here? I don't see a substitution field.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -0,0 +1,442 @@
...
+            flat_data_dict[module_key]["write"] = write_df
+
+    return flat_data_dict
+
+
+def get_aggregate_data(
+    report: Any,
+    mods: Sequence[str] = ["DXT_POSIX"],
+    ops: Sequence[str] = ["read", "write"],
+) -> pd.DataFrame:
```
tylerjereddy, line 221: Have you tested the `pandas` types with: https://github.com/pandas-dev/pandas/blob/master/pandas/_typing.py ?

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -0,0 +1,442 @@
...
+        if "write" in ops:
+            # add the concatenated dataframe to the flat dictionary
+            flat_data_dict[module_key]["write"] = write_df
+
+    return flat_data_dict
+
+
+def get_aggregate_data(
+    report: Any,
+    mods: Sequence[str] = ["DXT_POSIX"],
```
tylerjereddy, line 219: The docstring says "a list of keys.." and the typing is using the more generic `Sequence`--which is it?

tylerjereddy:
Yeah, CI failure is consistent with my local experience. Anyway, that particular issue is an easy fix.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -0,0 +1,442 @@
...
+        raise ValueError(f"No data available for selected module(s) and operation(s).")
+
+    return agg_df
+
+
+def calc_prop_data_sum(
+    tmin: float,
+    tmax: float,
+    total_elapsed: np.ndarray,
+    total_data: np.ndarray,
```
tylerjereddy, line 287: If we're going to start typing with NumPy we'll probably want to use: https://numpy.org/devdocs/reference/typing.html

It is pretty bleeding edge stuff, but it means we can type the data inside the array as well--an example is here:
https://github.com/scipy/scipy/pull/14281

So you get stuff like: `points: npt.NDArray[np.float64]`

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
nawtrey: I had it as `List[str]` originally, but [the documentation](https://docs.python.org/3/library/typing.html) led me to believe I should have `Sequence` or `Iterable`. 

The particular section I'm referring to:
```
class typing.List(list, MutableSequence[T])
Generic version of list. Useful for annotating return types. To annotate arguments 
it is preferred to use an abstract collection type such as Sequence or Iterable
```

tylerjereddy: Sounds good, so make the documentation consistent then? And be sure to point out the same in my PR, which I think may be using the List type for some args.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
tylerjereddy: Probably needs a guard like this as well because it is so new it will only be executable in a separate check that has newer NumPy that we can't expect folks to have in general:

```python
if TYPE_CHECKING:
    import numpy.typing as npt
```

Similar to this I suppose: https://github.com/scipy/scipy/pull/14174/files

darshan-util/pydarshan/tests/test_heatmap_handling.py
```diff
@@ -0,0 +1,561 @@
...
+    # run the aggregated data through the heatmap data code
+    actual_hmap_data = heatmap_handling.get_heatmap_data(agg_df=agg_df, xbins=xbins)
+
+    if filepath == "tests/input/sample-dxt-simple.darshan":
+        # check the data is conserved
+        assert actual_hmap_data.sum() == 4040
+        # make sure the output array is the correct shape
+        assert actual_hmap_data.shape == (1, xbins)
+        # make sure the output data contains identical values
+        assert_array_equal(actual_hmap_data, expected_hmap_data)
```
tylerjereddy, line 533: here and anywhere else that does a comparison of NumPy arrays that may contain floating point values should use `assert_allclose` to avoid potential issues with floating point imprecision

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -0,0 +1,442 @@
...
+                # calculate the elapsed time (from bin min to end time), use the
+                # elapsed time to calculate the proportionate data read/written,
+                # and add the data sum to the hmap data
+                prop_data_sum = calc_prop_data_sum(
+                    tmin=bmin,
+                    tmax=end_data[end_inside_idx],
+                    total_elapsed=elapsed_data[end_inside_idx],
+                    total_data=bytes_data[end_inside_idx],
+                )
+                hmap_data[i, j] += prop_data_sum
```
tylerjereddy, line 440: This massive pure-Python nested for loop is pretty awkward/likely to be slow, and if we're going to embrance pandas, we may as well use pandas to vectorize the operations.

Here is a sample diff I worked on for a few hours that passes most of the tests for this function and replaces that whole loop with vectorized pandas operations. Perhaps you can carry this across the finish line to get the last few cases passing, clean up the comments, reduce my messiness/duplication, etc. (and look for other parts of this PR that would benefit from a similar pandas vectorization, if applicable).

```diff
diff --git a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
index c2bd5c1..f948456 100644
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -355,88 +355,45 @@ def get_heatmap_data(agg_df: pd.DataFrame, xbins: int) -> np.ndarray:
     min_time = 0.0
     max_time = agg_df["end_time"].values.max()
     bin_edge_data = np.linspace(min_time, max_time, xbins + 1)
-
-    # calculate the elapsed time for each data point
-    elapsed_time_data = agg_df["end_time"].values - agg_df["start_time"].values
-
-    # generate an array for the heatmap data
-    hmap_data = np.zeros((unique_ranks.size, xbins), dtype=float)
-
-    # iterate over the unique ranks
-    for i, rank in enumerate(unique_ranks):
-        # for each rank, generate a mask to select only
-        # the data points that correspond to that rank
-        rank_mask = agg_df["rank"].values == rank
-        bytes_data = agg_df["length"].values[rank_mask]
-        start_data = agg_df["start_time"].values[rank_mask]
-        end_data = agg_df["end_time"].values[rank_mask]
-        elapsed_data = elapsed_time_data[rank_mask]
-
-        # iterate over the bins
-        for j, (bmin, bmax) in enumerate(zip(bin_edge_data[:-1], bin_edge_data[1:])):
-            # create a mask for all data with a start time greater than the
-            # bin minimum time
-            start_mask_min = start_data >= bmin
-            # create a mask for all data with an end time less than the
-            # bin maximum time
-            end_mask_max = end_data <= bmax
-            # use the above masks to find the indices of all data with
-            # start/end times that fall within the bin min/max times
-            start_inside_idx = np.nonzero(start_mask_min & (start_data <= bmax))[0]
-            end_inside_idx = np.nonzero((end_data > bmin) & end_mask_max)[0]
-            # using the start/end indices, find all indices that both start
-            # and end within the bin min/max limits
-            inside_idx = np.intersect1d(start_inside_idx, end_inside_idx)
-            # use the original masks to find the indices of data that start
-            # before the bin minimum time and end after the bin maximum time
-            outside_idx = np.nonzero(~start_mask_min & ~end_mask_max)[0]
-
-            if inside_idx.size:
-                # for data that start/end inside the bin limits,
-                # add the sum of the correspondign data to the hmap data
-                hmap_data[i, j] += bytes_data[inside_idx].sum()
-                # now remove any indices from the start/end index arrays
-                # to prevent double counting
-                start_inside_idx = np.setdiff1d(start_inside_idx, inside_idx)
-                end_inside_idx = np.setdiff1d(end_inside_idx, inside_idx)
-
-            if outside_idx.size:
-                # for data that start before the bin min time and end
-                # after the bin max time (run longer than 1 bin time),
-                # calculate the proportionate data used in 1 bin time
-                # and add it to the hmap data
-                prop_data_sum = calc_prop_data_sum(
-                    tmin=bmin,
-                    tmax=bmax,
-                    total_elapsed=elapsed_data[outside_idx],
-                    total_data=bytes_data[outside_idx],
-                )
-                hmap_data[i, j] += prop_data_sum
-
-            if start_inside_idx.size:
-                # for data with only a start time within the bin limits,
-                # calculate the elapsed time (from start to bin max), use the
-                # elapsed time to calculate the proportionate data read/written,
-                # and add the data sum to the hmap data
-                prop_data_sum = calc_prop_data_sum(
-                    tmin=start_data[start_inside_idx],
-                    tmax=bmax,
-                    total_elapsed=elapsed_data[start_inside_idx],
-                    total_data=bytes_data[start_inside_idx],
-                )
-                hmap_data[i, j] += prop_data_sum
-
-            if end_inside_idx.size:
-                # for data with only an end time within the bin limits,
-                # calculate the elapsed time (from bin min to end time), use the
-                # elapsed time to calculate the proportionate data read/written,
-                # and add the data sum to the hmap data
-                prop_data_sum = calc_prop_data_sum(
-                    tmin=bmin,
-                    tmax=end_data[end_inside_idx],
-                    total_elapsed=elapsed_data[end_inside_idx],
-                    total_data=bytes_data[end_inside_idx],
-                )
-                hmap_data[i, j] += prop_data_sum
-
-    return hmap_data
+    cats_start = pd.get_dummies(pd.cut(agg_df["start_time"], bin_edge_data))
+    cats_start["rank"] = agg_df["rank"]
+    cats_start["length"] = agg_df["length"]
+    cats_end = pd.get_dummies(pd.cut(agg_df["end_time"], bin_edge_data))
+    cats_end["rank"] = agg_df["rank"]
+    cats_end["length"] = agg_df["length"]
+    cats = cats_start.where(cats_start > cats_end, cats_end)
+    cats.iloc[:, :xbins] = cats.iloc[:, :xbins].replace(0, np.nan)
+    print("combined cats:\n", cats)
+    # at this point we have a dataframe of dummy (indicator)
+    # variables for each of the time segments (bins) for
+    # each IO event (row, which corresponds to a rank)
+    # each time bin that has an event has a 1 in it, otherwise
+    # NaN
+
+    # now, for each row (IO event) we want to fill in any empty (NaN)
+    # bins between filled bins because those are time spans b/w start
+    # and stop events; divide the indicators by the number of total
+    # bins spanned by the event that is filled in
+
+    # this is actually an inner linear interpolation, followed by
+    # division to produce weights on the bin events
+    cats.iloc[:, :xbins] = cats.iloc[:, :xbins].interpolate(method='linear',
+                                                            limit_area='inside',
+                                                            axis=1)
+    print("combined cats interpolated:\n", cats)
+    sums = cats.iloc[:, :xbins].sum(axis=1) 
+    print("sums:", sums)
+    cats.iloc[:, :xbins] = cats.iloc[:, :xbins].div(sums, axis=0)
+
+    # each full or fractional bin event is now multiplied by
+    # the bytes data
+    cats.iloc[:, :xbins] = cats.iloc[:, :xbins].mul(cats['length'], axis=0)
+    hmap_data = cats
+    hmap_data.set_index('rank', inplace=True)
+    hmap_data.drop('length', inplace=True, axis=1)
+    print("*** final hmap_data:\n", hmap_data)
+    grouped_res = hmap_data.groupby('rank').sum()
+    print("grouped result:", grouped_res)
+    array_res = grouped_res.to_numpy()
+    print("grouped ARRAY result:", array_res)
+    return array_res

```

tylerjereddy: There's always some tension between the convenience of the iterative solution and using the idiomatic vectorized approach--depends on whether the latter is actually possible within reason: https://stackoverflow.com/questions/7837722/what-is-the-most-efficient-way-to-loop-through-dataframes-with-pandas#comment9576192_7837947

The draft code above is actually slower I think, perhaps because it is making copies it doesn't need to (more opportunities for in-place ops?), using larger/duplicate dummy data structures than it has to, but probably for other reasons too. Fractional bin occupancy is perhaps the reason for the 3 test fails that remain.

tylerjereddy: Overarching objective is less concerned with raw speed initially and more on reducing our maintenance burden by abstracting to externally maintained (pandas/NumPy) functions vs. custom-rolled code to the extent possible.

It isn't a small ask though--looks challenging to vectorize, though I did do a fair bit of draft work above to demo one approximate approach.

tylerjereddy: A deeper question might be whether the data that is fed to `get_heatmap_data()` is actually in the most convenient form for idiomatic processing with NumPy/`pandas` given the target desired array data structure.

tylerjereddy: Ok, one more thought--the bins are kind of like time series data, and `pandas` also has extensive handling for those--makes me wonder if the custom-rolled "fractional occupancy" of a time bin might be handled by some pre-existing time series convenience functionality.

tylerjereddy: Circling back to my original suggested code, as opposed to the speculative comments, one thing I wonder about is this:

* I've made the assumption that each bin in an interpolated series should simply be an equal fraction of the bytes read or written in a particular DXT segment
* you might initially think this is wrong, because if i.e., the true start point for the interpolated series (the DXT segment start time) happens in the middle of the first time bin, then you've over-representing the total amount of data written in that bin and under-representing the fraction in subsequent frames
* however, you are lying regardless if you fill that first bin in--there's actually nothing happening at the start of that time bin, so formally it should be part-blank or a higher bin density is needed
* if you do down-weight these fractional bins, compared to the "internal"/interpolated bins, then you're also making another lie--that "lower data" first bin now suggests that the amount of data written per unit time for this rank fluctuates in ways the averaged data can't actually tell us
* is it better to misrepresent the *rate* of data transfer or the *amount* of data written per bin? if it doesn't really matter, then the less sophisticated approach (raw averaging) may allow for some convenient shortcuts

tylerjereddy:
Just a "brainstorm thought"--since binning is such a pain, what about plotting the data we actually have access to--a line/rectangle connecting the start and stop of each DXT segment with a colormap intensity set to `bytes / (stop - start)`, the averaged rate? 

Maybe there's a reason it doesn't work, can't quite think of one at the moment (overlapping segments possible for a single rank/row?).

carns:
> Just a "brainstorm thought"--since binning is such a pain, what about plotting the data we actually have access to--a line/rectangle connecting the start and stop of each DXT segment with a colormap intensity set to `bytes / (stop - start)`, the averaged rate?
> 
> Maybe there's a reason it doesn't work, can't quite think of one at the moment (overlapping segments possible for a single rank/row?).

Part of my rationale for trying to think of this as a histogram (and making ourselves lose fidelity in the process, one way or another) is the belief that in the long run we can come up with an algorithm for the Darshan library to generate (and save) a histogram at runtime by auto-scaling to a finite number of total bins per rank regardless of total execution time.  That's appealing because it would have a bounded size per rank and would be plausible to turn on by default for all applications.  We aren't going to be able to turn on pure DXT mode for general purpose production use because it produces too much data.

Eventually it would be nice for the job summary to handle either mode, perhaps just with some higher fidelity if the native DXT data is available.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
carns: I'm skeptical about not making proportional adjustments to bins, but I'm willing to be proven wrong :)  

Can we defer the decision about which translation method is the most faithful?  If we come up with a data structure we like for histogram, it would like to think (maybe naively) that there would be a `dxt_to_histogram()` function that we can treat as a black box for now and tinker with later. 

Philosophically, none of the translation methods are going to be correct exactly; we should think of it as a low fidelity intermediate representation for visualization purposes.  My intuition is that proportional binning will produce a better visual representation, but I don't really have anything to back that up.

tylerjereddy: Sounds good, so maybe the unit tests can remain enforced with proportional bins and my only remaining comment on this specific matter is that there may be an opportunity to clean the implementation up a bit/lot using the methods available on pandas/NumPy data structures.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -0,0 +1,442 @@
...
+    Custom type hint class for `dict_list` argument in `get_rd_wr_dfs()`.
+    """
+
+    id: int
+    rank: int
+    hostname: str
+    write_count: int
+    read_count: int
+    write_segments: pd.DataFrame
+    read_segments: pd.DataFrame
```
tylerjereddy, line 21: Can you clarify what this `SegDict` class does for static type checking? For example, if I modify it arbitrarily:

```diff
diff --git a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
index c2bd5c1..4bdd93f 100644
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -19,6 +19,7 @@ class SegDict(TypedDict):
     read_count: int
     write_segments: pd.DataFrame
     read_segments: pd.DataFrame
+    stuff: int
```

`mypy` is still perfectly happy with the check on the `darshan` namespace as far as I can tell (apart from the usual suspects)

tylerjereddy: Note that static type checking is probably pretty low on our priority list, but just chipping away at any complexity in the PR that I'm curious about.

nawtrey: The purpose of the `TypedDict` is to specify the types of each individual key/value pair in a dictionary. Since the record dictionaries are using `int`/`str`/pandas dataframes I figured it was worth adding detail. Otherwise I believe the only option is to use `dict`, which is really `dict[Any, Any]`, so it doesn't check anything inside the record dictionary. 

But this is interesting. I got the same thing when I tried it. I expected it would detect any missing/extra keys as-is, but now that I think about it, I'm not sure how it would know what keys were available since we never actually specify them. 

For example, if we had something like:
```python 
from typing import TypedDict 

class SegDict(TypedDict):
    foo: int 
    bar: str 

def func(test_dict: SegDict):
    test_dict['foo'] += 10
```
How would it ever know that `bar` is missing? It would have to _assume_ it is missing because we never explicitly write down what keys `test_dict` has. One could argue that if it isn't used in the function, then it should be assumed missing. But maybe `mypy` doesn't make that assumption because that would require you to write a custom `TypedDict` object for each function, even if the same object is getting passed around. 

Now, if you alter the `read_segment` or `write_segment` entries (i.e. remove one, change its type, etc.) you do get errors. So it is adding value in that way at least. You can also remove any of the entries that aren't used without error, and you can change `rank` to any type you want without error. Again, I think this is because none of them are used in a way that can be tested so `mypy` has no way of knowing what they are (and doesn't assume they are missing).

nawtrey: Adding to my last point -- if you change `read_segments` to `int`, the only reason it knows this is incorrect is because `int` doesn't have the `.size` or `.drop` methods. So I think they have to be used in a way that illustrates their type before `mypy` can make those determinations.

tylerjereddy: The explanation in the [docs](https://mypy.readthedocs.io/en/stable/more_types.html#typeddict):

> A TypedDict object with extra items is compatible with (a subtype of) a narrower TypedDict, assuming item types are compatible...

tylerjereddy: Anyway, this seems useful despite the limitations, so I'll mark this resolved, thanks!

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
tylerjereddy: We discussed this a bit recently--`pandas` typing isn't really mature/ready for prime time, but it is probably "ok" to leave in the placeholder pandas objects if folks want to, so I'll mark this resolved.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -0,0 +1,442 @@
...
+        # if the list is empty assign an empty dataframe
+        read_df = pd.DataFrame()
+
+    if write_df_list:
+        # concatenate the list of pandas dataframes into
+        # a single one with new row indices
+        write_df = pd.concat(write_df_list, ignore_index=True)
+    else:
+        # if the list is empty assign an empty dataframe
+        write_df = pd.DataFrame()
```
tylerjereddy, line 143: We could probably avoid the duplicated logic for reading and writing data structures I suspect, maybe with a loop if the data structures do indeed actually need to remain separate.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,412 @@
...
+    )
+    marg_y.tick_params(
+        axis="y", left=False, labelleft=False, right=False, labelright=False
+    )
+    marg_x.tick_params(
+        axis="y", left=False, labelleft=False, right=False, labelright=False
+    )
+    marg_y.tick_params(
+        axis="x", bottom=False, labelbottom=False, top=False, labeltop=False
+    )
```
tylerjereddy, line 243: Can reduce duplication/maintenance burden here:

```diff
+    for marg in [marg_x, marg_y]:
+        # turn the frame off for both bar graphs
+        marg.axis("off")
+        # remove all tick mark labels for both bar graphs
+        marg.tick_params(
+            axis="x", bottom=False, labelbottom=False, top=False, labeltop=False
+        )
+        marg.tick_params(
+            axis="y", left=False, labelleft=False, right=False, labelright=False
+        )
```

darshan-util/pydarshan/tests/test_heatmap_handling.py
```diff
@@ -0,0 +1,561 @@
...
+        elif ops[0] == "read":
+            assert actual_hmap_data.sum() == 22517726
+        elif ops[0] == "write":
+            assert actual_hmap_data.sum() == 13021781
+
+    elif filepath == "examples/example-logs/ior_hdf5_example.darshan":
+        # make sure the output array is the correct shape
+        assert actual_hmap_data.shape == (4, xbins)
+        # make sure the output data contains identical values
+        assert np.allclose(actual_hmap_data, expected_hmap_data)
```
tylerjereddy, line 553: It is far more helpful to use `assert_allclose()` here because it will actually report the extent of the mismatch/details, rather than using a plain assert on the plain NumPy function boolean return value.

tylerjereddy:
For the vectorization of `get_heatmap_data()`, feature branch `treddy_vectorize_pr_396_version2` on my fork passes the full test suite in this branch and saves around 90 lines of code (partly through removal of the custom bin proportion function). It removes all the Python looping in that function in favor of pandas operations.

The test suite slows down from ~23 to 30 seconds for me locally, but for larger datasets the vectorized code is on the order of ~2 minutes faster:

- running this twice with 200,000 ranks: `time python probe.py` (script below)
  - on branch `treddy_vectorize_pr_396_version2:`
    - python probe.py  12.42s user 0.51s system 104% cpu 12.359 total
    - python probe.py  12.38s user 0.47s system 106% cpu 12.083 total
  - on branch `nawtrey_add_dxt_heatmap_plot`:
    - python probe.py  128.91s user 0.84s system 100% cpu 2:09.05 total
    - python probe.py  128.96s user 0.78s system 100% cpu 2:09.04 total

When there's a large number of unique ranks, the interpreted Python loops are obviously going to be expensive. When the number of ranks is microscopically small (like many of the test cases), any gains from vectorization are probably offset by overhead of using more pandas data structures.

Anyway, that is just a draft, but may provide a useful reference if you get stuck with my original version above.

```python
import pandas as pd
import numpy as np
from numpy.testing import assert_allclose

from darshan.experimental.plots import heatmap_handling

def test_simple():
    size = int(2e5)
    agg_df = pd.DataFrame({'length': [10] * size,
                           'start_time':[0.1] * size,
                           'end_time':[0.9] * size,
                           'rank':list(range(size)),
                           })
    xbins = 10                     

    actual = heatmap_handling.get_heatmap_data(agg_df, xbins)

if __name__ == "__main__":
    test_simple()

```

nawtrey:
Well, the CI failed on 3.7:

`ImportError: cannot import name 'TypedDict' from 'typing'`

I'll try to push a fix before I comment on the new changes/resolve conversations.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
nawtrey: The new version uses this simplification (and passes tests), so I'll resolve this.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
nawtrey: The new version refactors this function removing this repeated code. I'll resolve this.

darshan-util/pydarshan/tests/test_heatmap_handling.py
nawtrey: I replaced all instances of `assert np.allclose()` with `assert_allclose()`, so this can be resolved.

darshan-util/pydarshan/tests/test_heatmap_handling.py
nawtrey: Yes, this was a good point and the newest version uses `assert_allclose()` for these cases. I'll resolve this.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
nawtrey: I was able to update these using the `numpy.typing` module and add in the `if TYPE_CHECKING` block. This should be resolved now.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
nawtrey: The newest version has the documentation updated for all cases like this (says `Sequence` instead of `List`), so this can be resolved.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
nawtrey: There was no point to this, so I removed it in the newest version. I'll resolve this.

nawtrey:
I removed the draft status. 

The newest version fixes the CI issues, fixes various type hinting and documentation things, simplifies some repeated code, and adds in text to display the number of x-axis (time) bins used (I'll post an example of below). 

There is a code coverage drop due to the `if TYPE_CHECKING` blocks, and the version check for importing the `TypedDict`. Since we are testing 3.7-3.9 these lines should all be executed anyways, but locally we will see 98/99% for both modules.

darshan-util/pydarshan/setup.py
```diff
@@ -20,8 +20,8 @@
 # NOTE: The Python C extension is currently only used to automate
 # the build process of binary wheels for distribution via PyPi.
 #
-# If you are building darshan yourself and make libdarshan-util.so 
-# discoverable in the environment by means of LD_LIBRARY_PATH or
```
nawtrey, line 24: Unnecessary formatting changes.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -0,0 +1,444 @@
...
+    # Currently "offset" data is not utilized
+    drop_columns = ["offset"]
+    # create empty dictionary to store
+    # the concatenated read/write dataframes
+    rd_wr_dfs = {}
+    # iterate over each operation
+    for op_key in ops:
+        # mypy can't tell that these keys are in fact
+        # in the ``SegDict``, so just ignore the type
+        seg_key = op_key + "_segments"  # type: ignore
```
nawtrey, line 127: I have it set to ignore the type here because `mypy` would raise an error saying the key wasn't one of the keys specified in `SegDict`. This isn't true, but I don't know another way to get around this.

nawtrey:
Here's an updated plot with the x/time bin count in the bottom right hand corner. It is set to be in the bottom right corner of the figure, and every log I have tested comes out "correct". Users could use this (along with the max time) to calculate an average bandwidth across a bin, which might be useful for some. 

ior_hdf5_example_read_write

nawtrey:
I should also mention that the latest version _does not_ include any of the changes from the branch mentioned [here](https://github.com/darshan-hpc/darshan/pull/396#issuecomment-882115047).

carns:
> Here's an updated plot with the x/time bin count in the bottom right hand corner. It is set to be in the bottom right corner of the figure, and every log I have tested comes out "correct". Users could use this (along with the max time) to calculate an average bandwidth across a bin, which might be useful for some.
> 
> ior_hdf5_example_read_write

This looks great!

tylerjereddy:
> I should also mention that the latest version does not include any of the changes from the branch mentioned here.

Ok, I'll wait for this branch to be fully revised with the idiomatic `pandas` code I provided before I take another look.

tylerjereddy:
After adding in that more concise/performant code, it would also be good to either:

1) open an issue about investigating whether you have test cases for overlapping DXT segments, or
2) just take the time to check if you, then add one in here with expected result if needed

nawtrey:
I've added the changes from @treddy's branch `treddy_vectorize_pr_396_version2`, updated the documentation to reflect the changes, removed unused code, and adapted the plotting code to work in terms of `pandas` dataframes. 

I spent some time looking over the `pandas` variant of `get_heatmap_data` (now `get_heatmap_df`) because my intuition tells me there is a way to simplify even further, but that's probably something we can look at later if it becomes an issue.

nawtrey:
> After adding in that more concise/performant code, it would also be good to either:
> 
> 1. open an issue about investigating whether you have test cases for overlapping (on a single rank) DXT segments, or
> 2. just take the time to check if you do, then add one in here with expected result if needed

`examples/example-logs/ior_hdf5_example.darshan` shows signs of overlapping DXT segments. Here is a subset of the aggregated (read/write) data (sorted chronologically) selected based on whether an entry's end time is greater than the start time of the following entry:
```
    start_time  end_time  rank     op
36    0.029974  0.033101     0  write
43    0.029977  0.031293     1  write
54    0.029977  0.031445     3  write
49    0.029979  0.032376     2  write
44    0.031337  0.180918     1  write
51    0.032957  0.180967     2  write
39    0.034222  0.182035     0  write
45    0.180940  0.181722     1  write
52    0.180990  0.221830     2  write
40    0.221892  0.231445     0  write
47    0.221896  0.231462     1  write
58    0.221898  0.231511     3  write
53    0.221902  0.231529     2  write
5     0.238504  0.238505     0   read
7     0.238519  0.238520     0   read
14    0.238519  0.238520     1   read
15    0.238529  0.238530     1   read
8     0.238570  0.249534     0   read
16    0.238585  0.286340     1   read
32    0.238597  0.251422     3   read
```  
You can see there are entries where multiple ranks are writing data simultaneously. 

Additionally, using the same strategy, I found that `examples/example-logs/dxt.darshan` shows signs of overlap using only a single rank:
```
       start_time     end_time  rank     op
1146     0.276722     0.276734     0   read
1147     0.276730     0.276740     0   read
5611  1466.613887  1466.613906     0   read
7278  1466.613896  1466.613902     0  write
5472  1466.617828  1466.617870     0   read
6175  1466.617864  1466.618047     0  write
6131  1466.622735  1466.622739     0  write
7361  1466.622738  1466.622742     0  write
```

For rows 5611/7278 and 5472/6175 you can even see overlap in read/write segments.

tylerjereddy:
Ok, it is only the single-rank overlap that was of some concern--since MPI frequently directs several ranks to do work simultaneously the case of several ranks working at the same time was basically guaranteed to be covered (even by simple inspection of the output plots).

As for having an intuition that the vectorized code could be simplified--yes, it was only provided as a draft to get rid of the triple or quadruple nested for loop stuff, and I did suggest checking it and the rest of the code here for opportunities for simplification/further vectorization. In any case, if you don't have a concrete plan to simplify it/anything else further I'll do another round of code review as time permits.

tylerjereddy:
This should be getting closer to merge-ready, assuming I don't find anything substantial in my next round of checks.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,425 @@
...
+        [cbar_x0, pos_joint_ax.y0, 0.9, pos_joint_ax.height]
+    )
+
+
+def plot_heatmap(
+    log_path: str,
+    mods: Sequence[str] = ["DXT_POSIX"],
+    ops: Sequence[str] = ["read", "write"],
+    xbins: int = 200,
+) -> Any:
```
tylerjereddy, line 288: since `plot_heatmap()` is likely to be "public"/used externally to some extent, may be worth adding some more error handling for things that currently cause errors in NumPy itself, like negative/invalid bin counts (see below). Can open an issue for that for now if you want; using `hypothesis` a little on the public function APIs may poke a few holes where we could provide helpful errors I suppose.

```python
from darshan.experimental.plots import plot_dxt_heatmap

log_path = './tests/input/sample-dxt-simple.darshan'

plot_dxt_heatmap.plot_heatmap(log_path=log_path,
                              xbins=-17)

plot_dxt_heatmap.plot_heatmap(log_path=log_path,
                              xbins='junk')
```

nawtrey: That's a good point -- I'll open an issue shortly. 

I haven't been too concerned about checking for invalid inputs since whatever framework we go with will likely do most of the input handling for the plotting functions. At least that's what I was anticipating. But with that in mind, it is still likely to help develop the framework things.

tylerjereddy: I think some of the plotting code will be exposed as public for individual use as well.

tylerjereddy: I don't know that the framework will do too much input handling apart from providing sensible defaults most of the time, otherwise if you have 10 plots in the report the command line might get pretty wild specifying stuff for each one.

Also, separating the input handling from each individual function into the framework might get pretty messy (you'd get dissociation between an individual function and its error handling).

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -0,0 +1,425 @@
...
+    ops: a sequence of keys designating which Darshan operations to use for
+    data aggregation. Default is ``["read", "write"]``.
+
+    xbins: the number of x-axis bins to create.
+
+    Returns
+    -------
+
+    jgrid: a ``sns.axisgrid.JointGrid`` object containing a heat
+    map of IO data, marginal bar graphs, and a colobar.
```
tylerjereddy, line 309: can fix this `colobar` typo later

darshan-util/pydarshan/tests/test_heatmap_handling.py
```diff
@@ -0,0 +1,575 @@
...
+        # check that we get the same column names
+        assert_array_equal(
+            actual_df_dict["DXT_POSIX"]["write"].columns,
+            expected_df_dict["DXT_POSIX"]["write"].columns,
+        )
+
+        # verify the returned values are the same
+        assert_array_equal(
+            actual_df_dict["DXT_POSIX"]["write"].values,
+            expected_df_dict["DXT_POSIX"]["write"].values,
```
tylerjereddy, line 236: these are floats being compared--I think I already added a comment about using `assert_allclose()` in these cases, so please do that in a follow-up when I'm done here

darshan-util/pydarshan/tests/test_plot_dxt_heatmap.py
```diff
@@ -0,0 +1,461 @@
+import pytest
+import numpy as np
+from numpy.testing import assert_array_equal, assert_allclose
+import pandas as pd
+import seaborn as sns
+import matplotlib.pyplot as plt
+
+import darshan
+from darshan.experimental.plots import heatmap_handling
+from darshan.experimental.plots import plot_dxt_heatmap
```
tylerjereddy, line 10: minor, for a follow-up: can do the imports from the same module without repeating the path

tylerjereddy:
Ok, the team seems happy with the draft of the plots, the CI is passing, and I tried to be thorough in reviewing this.

So, I'll put it in, thanks Nik.

Obviously this PR had to be big--hopefully follow-up refinements can be more surgical so it is easier to review.

PR #400, opened 2021-06-17
add explicit lustreapi library dependency

### LINKED ISSUE #399: undefine llapi_layout_get_by_xattr error
Darshan 3.3.0 detects and uses the `lustreapi` library to gather Lustre layout information on systems that use the Lustre file systems.  In some dynamically linked cases, the application could fail with an error similar to the following:

```
symbol lookup error:
/soft/perftools/darshan/darshan-3.3.0/lib/libdarshan.so: undefined symbol:
llapi_layout_get_by_xattr
```
The problem is likely caused by insufficient library dependency information in libdarshan.so (despite the pkgconfig file for Darshan requesting the `lustreapi` library).

carns:
Fixes #399

Marked as WIP until trying a few test cases on Theta.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,12 @@
 Darshan Release Change Log
 --------------------------
 
+Darshan-3.3.1
+=============
+* Fix github issue 399: undefined llapi_layout_get_by_xattr function in some
+  scenarios because of incomplete dependency information in Darshan shared
+  library
+
 Darshan-3.3.0
 =============
 * added AutoPerf modules for tracking MPI communication usage (APMPI) and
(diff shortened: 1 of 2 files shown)


carns:
[527784.error.txt](https://github.com/darshan-hpc/darshan/files/6677062/527784.error.txt)
[527784.output.txt](https://github.com/darshan-hpc/darshan/files/6677063/527784.output.txt)
[darshan-lustre-lib.qsub.txt](https://github.com/darshan-hpc/darshan/files/6677064/darshan-lustre-lib.qsub.txt)

Attached job script and output/error files show a test case on Theta executing a dynamically linked mpi-io-test benchmark with either the upstream main branch or this PR's branch LD_PRELOAD'd.  The former produces the symbol error; the latter does not.

This is an inentionally difficult test case with the Darshan module unloaded.  In this scenario, the darshan.so library has to be capable of resolving it's own library dependencies.  There are probably more straightforward ways to trigger the original bug, but if  this one works then the other ones should as well.

shanedsnyder (approved):

PR #404, opened 2021-06-22
MAINT: Removes module `matplotlib.py`

nawtrey:
* Removes module `matplotlib.py` from `darshan/experimental/plots`

* Fixes issue #384


nawtrey:
Labeled as a draft because I'm not sure if simply removing `matplotlib.py` is considered a "fix" for #384. 

I think to call this a fix we should move the #384-related changes from #388 here, which would also cleanup the diff there. I could move `darshan/experimental/plots/plot_access_histogram.py` and `darshan/experimental/plots/plot_opcounts.py` here so we could review them separately, but I'm not sure about the jupyter notebooks. Maybe @jakobluettgau could make a recommendation about those changes. 

Lastly, I think https://github.com/darshan-hpc/darshan/pull/388/files?file-filters%5B%5D=.py&hide-deleted-files=true#r646909096 should be addressed.

jakobluettgau:
I am a little bit confused why we have a second pull request on this now, giving the discussions https://github.com/darshan-hpc/darshan/pull/396#discussion_r649322133. In any case if it was missed, https://github.com/darshan-hpc/darshan/pull/388 addressed the matplotlib.py issues, and just split the files into three while also updating the examples to make sure everything still works.

You could replicate the changes in https://github.com/darshan-hpc/darshan/pull/388 in this pull request of course. I have updated the pull requests title of https://github.com/darshan-hpc/darshan/pull/388 to reflect a little better it's content after the changes such as not conflicting with your DXT plot branch anymore ;)

tylerjereddy (approved):
This seems fine--it addresses the issue and anything that shouldn't be broken, which should be unit tested, doesn't break--locally:

```
python setup.py install
pytest tests/
```

passes the tests on this feature branch

I realize that this does actually break some of the experimental notebooks/modules, so I will aim to merge in Jakob's additional changes next. That said, let's try to regression test pretty much anything that anyone doesn't think should break on this feature branch (I should be able to run `pytest` to see if a PR has breaking changes).

tylerjereddy:
Thanks Nik and Jakob

PR #405, opened 2021-06-23
WIP: Add Darshan config file support

shanedsnyder:
Add the ability for Darshan users to specify a configuration file indicating instrumentation settings to be used at runtime. Currently supported settings are:

- MOD_ENABLE/MOD_DISABLE
  - Indicate whether particular modules should be enabled/disabled at runtime
  - Syntax: MOD_{ENABLE|DISABLE}  <MODULE_CSV>
    - MODULE_CSV is just a list of modules to enable/disable in csv format (e.g., "DXT_POSIX,DXT_MPIIO")
- MAX_RECORDS
  - Override Darshan's default value of maximum records per-module (1024), allowing user to specify how many records modules should be able to instrument at runtime
  - Syntax: MAX_RECORDS  <RECORD_COUNT>  <MODULE_CSV>
- NAME_EXCLUDE/NAME_INCLUDE
  - Set inclusion and exclusion lists for instrumentation record names based on given regular expressions
    - Multiple exclusion/inclusion statements may be provided with the name being excluded/included if it matches any of the corresponding rules given
    - Inclusions override exclusions (i.e., if a name matches an exclusion rule and an inclusion rule, it will be included)
  - Syntax: NAME_{EXCLUDE|INCLUDE}  <NAME_REGEX>  <MODULE_CSV>
    - NAME_REGEX is a POSIX regular expression (i.e., regcomp, regexec interface provided by regex.h) describing a pattern that must be matched for a instrumentation record name to be excluded/included
      - See https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html
- APP_EXCLUDE/INCLUDE
  - Disable/enable Darshan's use based on an application's name
  - Syntax: APP_{EXCLUDE|INCLUDE}  <APP_REGEX>
    - APP_REGEX is a POSIX regular expression that must match the application name (i.e., the executable name/path that Darshan records) 

The Darshan config file is provided to Darshan using an environment variable: `DARSHAN_CONF_PATH`

An example config file demonstrating each of the implemented settings is given below:
```
# comments are allowed using '#'

#   KEY                VALUE            MODULES
NAME_EXCLUDE      ^/home                  *
NAME_INCLUDE      .out$                   *

MOD_ENABLE        DXT_POSIX,DXT_MPIIO

MAX_RECORDS       2048                    H5D

APP_EXCLUDE       (git|ls|sed)
```

Note that users can provide a '*' value for MODULE_CSV in relevant settings to apply to all active Darshan modules.


shanedsnyder:
Following still needs work to resolve WIP:
- [ ] Make sure this implementation supports the DXT trace triggering functionality we recently added. This is all partially supported, but needs more work.
- [ ] Related to above, we should add a capability to exclude/include application ranks based on regular expression. DXT trace triggering supports this, and seems generally useful for other modules, too.
- [ ] Smooth out integration with Darshan's default path exclusions and with the existing DARSHAN_EXCLUDE_DIRS environment variable
- [ ] Add companion environment variables for some/all of these settings so users can provide them without creating a file

shanedsnyder:
878daf8 allows runtime setting of DARSHAN_NAMEMEM env variable to modify Darshan's default per-process memory max for name records.

The following should also be squared away as part of this PR:

- [ ] Config file support for DARSHAN_MODMEM and DARSHAN_NAMEMEM
- [ ] Config file parsing needs to be moved earlier in `darshan_core_init()` to ensure configs are parsed ahead of Darshan's memory allocations
- [ ] Determine what units config/env settings should have for flexibility/consistency

shanedsnyder:
This PR has been replaced by #703

PR #413, opened 2021-07-08
wrapper performance optimizations

carns:
This pull requests adds multiple performance enhancements to the Darshan runtime wrappers.  The optimizations are targeted for the Theta system at the ALCF, but will likely benefit other platforms as well.

These changes reduce function call overhead, reduce locking overhead, uses atomics in place of locks in some locations, and add optional functionality to use the Intel-specific rdtscp timer instruction on platforms where that is more performant than standard timer calls.

The following example shows the impact on a degenerate test case on Theta with 5 million I/O operations to tmpfs on a single process:

```
## NO DARSHAN static:
50000000 fprintf()s of size 1 each in 12.465365 seconds (4011113.939493 ops/s)
50000000 fscanf()s of size 1 each in 29.004269 seconds (1723884.156082 ops/s)
50000000 fread()s of size 1 each in 3.130983 seconds (15969424.993880 ops/s)

## darshan@main, enabled, static:
50000000 fprintf()s of size 1 each in 215.936336 seconds (231549.728577 ops/s)
50000000 fscanf()s of size 1 each in 263.471802 seconds (189773.628985 ops/s)
50000000 fread()s of size 1 each in 200.895207 seconds (248885.977447 ops/s)

## darshan@perf-opt, enabled, static:
50000000 fprintf()s of size 1 each in 27.554754 seconds (1814568.911264 ops/s)
50000000 fscanf()s of size 1 each in 52.301563 seconds (955994.373946 ops/s)
50000000 fread()s of size 1 each in 18.619044 seconds (2685422.507416 ops/s)
```

This PR should be rebased after the autotools changes are merged.  It also needs additional testing, in particular to confirm that the rdtscp instruction produces counter values consistent with the default timer functions.


carns:
Rebased this morning to a) consolidate commits more sensibly and eliminate reverted auxiliary timer library experiment and b) sync up with current origin/main

Will still need to rebase again after #381 .

carns:
Rebased to current origin/main (i.e., to reflect automake changes)

darshan-runtime/lib/darshan.h
```diff
@@ -211,12 +299,66 @@ void *darshan_core_register_record(
...
+
 /* darshan_core_wtime()
  *
  * Returns the elapsed time relative to (roughly) the start of
  * the application.
  */
-double darshan_core_wtime(void);
+static inline double darshan_core_wtime(void)
+{
+    double wtime_offset = __darshan_core ? __darshan_core->wtime_offset : 0;
```
shanedsnyder, line 358: I think we actually need a lock around this, as in `darshan_core_disabled_insrumentation()`?

carns: Yeah, you are right.  I hate to add a lock there for a function called that frequently, though.

Even though it doesn't look as nice, it might be better to store the time offset in a global variable instead of within the __darshan_core pointer structure, so that it is always accessible.

I'll address this and update the PR.

darshan-runtime/lib/darshan-hdf5.c
```diff
@@ -116,17 +116,19 @@ static int my_rank = -1;
...
+    __darshan_disabled ? 0 : darshan_core_wtime();
+
 /*********************************************************
  *        Wrappers for H5F functions of interest         * 
  *********************************************************/
 
 #define H5F_PRE_RECORD() do { \
-    HDF5_LOCK(); \
-    if(!darshan_core_disabled_instrumentation()) { \
+    if(!__darshan_disabled) { \
```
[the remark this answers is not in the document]
carns: Actually, maybe a slightly cleaner variation of the above would be to modify `darshan_core_register_module()` so that it can report failure, and have all of the modules check for success before allocating a module runtime structure.

The safety check to prevent re-initialization after shutdown has started could then be put in `darshan_core_register_module()` so it's not duplicated in N modules.

darshan-runtime/lib/darshan.h
shanedsnyder: That make sense. The net result would be the same in any case, if the global variable `wtime_offset` is initialized to 0 (and I guess set to 0 at shutdown time, too).

shanedsnyder: I took a crack at this via c3cb160.

`wtime_offset` is now a global. It is statically initialized to 0 and set at the same time `__darshan_core` is set in `darshan_core_initialize()`. Ultimately, this just means that if someone calls `darshan_core_wtime()` before Darshan has fully initialized, it will return absolute timestamps (rather than timestamps relative of job start). Otherwise, Darshan has already determined the job `wtime_offset`, and can return timestamps relative to that. That all sounds perfectly reasonable, in that callers shouldn't expect sensible results out of `darshan_core_wtime()` if they haven't ensured Darshan is actually initialized.

darshan-runtime/lib/darshan-hdf5.c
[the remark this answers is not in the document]
carns: After further discussion: the modules are already protected against re-registration, because the core code sets the `__darshan_core` variable to NULL (preventing further registrations)  before invoking any module output routines.

There was,  however, still a possibility of modules updating counters in between the module `_output()` and `_cleanup()` routines, which could possibly cause inconsistent results in the log.  This is fixed now in 9c86c8d.

PR #427, opened 2021-08-01
MAINT, CI: add mypy to CI

### LINKED ISSUE #418: CI: add mypy check to CI
It would be useful to add a basic mypy check to CI for static analysis of the Python type hints; SciPy and other OSS projects have examples of how to ignore errors in config files until they are dealt with.

tylerjereddy:
Fixes #418

* add a basic `mypy.ini` config file, based on
similar type checking config files in other Python
OSS projects (to keep type checking consistent on the
project)

* add type checking to CI matrix

* fix a few minor typing issues in the source
to get `mypy` passing

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -125,7 +125,7 @@ def get_rd_wr_dfs(
     for op_key in ops:
         # mypy can't tell that these keys are in fact
         # in the ``SegDict``, so just ignore the type
-        seg_key = op_key + "_segments"  # type: ignore
+        seg_key = op_key + "_segments"
         # create empty list to store each dataframe
         df_list = []
         # iterate over all records/dictionaries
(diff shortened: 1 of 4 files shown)


darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -125,7 +125,7 @@ def get_rd_wr_dfs(
     for op_key in ops:
         # mypy can't tell that these keys are in fact
         # in the ``SegDict``, so just ignore the type
-        seg_key = op_key + "_segments"  # type: ignore
+        seg_key = op_key + "_segments"
```
tylerjereddy, line 128: @nawtrey had reported a requirement for the `type: ignore` here, but with bleeding edge versions of all dependencies and the `mypy.ini` proposed here, `mypy` considers the presence of the ignore a failure because it isn't needed, so I've removed it.

tylerjereddy:
The `mypy` checks are passing in CI and there are no objections, so merging to keep things moving forward...

Issue #426, opened 2021-07-30
BENCH: add asv benchmarking for pydarshan


tylerjereddy:
There seems to be some interest in performance matters--many large OSS Python projects keep track of performance with more formal [`asv` benchmarks](https://asv.readthedocs.io/en/stable/). It seems a bit early for that just yet I suppose. On the other hand, if arguments about performance do start cropping up regularly it is far better to use that than to just talk about random things you checked locally *ad hoc* with a random script (I've already been guilty of doing that).

The infrastructure to run that may cooperate in some way with a convenient way to access a variety of useful log files in https://github.com/darshan-hpc/darshan/issues/420. Could probably just have a hard-dependency on the log file repository in this case though, since running the benchark suite is typically time consuming and done only occasionally (except for huge projects that run it continuously on a dedicated server).

tylerjereddy:
I'll try to draft a little PR to get the base infrastructure in place, then folks can build up from there.

tylerjereddy:
the basic infrastructure is in place, so I'll close this in favor of more specific issues related to `asv` and using the logs repo with it via PyPI if needed, etc.

PR #428, opened 2021-08-02
WIP, BENCH: start asv infrastructure

tylerjereddy:
Related to #426 -- add some example `asv` benchmark infrastructure that the team can build upon. Includes a few tricky packaging/module-related changes to make the data files available for benchmarking.

On this feature branch the output of `asv run` is below. The overarching purpose is to allow more formal/statistically significant assessment of claims of performance improvement for changes in the Python code base (with i.e., `asv continuous` for comparing two commit hashes/branches).

One thing to be cautious of is changes to the C code underneath--as I understand it, the `setup.py` build doesn't currently link in the `darshan-util` C deps automatically as part of the build procedure--if that were to change, we could also detect perf improvements underneath the hood if they span two hashes and affect Python wrapper performance.

`asv run`

```
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[  0.00%] · For pydarshan commit f35e5012 <treddy_asv_issue_426>:
[  0.00%] ·· Benchmarking virtualenv-py3.8-cffi-numpy-pytest
[ 50.00%] ··· Running (dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs--).
[100.00%] ··· dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs                                                                                                                                                                                                                                              ok
[100.00%] ··· ================================================ ========== ========== ============
              --                                                             xbins               
              ------------------------------------------------ ----------------------------------
                              darshan_logfile                      10        100         1000    
              ================================================ ========== ========== ============
               examples/example-logs/ior_hdf5_example.darshan   381±8ms    559±10ms   1.22±0.01s 
                     examples/example-logs/dxt.darshan          381±20ms   539±20ms   1.21±0.01s 
                   tests/input/sample-dxt-simple.darshan        341±5ms    485±7ms    1.13±0.01s 
              ================================================ ========== ========== ============


```

### DIFF
--- a/darshan-util/pydarshan/benchmarks/benchmarks/dxt_heatmap.py
+++ b/darshan-util/pydarshan/benchmarks/benchmarks/dxt_heatmap.py
@@ -1,0 +1,40 @@
+import os
+import importlib
+
+from darshan.experimental.plots import plot_dxt_heatmap
+# TODO: no good reason pydarshan should have hyphenated module
+# names... for now I hack around it...
+example_logs = importlib.import_module("examples.example-logs")
+from tests.input import test_data_files_dxt
+
+
+class PlotDXTHeatMapSmall:
+    params = [
+        ["examples/example-logs/ior_hdf5_example.darshan",
+         "examples/example-logs/dxt.darshan",
+         "tests/input/sample-dxt-simple.darshan",
+        ],
+        [10, 100, 1000],
+        ]
+    param_names = ['darshan_logfile', 'xbins']
+
+    def setup(self, darshan_logfile, xbins):
+        # darshan example files are found in
+        # various locs in the code base, so shim
+        # around that...
+        filename = os.path.basename(darshan_logfile)
+        if "examples" in darshan_logfile:
+            self.logfile = example_logs.example_data_files_dxt[filename]
+        else:
+            self.logfile = test_data_files_dxt[filename]
+
+    def time_plot_heatmap_builtin_logs(self, darshan_logfile, xbins):
+        # benchmark DXT heatmap plotting for
+        # some log files available in the darshan
+        # repo proper--these are likely to be quite
+        # small for the most part
+        plot_dxt_heatmap.plot_heatmap(
+            log_path=self.logfile,
+            mods=["DXT_POSIX"],
+            ops=["read", "write"],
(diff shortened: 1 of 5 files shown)


darshan-util/pydarshan/tests/input/__init__.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+def installed_path(filename):
+    cwd = os.path.dirname(__file__)
+    return os.path.join(cwd, filename)
```
tylerjereddy, line 5: we could perhaps abstract this to avoid duplication if we stick with this approach to making data files available

darshan-util/pydarshan/setup.py
```diff
@@ -71,11 +71,12 @@
...
     name="darshan",
-    packages=find_packages(include=["darshan*"]),
+    packages=find_packages(include=["darshan*", "examples*", "tests"]),
     setup_requires=setup_requirements,
     test_suite="tests",
     tests_require=test_requirements,
     url='https://www.mcs.anl.gov/research/projects/darshan/',
     version='3.3.0.3',
     zip_safe=False,
+    package_data={"": ["*.darshan"]},
```
tylerjereddy, line 81: current release manager may want to take note of consequences to release process--I believe we may end up with larger packages as a result of making the data files broadly available in installs like this

darshan-util/pydarshan/benchmarks/benchmarks/dxt_heatmap.py
```diff
@@ -0,0 +1,40 @@
+import os
+import importlib
+
+from darshan.experimental.plots import plot_dxt_heatmap
+# TODO: no good reason pydarshan should have hyphenated module
+# names... for now I hack around it...
+example_logs = importlib.import_module("examples.example-logs")
```
tylerjereddy, line 7: we really should just rename this module; to some extent the longer-term plan was to make the logs available in another repo, but their *de facto* availability in the current repo is already being leveraged elsewhere

tylerjereddy:
Alright, CI is passing and I don't see any objections after a few days so I'll put this in so we can build up a few more benchmarks over time, etc.

PR #433, opened 2021-08-03
MAINT: rm extraneous to_df()

tylerjereddy:
* the `to_df()` call in `get_single_df_dict()` does not
appear to serve any purpose, so remove it

* the change was performance agnostic using the heatmap
benchmarks in gh-428 
(mostly dominated by `seaborn` stuff for smaller data files at least)


tylerjereddy:
Ok, CI is passing and trivial cleanup so in it goes.

Issue #421, opened 2021-07-29
CI: add codecov


tylerjereddy:
Adding [codecov](https://about.codecov.io/) to the pydarshan CI may make sense to see the test line coverage on diffs.

It can be flaky, probably best to mimic the settings from other large OSS projects. i.e.,:

- disable the annoying bot comments that show up in PRs
- probably disable codecov ever actually failing the CI, because there are so many cases where it isn't practical to get the coverage threshold
- mostly just a convenient way for reviewers to click through to a coverage report for a PR if the `%` coverage on the diff looks a bit low, to see what is going on

tylerjereddy:
After merging #434, the only remaining steps for this to be closed should be:

- @nawtrey can help @shanedsnyder setup the codecov app for this repo if any help is needed there (since Nik just did the same thing on their fork)
- make sure an example PR looks ok when targeted at `pydarshan-devel`

tylerjereddy:
@shanedsnyder has Nik touched based with you about this? I think the process is basically just to turn on the codecov "app" for the repo.

The main concern I might have is that we only have the config file on `pydarshan-devel`, so there's perhaps a non-zero chance of getting undesired noise on `main`, but hopefully that doesn't happen or could be easily silenced with a config file on `main` if that starts to happen.

shanedsnyder:
I'll have a look and see if I can figure out how to turn it on. Hopefully we can make it not grumpy about lack of coverage outside of pydarshan directory, as well.

shanedsnyder:
I _think_ this is done (I enabled codecov for darshan-hpc at least)? Not exactly sure how to check, looks like you might just need to confirm it works after setting up an action for running this on commits or something?

tylerjereddy:
@shanedsnyder @nawtrey I don't see the reports in the CI matrix yet (I merged gh-467 and rebased gh-397 with no success on getting the reports).

Maybe Nik can confirm that what you did matches what they did on their fork. Sometimes @thomasrockhu on the codecov team is kind enough to take a quick look as well.

The pull requests are registering on the codecov site here: https://app.codecov.io/gh/darshan-hpc/darshan/pulls?page=1&state=open&order=-pullid

Just not seeing the CI matrix entries yet.

nawtrey:
I never messed with any of the GitHub apps to get `codecov` working on my fork. When I opened my test PR's the `codecov` checks automatically popped up, so I never had to go looking for how to turn things on.

[...]

PR #434, opened 2021-08-04
CI: add codecov to CI

nawtrey:
* add basic `.codecov.yml` config file to mute
the bot comments

* add code coverage checking to CI matrix

* Fixes issue #421


nawtrey:
By default the CI should not fail if the coverage drops for a given PR, and the bot comments are turned off. I'm indifferent to which `codecov` features we use, so I just made this fit the requirements @tylerjereddy  requested in #421.

tylerjereddy:
Did you check it on your fork to make sure it works properly? That may be a good first step before asking for permission to activate the hooks here (since you have permission to activate the hooks on your own fork and test there). The codecov settings don't always behave as expected--I know some projects that basically had to "ban" the bot/user.

darshan-util/pydarshan/.codecov.yml
```diff
@@ -0,0 +1,22 @@
...
+
+# default coverage settings
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
+
+# default parser settings
+parsers:
+  gcov:
```
tylerjereddy, line 14: `gcov` is for line coverage of compiled code, but you haven't activated coverage builds of the C source underneath the hood, so I'm not sure I understand the motivation for the Python side of things.

We could discuss that if we want it for the C code eventually, but that's probably another can of worms we don't want to deal with initially.

Also, did you check that `.codecov.yml` is properly picked up when not at the root of the repository proper (it is usually placed there)

darshan-util/pydarshan/.codecov.yml
```diff
@@ -0,0 +1,22 @@
+# Codecov configuration to make it a bit less noisy
+codecov:
+  # don't fail the CI
+  require_ci_to_pass: yes
```
tylerjereddy, line 4: can you explain the reasoning for choosing this setting vs: https://github.com/scipy/scipy/blob/master/codecov.yml#L2

Also, maybe discuss some of the reasons for not adopting the battle-tested settings there or in other OSS projects?

tylerjereddy:
Annotations are annoying too--they should be disabled like here: https://github.com/scipy/scipy/blob/master/codecov.yml#L23

Basically, build up from battle-tested examples..

tylerjereddy:
Another comment here is that you're trying to run `codecov` when you don't even have a coverage report file to begin with--the `pytest` command would have to be altered to generate a coverage report in the appropriate format, and you'll need to make sure the coverage-related deps are installed.

tylerjereddy:
any update here?

tylerjereddy:
Nik has draft version here: https://github.com/nawtrey/darshan/pull/1

Looks closer now

.github/workflows/linux.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           # the test suite is sensitive to
           # relative dir for test file paths
           cd darshan-util/pydarshan
-          pytest
+          pytest --cov-report xml --cov=darshan
```
tylerjereddy, line 47: may want to cover the folder with the test modules in it as well--that can help catch cases with tests that accidentally have the same name for example

nawtrey: I've added `--cov=tests` to the above command. I tested it [on my fork](https://github.com/nawtrey/darshan/pull/2) and `/tests` showed up in the `codecov` GUI

tylerjereddy:
I think the only thing left here is to point to an example on your fork where we can see the "%" change for the "patch" entry like this: 

image

I believe it *should* work if you open a PR on your fork against a branch that has these changes merged in and the "base data" registered with codecov.

nawtrey:
> I think the only thing left here is to point to an example on your fork where we can see the "%" change for the "patch" entry like this:
> 
> image
> 
> I believe it _should_ work if you open a PR on your fork against a branch that has these changes merged in and the "base data" registered with codecov.

I had my forked `pydarshan-devel` branch merged with my original codecov-related [PR](https://github.com/nawtrey/darshan/pull/1), and when I opened the [new PR](https://github.com/nawtrey/darshan/pull/2) it did exhibit that behavior (it gave me the % of diff hit and the target). 

You can view it by clicking the checkmark for the commit just above [this comment](https://github.com/nawtrey/darshan/pull/2#issuecomment-901975301)(the hash in the photo):
Screen Shot 2021-08-20 at 1 15 51 PM

tylerjereddy:
Ok, I'll put this in and comment on the final steps to close the matching issue.

Issue #438, opened 2021-08-06
BUG: file hash uint64's converted to float64's in darshan.report.mod_read_all_records() 


nawtrey:
As mentioned [here](https://github.com/darshan-hpc/darshan/pull/397#issuecomment-893984211), we are seeing some file system hashes that are floats despite the name records (`report.data["name_records"]`) containing integer values. 

I looked into this and apparently the issue is log-dependent, where some logs are generating `pandas` dataframes with ids/hashes of type `int64` and other logs are returning dataframes with ids/hashes of type `float64`. I traced this back to `darshan.report.mod_read_all_records()`, in the `if dtype == pandas` [code block](https://github.com/darshan-hpc/darshan/blob/main/darshan-util/pydarshan/darshan/report.py#L620). Each record (pre-concatenation) has id's that are either `int64`'s or `uint64`'s, but upon concatenation, get `uint64`'s get converted to `float64`'s. This is apparently expected behavior for `pandas` since `np.int64`'s cannot contain the largest values from `uint64`'s. I was able to find some discussion about it [here](https://github.com/pandas-dev/pandas/issues/34356).

This is unwanted behavior since we want to be able to make safe comparisons between the file hashes stored in `report.data["name_records"]` and the counter/fcounter dataframes.  As far as solutions go, I don't think we can force `pandas` to keep the `uint64` type when it concatenates the dataframes, but we *could* use `astype()` to change the `float64`'s back to `uint64`'s after the the concatenation step.

tylerjereddy:
Looks like @jakobluettgau set the hash to `typedef uint64_t darshan_record_id;` in the `api_def_c.py` source--perhaps `uint64_t` could be enforced consistently farther up in the control flow.

Indeed, this seems to do the trick locally so that we never cast to or from floats for the `id` at all (for the `pandas` case):

```diff
index 817f22f..587b780 100644
--- a/darshan-util/pydarshan/darshan/backend/cffi_backend.py
+++ b/darshan-util/pydarshan/darshan/backend/cffi_backend.py
@@ -356,6 +356,7 @@ def log_get_generic_record(log, mod_name, dtype='numpy'):
             })
 
     if dtype == "pandas":
+        rec['id'] = np.uint64(rec['id'])
         df_c = pd.DataFrame(cdict, index=[0])
         df_fc = pd.DataFrame(fcdict, index=[0])
```

[...]

PR #442, opened 2021-08-09
BUG: Correct file hash data types to uint64

nawtrey:
* Enforce data type uint64 for file hashes/ids when reading in
darshan log data as a pandas dataframe

* Add a regression test to ensure the file hashes are uint64's

* Fix issue #438


nawtrey:
Adds the 1-liner @tylerjereddy suggested [here](https://github.com/darshan-hpc/darshan/issues/438#issuecomment-894528913), as well as a regression test that verifies we are maintaining the `uint64` data type.

tylerjereddy (approved):
looks ok, thanks

PR #445, opened 2021-08-10
WIP, ENH: Add initial darshan job summary PDF report

nawtrey:
* Add module, `darshan/cli/summary.py`, to generate
darshan job summary PDF report

**To generate a report:** 
`python -m darshan summary ~/path/to/log.darshan` 

(will save report in current working directory as `report_<filename>.pdf`)


nawtrey:
Initial draft of the PDF report using [`fpdf2`](https://pyfpdf.github.io/fpdf2/index.html). 

Currently report contains some of the original metadata entries as well as the DXT heatmap, and the operation count histogram. Pretty much all of that is flexible, I just picked some of the things that are easy to implement to illustrate what we can do with `fpdf2`. We should also have the ability to create tables using `pandas` and output them as html since `fpdf2` can read it in, so if that is of interest I can definitely look into it. 

I haven't done rigorous documentation, type hints, or tests yet. I'll write those once I know what is nailed down and what isn't.

nawtrey:
Here is a report I generated from `examples/example-logs/ior_hdf5_example.darshan`:

[report_ior_hdf5_example.pdf](https://github.com/darshan-hpc/darshan/files/6964721/report_ior_hdf5_example.pdf)

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,374 @@
...
+        plotting_func=plot_opcounts,
+        func_args=opct_func_args,
+        fig_y=410,
+        fig_w=fig_w,
+        center=True,
+        report=report,
+    )
+    # save the PDF at the desired location with the filename
+    fname_noext = os.path.splitext(log_filename)[0]
+    output_filename = os.getcwd() + f"/report_{fname_noext}.pdf"
```
tylerjereddy, line 369: `os.path.join()` is generally more robust than raw string concatenation

nawtrey: Absolutely, I'll make sure to change that.

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,374 @@
...
+    n_available_chars = chars_per_pt * max_width
+    # the first index is just the first half of the string
+    first_idx = int(n_available_chars / 2)
+    # the second index is the last half minus 3, adjusted for the 3 dots
+    second_idx = first_idx - 3
+    truncated_label = label[:first_idx] + "..." + label[-second_idx:]
+    return truncated_label
+
+
+class PDF(FPDF):
```
tylerjereddy, line 51: Inheritance has been falling out of favor compared to composition for objected-oriented designs. Is `fpdf2` still recommending inheritance?

https://en.wikipedia.org/wiki/Composition_over_inheritance
https://en.wikipedia.org/wiki/Liskov_substitution_principle

tylerjereddy (commented):
It may be useful to also comment on:

- how much extra time the PDF generation/fpdf2 code adds on top of i.e., the heatmap code on its own for small and large log files
- if you can just `pip install` the dependency on Linux and MacOS (are there pre-built binaries?) or if there is more to it/it has to be built from source (how long does that take?)
- what's the intended design for "registering" new plots/tables at different locations (either at a specified location, or at a default appended location if location not specified?)

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: Yeah, in their [tutorial page](https://pyfpdf.github.io/fpdf2/Tutorial.html) they have an example that uses inheritance.

tylerjereddy: Hmm, that's unfortunate, we may still be able to use composition if we want though.

nawtrey:
> * how much extra time the PDF generation/fpdf2 code adds on top of i.e., the heatmap code on its own for small and large log files

At the moment the majority of the run time seems to be generating the figures themselves, but I haven't quantified that observation. The PDF code itself *should* be agnostic to the log size (there aren't any loops over records or anything), but I can't say for sure at the moment. I'll see what I can do to quantify these things. 

> * if you can just `pip install` the dependency on Linux and MacOS (are there pre-built binaries?) or if there is more to it/it has to be built from source (how long does that take?)

`fpdf2` is pip-installable, so `pip install fpdf2` is all that's needed.

> * what's the intended design for "registering" new plots/tables at different locations (either at a specified location, or at a default appended location if location not specified?)

This is something I haven't developed much, so I'm certainly open to ideas. I think having default locations is a good idea, and it would be great to have a way to generate them for new pages. It's probably a little difficult because each figure has its own aspect ratio, so it's hard to say how well a given set of figures are going to fit on a page together without visual inspection. At the moment, we set the figure width and the height is determined automatically, so there *should* be a way to get the aspect ratio, which would help with selecting appropriate locations.

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,374 @@
...
+        )
+        self.cell(
+            w=cell_width, h=cell_height, txt=uid_label, border=True, ln=0, align="L"
+        )
+        self.cell(
+            w=cell_width, h=cell_height, txt=nprocs_label, border=True, ln=0, align="L"
+        )
+        self.cell(
+            w=cell_width, h=cell_height, txt=runtime_label, border=True, ln=0, align="L"
+        )
```
tylerjereddy, line 210: Can't these repeated calls to `self.cell` be simplified to to a `for` loop over the `txt` value to avoid the duplication?

kevin-harms:
You're sure you want to do all the layout by hand rather than using something like HTML so it can handle all the layout?

nawtrey:
> You're sure you want to do all the layout by hand rather than using something like HTML so it can handle all the layout?

I think it would be nice to have HTML handle the layout, I just think that approach requires us to start from an HTML report and work towards a PDF report, which would rely on a 3rd party tool being able to convert our HTML report properly. This approach is more PDF-first because there is a bit of a push to get the PDF report off the ground.

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: I had issues with using a for-loop initially, but I think I've figured out what I was doing wrong. So yes, I should be able to make this a for-loop.

nawtrey: Latest version uses a for-loop here.

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: Current version uses `os.path.join()`.

nawtrey:
I wasn't expecting the CI to run yet, but I guess it's not surprising the `mypy` check failed since I have yet to add `fpdf2` to the dependencies.

nawtrey:
I'm wondering how we want to handle non-DXT logs here since the only figure we the heatmap relies on `DXT_POSIX` being present? I thought the idea was to have the IO heatmap as the first figure in every report, but we will not be adding `POSIX` support to it. So I'm not sure what, if anything, we will be putting there for non-DXT tracing logs.

tylerjereddy:
Maybe just excluding it when there's no DXT data available, but good to get feedback from Shane and Phil on this matter. I'm pretty sure Phil was interested in designing initially for the case where we assume that DXT data IS available, and then take it from there.

As I understood it, a subset of the table data from the old reports will be moved in as well, and a few other plot types should still be valid.

Having the code designed to gracefully handle missing parts of the layout seems like a good idea perhaps. Whether it is acceptable for that to be in the form of empty placeholder "frames" (might be easier, but not as pretty) or reflowing the document to fill in the gaps (may take more engineering) is perhaps up to Shane/Phil.

shanedsnyder:
> > Another question might be--why is everything we've been working on so far in the `experimental` namespace, but this is not?
> 
> I was following the structure used in #386, which is also not in the `experimental` namespace. Maybe @shanedsnyder can comment on the ideal location for this module, I have no preference.

I think it's fine to move stuff out of experimental, specifically any bits and pieces we want to make sure we're fully supporting as part of report generation tool, and other things. This stuff is reasonably sandboxed in branches that aren't `main` so it's not a big deal -- we could always shuffle things around prior to a merge if thoughts change on what we want to fully support vs what's still experimental.

shanedsnyder:
> Maybe just excluding it when there's no DXT data available, but good to get feedback from Shane and Phil on this matter. I'm pretty sure Phil was interested in designing initially for the case where we assume that DXT data IS available, and then take it from there.
> 
> As I understood it, a subset of the table data from the old reports will be moved in as well, and a few other plot types should still be valid.
> 
> Having the code designed to gracefully handle missing parts of the layout seems like a good idea perhaps. Whether it is acceptable for that to be in the form of empty placeholder "frames" (might be easier, but not as pretty) or reflowing the document to fill in the gaps (may take more engineering) is perhaps up to Shane/Phil.

Yeah, if no DXT information is available I think we just want to avoid displaying this graph for now. We have ideas on how to generate a graph like this if tracing isn't enabled (one using existing coarse-grained counters in the POSIX module, the other would require modifications to what Darshan captures at runtime to capture some sort of condensed representation of the trace data), but we still need to think more about that. Making progress on other elements of the report and arriving at a good first draft are a little higher priority probably.

Generally speaking, the report needs to be somewhat resilient to missing information. Not only is the DXT data not always available, but we also have HDF5 and AutoPerf data which are not always available either. It would be great if the document could gracefully handle missing components without requiring lots of filler white space, but it looks like there's not really an elegant way to handle that in these PDF-oriented tools (e.g., it would require some fragile modifications to locations of all components that follow in the report). How much work would it be to support something that flexible using fpdf (or reportlab or whatever)?

PR #447, opened 2021-08-11
ENH: Optimize cffi_backend.log_get_generic_record()

nawtrey:
* Performs several small changes to `log_get_generic_record`
to improve the performance of the `pandas` dataframe retrieval.
One of these optimizations is allowing the counter/fcounter
column names to be fed into the function instead of
generating them for each record.

* Add regression test `test_log_get_generic_record`
to `test_cffi_misc.py`

* Fix issue #440

PR #453, opened 2021-08-19
MAINT: cleanups after gh-447

tylerjereddy:
* make the cleanups I requested for gh-447

* don't use `np.asarray()` when the argument cannot
possibly be an array that would get passed through
unchanged; in such cases, simply use `np.array()` since
a copy of the list data must be made regardless (and there
can be a microscopic performance benefit for small lists
on top of improved clarity of intention)

* fix the incorrect order of arguments used for
`assert_allclose()` and `assert_array_equal()`


nawtrey:
Looks good to me, thanks for the cleanup.

PR #471, opened 2021-09-02
TST, MAINT: logs repo resources

### LINKED ISSUE #460: TST, MAINT: integrate the logs repo for testing purposes
This is something that would be useful to do to give us easy access to data/test files in the logs repo. Approximate steps for the integration might be:

- [x] add a `conftest.py` file with the import hooks described in https://github.com/darshan-hpc/darshan-logs/pull/2 (so that `has_log_repo` is either `True` or `False` and tests get executed/skipped accordingly)
- [ ] add an example test or two that leverages one or more logs from the logs repo to demonstrate the tests executing/skipping as appropriate, depending on whether the logs package is installed in the environment; some of the logs in the logs repo may not work with an installation of the current version of `pydarshan-devel` branch, depending on the config options that were used to build `darshan-util` I think (see above-linked PR for rough draft of example test)
- [ ] adjust the CI such that it tests both cases--probably most CI matrix entries should have the darshan-logs project installed, but at least one CI matrix entry should demonstrate that the test suite correctly skips tests that depend on the logs repo when its package is not installed; sample command to install the package: `python -m pip install git+https://github.com/darshan-hpc/darshan-logs.git@main`

### LINKED ISSUE #464: TST, MAINT: provide consistent interface to all darshan logs in the logs repo
Requested by @shanedsnyder in: https://github.com/darshan-hpc/darshan-logs/pull/2

In Python `3.9.x` we can fairly easily get an iterator over all the log files in the logs repo with this kind of thing:

```python
    p = importlib.resources.files('darshan_logs')
    for log_file in p.glob('**/*.darshan'):
        with importlib.resources.as_file(log_file) as darshan_log:
            report = darshan.DarshanReport(str(darshan_log))
```

However, there is a bit more work to it with older versions of Python `3.x` -- see what they have access to here: https://docs.python.org/3/library/importlib.html#module-importlib.resources

It may be useful for us to have a shim that provides a consistent interface for Python `3.7-3.9` or so, with the simple code path for `>=3.9`. This might exist as a utility function or pytest fixture, so we can conviently have i.e., tests that run on all log files vs. i.e., selected log files as demonstrated in the cross-linked PR.

tylerjereddy:
Fixes #464 
Fixes #460

* provide a `pytest` fixture `log_repo_files()` that
returns a list of absolute paths for *all* files in the
`darshan_logs` package, and an empty list when this package
is absent

* add a demo test that leverages the new `log_repo_files()`
fixture, to make it easier for others to replicate this
kind of test that leverages all files in the logs repo; note
that some of the logs are large--overall test suite time jumps
from to 27.21s to 38.49s on a single core locally

* adjust CI such that it:
1) installs `importlib_resources` for testing on appropriate
Python versions
2) installs the `darshan_logs` project in some, but not all,
matrix entries, to ensure that tests can be skipped when this
package is missing

### DIFF
--- a/darshan-util/pydarshan/tests/conftest.py
+++ b/darshan-util/pydarshan/tests/conftest.py
@@ -25,10 +25,7 @@ def pytest_configure():
 def log_repo_files():
     # provide a convenient way to access the list
     # of all *.darshan log files in the logs repo,
-    # returning an empty list in the absence of the
-    # logs repo package and a list of absolute file
-    # paths to the logs otherwise
-    if not has_log_repo:
-        return []
+    # returning a list of absolute file paths to
+    # the logs
     p = importlib_resources.files('darshan_logs')
     return [str(p) for p in p.glob('**/*.darshan')]
--- a/darshan-util/pydarshan/tests/conftest.py
+++ b/darshan-util/pydarshan/tests/conftest.py
@@ -1,3 +1,15 @@
+import platform
+python_version = platform.python_version_tuple()
+
+# shim for convenient Python 3.9 importlib.resources
+# interface
+if int(python_version[1]) < 9 and int(python_version[0]) == 3:
+    import importlib_resources
+else:
+    # see: https://github.com/python/mypy/issues/1153
+    import importlib.resources as importlib_resources # type: ignore
+
+
 import pytest
 
 try:
@@ -8,3 +20,15 @@ except ImportError:
 
 def pytest_configure():
     pytest.has_log_repo = has_log_repo
+
+@pytest.fixture
+def log_repo_files():
+    # provide a convenient way to access the list
+    # of all *.darshan log files in the logs repo,
+    # returning an empty list in the absence of the
+    # logs repo package and a list of absolute file
+    # paths to the logs otherwise
+    if not has_log_repo:
+        return []
+    p = importlib_resources.files('darshan_logs')
+    return [str(p) for p in p.glob('**/*.darshan')]
--- a/darshan-util/pydarshan/tests/test_report.py
+++ b/darshan-util/pydarshan/tests/test_report.py
@@ -29,20 +29,6 @@ def test_jobid_type_all_logs_repo_files(log_repo_files):
     # this is primarily intended as a demonstration of looping
     # through all logs repo files in a test
     for log_filepath in log_repo_files:
-        # darshan_logs/apmpi/darshan-apmpi-2nodes-64mpi.darshan
-        # causes an issue so we skip it here:
-        # 
-        # Error: incompatible darshan file.
-        # Error: expected version 3.21, but got 
-        # <!
-        # Error: darshan_log_open failed to read darshan log file header: Undefined error: 0.
-        if 'darshan-apmpi-2nodes-64mpi' in log_filepath:
-            continue
-        # we also need to skip apmpi data for now to avoid
-        # a segfault on pydarshan-devel with current standard
-        # config/setup options (see the CI for example)
-        if 'apmpi_apxc' in log_filepath:
-            continue
         report = darshan.DarshanReport(log_filepath)
         assert isinstance(report.metadata['job']['jobid'], int)
 
--- a/darshan-util/pydarshan/tests/test_report.py
+++ b/darshan-util/pydarshan/tests/test_report.py
@@ -21,6 +21,32 @@ def response():
     pass
 
 
+@pytest.mark.skipif(not pytest.has_log_repo, # type: ignore
+                    reason="missing darshan_logs")
+def test_jobid_type_all_logs_repo_files(log_repo_files):
+    # test for the expected jobid type in each of the
+    # log files in the darshan_logs package;
+    # this is primarily intended as a demonstration of looping
+    # through all logs repo files in a test
+    for log_filepath in log_repo_files:
+        # darshan_logs/apmpi/darshan-apmpi-2nodes-64mpi.darshan
+        # causes an issue so we skip it here:
+        # 
+        # Error: incompatible darshan file.
+        # Error: expected version 3.21, but got 
+        # <!
+        # Error: darshan_log_open failed to read darshan log file header: Undefined error: 0.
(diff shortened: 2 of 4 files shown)


darshan-util/pydarshan/tests/test_report.py
```diff
@@ -21,6 +21,32 @@ def response():
...
+        # Error: expected version 3.21, but got 
+        # <!
+        # Error: darshan_log_open failed to read darshan log file header: Undefined error: 0.
+        if 'darshan-apmpi-2nodes-64mpi' in log_filepath:
+            continue
+        # we also need to skip apmpi data for now to avoid
+        # a segfault on pydarshan-devel with current standard
+        # config/setup options (see the CI for example)
+        if 'apmpi_apxc' in log_filepath:
+            continue
```
tylerjereddy, line 45: @shanedsnyder the notes above re: some of the logs in the repo may be of interest--you should be able to see the steps we're using in CI to build and then reproduce the errors locally on `pydarshan-devel` branch.

My assumption here is that some of these are either newer features on `main` and/or require extra `config` stage build flags for support (feel free to suggest those). For now, I'm not too worried, but probably smoother if we can just flush through all logs unconditionally when needed for testing.

tylerjereddy: (I'm not sure if turning on `apmpi` works super well on `pydarshan-devel` branch)

darshan-util/pydarshan/tests/conftest.py
```diff
@@ -8,3 +20,15 @@
...
+
+@pytest.fixture
+def log_repo_files():
+    # provide a convenient way to access the list
+    # of all *.darshan log files in the logs repo,
+    # returning an empty list in the absence of the
+    # logs repo package and a list of absolute file
+    # paths to the logs otherwise
+    if not has_log_repo:
+        return []
```
tylerjereddy, line 32: honestly, it might be better if this case is not handled at all, and we simply get an error if someone uses the log_repo_files fixture without the appropriate `pytest` skip guards

darshan-util/pydarshan/tests/test_report.py
shanedsnyder: @tylerjereddy one thing we should consider doing soon is just rebasing `pydarshan-devel` branch back onto our `main` branch? I don't think much of anything has changed related to pydarshan on `main`, so I'm hoping it wouldn't be too difficult.

I can help look into what exactly is causing the error in general, but maybe the rebase would help. We should definitely have a more graceful way of handling logs that we can't understand with specific versions of pydarshan, though. As we add logs to that repo with new format versions, it's probably important to make sure pydarshan tests don't fail on older versions. We might need to make sure we have appropriate checks for this in place, I seem to remember there being some version checking logic, but maybe it doesn't account for log format versions yet.

darshan-util/pydarshan/tests/conftest.py
tylerjereddy: I've made this adjustment/revision

darshan-util/pydarshan/tests/test_report.py
tylerjereddy: @shanedsnyder Ok, I pushed in a commit to turn on autoperf submodule stuff in the Linux CI and disable the test skips here. At the moment, the DXT heatmap tests are failing for some strange reason, which is preventing me from probing this directly in CI.

@nawtrey can you help us figure out why the DXT heatmap tests are failing here? we haven't touched that code

nawtrey: Yeah, I can take a look at that.

tylerjereddy: I wonder if it is dependency versions or something--seemed "ok" when I did a quick Python 3.9 check on Linux this morning. Otherwise, something really strange

nawtrey: Yeah, I think it is something like that. I did a clean install on a Linux machine and it was throwing errors on this branch as well as `pydarshan-devel`. I tried to find the dependency that could be reverted, but I haven't been able to pin it down yet. I assumed it would be `seaborn`, but updating to the latest version on my Mac install doesn't cause the tests to fail.

tylerjereddy: It is `matplotlib`:

`python -m pip install matplotlib==3.5b1`

causes the failure for me locally, and that is the version that gets installed in a clean environment with `python setup.py install`

We really should be using `pyproject.toml` to controll installs with `python -m pip install .`, but for now we can't use that because of another issue.

In any case, probably good to investigate what has changed in the new version of `matplotlib`, and if it isn't an obvious bug, then I suppose we'll want to support the old and new approaches to whatever changed.

tylerjereddy: For now, I'm probably going to pin `matplotlib` and open another issue to investigate this though.

nawtrey: Okay, that's what I'm seeing on both of my installations, just to confirm.

tylerjereddy: @shanedsnyder Circling back to the original problem, with the other stuff out of the way now, the CI here clearly shows the log file reading problem for a Linux matrix entry, even with the autoperf stuff turned on.

I'll try to revisit rebasing to see if that alleviates the issue now that things are slightly less chaotic with the CI/MacOS.

tylerjereddy: So, the problem is clearly described in: https://github.com/darshan-hpc/darshan-logs/pull/5

We're dealing with an html file masquerading as a darshan binary log file in the logs repo. Once that is solved, I suspect we're actually going to be good to go here.

tylerjereddy:
rebase/force push to update, and also pull in the new logs repo install which should no longer have the masquerading html file

darshan-util/pydarshan/tests/test_report.py
tylerjereddy: this is resolved now--there is no longer an html file in place of a log file in the logs repo

PR #474, opened 2021-09-08
MAINT, CI: more actions labels

tylerjereddy:
* a few more labels for the label bot
because it looks like GitHub actions treats
its own config paths separately per the bot
activity in gh-471


tylerjereddy:
Ok, trivial adjustment for CI/GitHub infrastructure and CI is passing, so let's see if this makes the auto-labeling a bit better.

Issue #457, opened 2021-08-26
TST: improve coverage of DarshanReport.mod_read_all_dxt_records()


tylerjereddy:
@Lcjacobpd This may be a reasonable "onboarding issue," that still has a concrete value to the project, before ramping up to something bigger/messier.

I noticed in the coverage report produced by `pytest --cov=darshan --cov-report=html --cov-report=term-missing` that the `DarshanReport.mod_read_all_dxt_records()` method has a bit of uncovered code:

![image](https://user-images.githubusercontent.com/7903078/131013335-eb42499c-43ba-48e1-ab85-e0efd2653e74.png)

It looks like the code that is not covered is related to logging warning messages for unsupported/missing modules. One idea to cover it might be to write a test that opens a darhan log file that has DXT data in it (there are some in the repo proper: `find . -name "*dxt*.darshan"`). Some examples of producing the initial `darshan.DarshanReport` objects in tests are availble in `tests/test_report.py`. As for actually capturing and verifying that the text of the warning message is appropriate, the `pytest` [`caplog` fixture](https://docs.pytest.org/en/6.2.x/logging.html#caplog-fixture) looks promising.

Note that we do actually use this method in production code (like @nawtrey DXT heatmap code), so it is useful to test a bit.

tylerjereddy:
Fixed by the cross-listed PR

PR #478, opened 2021-09-09
Report.py Coverage

Lcjacobpd:
Added two new tests to tests/test_report.py to improve code
coverage. Resolves Issue #457

### DIFF
--- a/darshan-util/pydarshan/tests/test_report.py
+++ b/darshan-util/pydarshan/tests/test_report.py
@@ -63,6 +63,20 @@ def test_unsupported_record_load(caplog, unsupported_record):
         assert unsupported_record in record.message
 
 
+def test_invalid_dxt_mod():
+    """Invalid dxt module case"""
+
+    report = darshan.DarshanReport("tests/input/sample-dxt-simple.darshan")
+
+    report.mod_read_all_dxt_records("BUGGER")  # Intentional error.
+
+def test_unsupported_dxt_mod():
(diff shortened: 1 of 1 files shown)


darshan-util/pydarshan/tests/test_report.py
```diff
@@ -63,6 +63,20 @@ def test_unsupported_record_load(caplog, unsupported_record):
...
+    report = darshan.DarshanReport("tests/input/sample-dxt-simple.darshan")
+
+    report.mod_read_all_dxt_records("BUGGER")  # Intentional error.
+
+def test_unsupported_dxt_mod():
+    """Valid module but unsupported case"""
+
+    report = darshan.DarshanReport("tests/input/sample-dxt-simple.darshan")
+
+    report.mod_read_all_dxt_records("STDIO")
```
tylerjereddy, line 78: A few comments:

- you haven't actually checked that the `logger` module issues the appropriate warning text (this is discussed in the matching issue)
- you could reduce some duplication and make failures easier to identify by using [parametrization](https://docs.pytest.org/en/6.2.x/parametrize.html)
- it looks like this only covers half of the code mentioned in the matching issue (checking locally on this branch with: `pytest --cov=darshan --cov-report=html`:
image

Codecov seems to concur as well:
image

tylerjereddy:
LGTM now so I'll put it in, thanks.

You can see the lines noted as missing coverage in the issue as added to coverage in the new codecov report as well:

image

PR #482, opened 2021-09-10
MAINT: rebase pydarshan-devel on main

tylerjereddy:
* rebase `pydarshan-devel` on `main` to reduce confusion surrounding issues opening log files (see gh-471)

* @shanedsnyder you may want to skim this over a little to make sure that the CI changes reflect the changes you'd expect to the build process with the latest build system on `main` (I had to add `automake` install to MacOS for the rebased branch here); do we want to squash the commits when this is ready to merge? probably a little messy regardless, but I guess that could make sense

* I still can't open `apmpi/darshan-apmpi-2nodes-64mpi.darshan` in the logs repo on my Mac with this rebased feature branch


tylerjereddy:
@shanedsnyder ok, so the reason the Python bindings can't read the log file is because it is actually an HTML file masquerading as a darshan binary log: https://github.com/darshan-hpc/darshan-logs/pull/5

So, I think once that is removed from that repo and/or replaced with the genuine log file proper, we should be in pretty good shape--I'd suggest we rebase regardless since I've gone through the trouble of dealing with the merge conflicts and to keep the divergence between branches manageable.

tylerjereddy:
Ok, no comments in the last 10 days and CI is passing so I'll put this in so that I don't have to deal with more merge conflicts when other things get merged.

Other devs on the Python project should take note of the change in build procedure moving forward (new build system from `main`).

PR #485, opened 2021-09-14
WIP: Adds simple darshan job summary template

### LINKED ISSUE #462: generate a simple darshan job summary template with header + footer + log details
After deciding on the templating engine in #461, we should come up with a simple HTML template for darshan job summaries that only includes some high-level job details. This can serve as a template starting point that we can then build off of when implementing other sections of the report we have targeted for V1 of these tools.

The Python code for generating job summaries will need to extract job metadata from the log to pull out some simple parameters that will be used to quickly provide context about the report in question.

Header is probably something really simple like "app-name (date)", like `ior (11/11/21)`, with some horizontal lines to separate it from following sections of the report.

Footer maybe just something indicating what version of PyDarshan generated the report (e.g., `Job summary generated using PyDarshan 3.3.0.2.`), mostly for us to quickly inspect reports from users to know which versions they are using and things like that. 

The first section of the report should summarize important metadata about the log so users can quickly orient themselves and understand what job the report is for and what the log contains:

image

One new thing we haven't done before is the details about the different modules present in the log at the bottom of the details listed above. We think it will be helpful for users to understand straight away what type of module data is contained in the log, how much data each module produced, and whether any modules ran out of memory at runtime (the warning sign).

nawtrey:
* Adds initial framework for building html darshan job summary
reports which contain metadata and module summary tables

* Fixes #462


nawtrey:
First draft for a simple HTML report containing the log metadata, module summary table, and a basic header/title/footer. I know @shanedsnyder wanted to get an idea of what the templating looks like for something simple, so this should be helpful. 

At the moment, I would like to find a way to move the CSS out of the template and find a better way to register sections/figures/tables than simply adding entries to a dictionary. I think building a class for this is probably a better route, but I'm open to recommendations. Some of those things can be handled in the template itself as well.

To generate a report:
---------------------
`python -m darshan summary /path/to/log.darshan`

Here is a sample report (converted to pdf): [ior_hdf5_example_report.pdf](https://github.com/darshan-hpc/darshan/files/7164179/ior_hdf5_example_report.pdf)

tylerjereddy (commented):
Seems fairly straightforward at first glance--I'll let the Argonne folks comment on the design aspects, since UX and frontend aspects are not really my area.

I did add one comment about consistency with our other ongoing `pydarshan` work, and the `mypy` failures in CI are real though not a priority until Argonne folks provide feedback anyway.

tylerjereddy:
> Here is a sample report (converted to pdf): ior_hdf5_example_report.pdf

I guess we might as well just start sharing `html` at this point? I think part of the idea is that it should be easy to share the single html file, so we'd be more likely to catch issues if we just share that regularly.

nawtrey:
> > Here is a sample report (converted to pdf): ior_hdf5_example_report.pdf
> 
> I guess we might as well just start sharing `html` at this point? I think part of the idea is that it should be easy to share the single html file, so we'd be more likely to catch issues if we just share that regularly.

I tried to drop an html file originally but GitHub says it doesn't support that file type. I updated the comment to include a dropdown box with the raw HTML in it.

tylerjereddy:
putting in a tarball is probably fine

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,213 @@
+import os
+import argparse
+import datetime
+import importlib.resources as pkg_resources
```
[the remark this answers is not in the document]
nawtrey: I wasn't aware of those things, it was just a result of starting from a stripped down version of #386. I don't see a reason to keep it as `pkg_resources` so I'll just change it to `importlib_resources`.

nawtrey:
> putting in a tarball is probably fine:
> 
> image

Yeah, that was a better idea. I've updated it to an archive with 3 example reports I generated.

tylerjereddy:
It is also customary to add a comment in the original PR description that mentions the issue that the PR is meant to address, unless no such issue exists.

In this case, it seems fairly clear to me that this is related to gh-462, which was opened by Shane, and that connection would also helps us label this PR as a milestone item for reviewer priority filtering etc. So, I'd suggest:

- edit the PR description above to mention the issue(s) this is meant to fix
- maybe open an issue to create a PR template so that folks can be reminded of some basic things to add when creating a PR (this is quite common in various projects)

tylerjereddy:
In that issue, Shane also mentioned some horizontal line(s) related to header I don't see here--I'll leave that for you and Shane to discuss I guess:

> Header is probably something really simple like "app-name (date)", like ior (11/11/21), with some horizontal lines to separate it from following sections of the report.

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,213 @@
...
+            # integer label it anonymized
+            command = "Anonymized"
+    else:
+        # if there is no executable
+        # label as not available
+        command = "N/A"
+    return command
+
+
+def get_header(report: Any):
```
tylerjereddy, line 72: for all the functions labeled with `report: Any`

I checked locally:

```python
reveal_type(report)
reveal_type(pd.DataFrame) # control for Any
```

`mypy` puts a more specific type on the `report` and labels a `pandas` dataframe as `Any`, so seems like we could be more specific here

```
darshan/cli/summary.py:189: note: Revealed type is "darshan.report.DarshanReport"
darshan/cli/summary.py:190: note: Revealed type is "Any"

Comment (nawtrey) [reply darshan-util/pydarshan/darshan/cli/summary.py]: The latest version has changes all instances of `report: Any` to `report: darshan.report.DarshanReport`, so I believe this can be resolved.

**Comment by nawtrey** [comment]
The latest version has two substantial changes from the original:
1.) The summary report data handling is done inside of the class `SummaryReport` instead of using a Python dictionary
2.) The CSS is moved out of the template and into a new module `darshan.css`
3.) Made various style changes so the report is easier to print as a PDF (mainly the fonts were huge before, so things were unreasonably large outside of a browser)

I'll update the archive of example files shortly.

File: darshan-util/pydarshan/darshan/cli/summary.py [hunk 3]
```diff
@@ -0,0 +1,226 @@
...
+import argparse
+import datetime
+import importlib.resources as importlib_resources
+from typing import Any, Union
+
+import pandas as pd
+from mako.template import Template
+
+import darshan
+import darshan.css
```
tylerjereddy, line 11: I wonder if we really want to be adding two whole top-level namespaces to the `darshan` package when they are both thematically related to "html/template" stuff and could be organized into a single namespace at most, if not encapsulated in `cli` or whatever.

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,226 @@
...
+        else:
+            runtime = str(runtime_val)
+        return runtime
+
+    def get_report(self):
+        """
+        Retrieves the Darshan report used to generate the summary report.
+        """
+        report = darshan.DarshanReport(self.log_path, read_all=True, dtype="pandas")
+        self.report = report
```
tylerjereddy, line 77: What's the advantage of having `self.report = darshan.DarshanReport` one-liner encapsulated in a function/method like this? In particular, `self.report = report` seems a bit silly here perhaps, there's just an extra variable we don't need, but probably an entire function which is basically just `DarshanReport`.

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,226 @@
...
+                    command = report.metadata["exe"]
+            else:
+                # if it can be converted to an
+                # integer label it anonymized
+                command = "Anonymized"
+        else:
+            # if there is no executable
+            # label as not available
+            command = "N/A"
+        return command
```
tylerjereddy, line 53: This control flow certainly feels a bit awkward for what it does. Noting cleaner with `str.isdigit()`?

Seems like something like this would be a fair bit less gruesome if it works:

```python
cmd = report.metadata["exe"]
if not cmd:
    cmd = "N/A"
elif cmd.isdigit():
    cmd = "Anonymized"
return cmd
```

tylerjereddy:
One question for Argonne reviewers when they look at this is if they are confident enough to start cementing with basic regression tests (for production of correct strings, etc.), keeping in mind any uncertainty related to injection of plots/"sections," or if they'd prefer to delay (both approaches have drawbacks).

tylerjereddy:
Another thing for reviewers to check here is whether the `.css` and `.html` "assets" are properly distributed when you install this feature branch from a wheel--for example, my guess would be that without some adjustments to the Python setup/distribution code, you may end up with something missing because Python packaging systems will typically try to leave out extra data that isn't overtly Python-related when packaging.

This kind of thing came up a few times recently--being handled for assets in the logs repo and also noted in comments in Kevin's original template PR.

A useful test (to be added here when it is ready or described in sufficient detail in an issue) might be to:
- have a new CI matrix entry that tests "command line job summary generation from a wheel install"
- it may be most suitable to ensure this can be done inside a conventional regression test without having to use a subprocess--for example, that it is possible to call `main()` with appropriate arguments on each of the log files in the logs repo and make sure an `.html` artifact gets generated as expected in a temporary test directory

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: I think it just made the `__init__()` a bit cleaner, but I agree it's not necessary. I can change all of the class methods to take the report as input instead.

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: I remember trying something like this originally and I thought some cases weren't working. I'll see if I can clean it up though since it is a bit ugly.

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: I don't have a preference here, I ended up splitting them up but I could have gone the other way just as easily. If we want to throw them all in 1 namespace that's fine with me.

darshan-util/pydarshan/darshan/cli/summary.py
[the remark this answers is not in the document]
tylerjereddy: > I wasn't aware of those things

Ok, try to keep an eye on changes other folks are doing as well.

darshan-util/pydarshan/darshan/cli/summary.py
tylerjereddy: Hmm, not sure that's what I was asking for--was just thinking this method could basically be removed and the code used directly.

darshan-util/pydarshan/darshan/cli/summary.py
tylerjereddy: (unit tests are useful for demonstrating this kind of stuff if it is tricky--if you can show a case where it fails, etc.)

tylerjereddy:
FWIW, I don't think a Python project should have a separate top-level namespace for each type of non-Python asset (CSS, HTML, etc.) it stores.

darshan-util/pydarshan/darshan/cli/summary.py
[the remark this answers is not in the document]
nawtrey: I've made the change discussed here, I'll resolve this.

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: Oh okay, I thought you were saying it wasn't worth assigning to `self.report` at all (I think the argument could be made) but yeah I can just move the code.

tylerjereddy: It looks like that's already the case anyway--class (static) methods seem to accept `report`?

tylerjereddy: `self.report` seems ok since the instance methods use it; the intermediate `report` seems harder to justify I think.

nawtrey: Yeah, the static methods accept the `DarshanReport` as input but the class instance methods do not, despite most of them using it one way or another.

carns:
Out of the 3 examples in the log file it looks like maybe sample_report.html still needs to be updated, but otherwise I think this is a great start.  I like that it also indicates the PyDarshan version as well, since each of the Darshan library version, Darshan log version, and PyDarshan version could influence what shows up in the report.

For my vote: I'm happy to proceed with this (re: CI implications etc.) and build from there.

I don't have enough expertise to comment on the discussions related to how artifacts are distributed.

tylerjereddy:
A rebase may be good on the next round of revisions as well, just so I can see that things are relatively "ok" with the newer build system now merged in from `main`.

tylerjereddy:
That sounds like a vote for adding some basic tests as well--whether they are done here or issues are opened for basic tests and the more detailed testing noted above I'm not super particular, but certainly should be done eventually (0 % coverage at the moment).

The build system/packaging stuff could be deferred to an issue as well, as long as the issue is written in sufficient detail to describe the problem accurately.

darshan-util/pydarshan/mypy.ini
```diff
@@ -30,12 +30,11 @@ ignore_missing_imports = True
 [mypy-darshan_logs]
 ignore_missing_imports = True
 
-[mypy-importlib_resources]
+[mypy-mako.*]
```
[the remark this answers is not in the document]
nawtrey: Okay, my second attempt at rebasing fixed this, so I'll resolve this comment.

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: All report-related things (`base.html`, `style.css` and `summary.py`) have been moved into `darshan.cli`

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: `self.report = darshan.DarshanReport()` has been moved into `__init__()` as a one-liner and the method has been removed.

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: I ran this through every log I have access to and I didn't see any strange behavior, so I've adopted the cleaner code block above.

nawtrey:
> That sounds like a vote for adding some basic tests as well--whether they are done here or issues are opened for basic tests and the more detailed testing noted above 
> 
> The build system/packaging stuff could be deferred to an issue as well

I've opened issues for both the testing and system/packaging things. 

> A rebase may be good on the next round of revisions 

I've also rebased this branch after making the suggested changes.

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,235 @@
...
+    Parameters
+    ----------
+    log_path: path to a darshan log file.
+
+    """
+    def __init__(self, log_path: str):
+        # store the log path and use it to generate the report
+        self.log_path = log_path
+        # store the report
+        self.report = darshan.DarshanReport(log_path, read_all=True, dtype="pandas")
```
tylerjereddy, line 28: I think this call signature with dtype specification is the exact one that @jakobluettgau wanted us to avoid per discussion this afternoon and gh-490, so probably good to open an issue for this specific usage.

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,235 @@
...
+        Retrieves the locally stored CSS.
+        """
+        # get the path to the style sheet
+        with importlib_resources.path(darshan.cli, "style.css") as path:
+            # collect the css entries
+            with open(path, "r") as f:
+                self.stylesheet = "".join(f.readlines())
+
+
+def setup_parser(parser: Any):
```
tylerjereddy, line 172: I can't reproduce the need to use `Any` here:

```diff
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -197,6 +197,7 @@ def main(args: Union[Any, None] = None):
     """
     if args is None:
         parser = argparse.ArgumentParser(description="")
+        reveal_type(parser)
         setup_parser(parser)
         args = parser.parse_args()
```

`darshan/cli/summary.py:200: note: Revealed type is "argparse.ArgumentParser"`

darshan-util/pydarshan/darshan/cli/summary.py
tylerjereddy: ok, I opened the issue...

darshan-util/pydarshan/darshan/cli/summary.py
tylerjereddy: I'll open an issue for a typing cleanup on this PR/code..

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,235 @@
...
+        cmd = report.metadata["exe"]
+        if not cmd:
+            # if there is no executable
+            # label as not available
+            cmd = "N/A"
+        elif cmd.isdigit():
+            # if it can be converted to an
+            # integer label it anonymized
+            cmd = "Anonymized"
+        return cmd
```
tylerjereddy, line 62: FWIW, I think the code is so readable now that the comments are superfluous, but I'll leave that up to you.

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,235 @@
...
+    ----------
+    parser : command line argument parser.
+
+    """
+    parser.add_argument(
+        "log_path",
+        type=str,
+        help="Specify path to darshan log.",
+    )
+    parser.add_argument("--output", type=str, help="Specify output filename.")
```
tylerjereddy, line 186: I thought I asked for an issue to be opened about argparse problems. For example, the help not listing the options like `--output` here with `python -m darshan summary --help`:

```
usage: darshan <command>

PyDarshan CLI Utilities

optional arguments:
  -h, --help  show this help message and exit
  --debug
  --version
```

I'll open it now...

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -0,0 +1,235 @@
...
+            print(
+                f"Report generated successfully. \n"
+                f"Saving report at location: {save_path}"
+            )
+            # save the rendered html
+            f.write(stream)
+
+
+if __name__ == "__main__":
+    main()
```
tylerjereddy, line 235: I would normally try to place `argparse` stuff in this block, then have `main()` accept the arguments that were pulled out of the original command line. Wouldn't be surprised at all if some complications arose because of indirection here.

That may also substantially simplify the task of making `main()` directly/elegantly testable with regression suite without resorting to direct calls of the script via command line ugliness (gh-489).

tylerjereddy:
Ok, I'll put this in after a combing through it again, and based on feedback from Shane and Phil re: the basic appearance of the report.

A substantial number of issues have been opened so that this can be merged swiftly rather than blocked on several additional rounds of revisions. Many of those issues can likely be dealt with using small/surgical PRs that should be easy for me to review (i.e., some typing fixes in 1 PR, one or two tests in another, but not hundreds of lines at once for most of those things..).

Thanks

darshan-util/pydarshan/darshan/cli/summary.py
jakobluettgau: Having main() not require any of the argument parsing seems to me as well!

kevin-harms:
I would suggest not creating our own css and instead use an existing one. This avoids dealing with the various cross-browser compatibility issues as the css frameworks handle it.

Bootstrap is one, but there are many... short list I found: https://geekflare.com/best-css-frameworks/

nawtrey:
> I would suggest not creating our own css and instead use an existing one. This avoids dealing with the various cross-browser compatibility issues as the css frameworks handle it.

Yeah at some point I think our css file will be replaced with something more substantial. For now I think the focus is on building the figures we want and getting them into the report. We can open an issue about the css file if folks want to make suggestions there.

Issue #494, opened 2021-09-23
MAINT: replace use of Any with actual types in gh-485


tylerjereddy:
gh-485 will likely be merged with type hints that are too broad--it is a simple matter to use `reveal_type()` with `mypy` to verify if `Any` is really justified in various location so that should be done..

See this comment for example (and I saw at least 1 other location that should be double-checked, could be others): https://github.com/darshan-hpc/darshan/pull/485/files#r714428486

PR #500, opened 2021-09-27
MAINT: Replace overly broad type hint

nawtrey:
* Replace `Any` type hint in `setup_parser()`
with `argparse.ArgumentParser`

* Fix gh-494


nawtrey:
This was the only type hint I could find that wasn't specific. The only other instance of `Any` is in `main()`, where I used `reveal_type()` to get the current type hint that is there.

Issue #476, opened 2021-09-08
BLD: modernized build for pydarshan


tylerjereddy:
We should be able to do `python -m pip install .` instead of `python setup.py install` in `pydarshan`. At the moment this fails with autoperf discovery issues. The direct `setup.py` method is largely discouraged in the community these days, so supporting the PEP-compliant `pip` approach would be useful eventually, and is also relevant for wheel builds.

Here's part of the traceback:

```
  ERROR: Command errored out with exit status 1:
   command: /Users/treddy/python_venvs/python_381_venv/bin/python -u -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/j8/6zgt1c2x1jj3b5z5h29n38xw000w2g/T/pip-req-build-k9b19664/setup.py'"'"'; __file__='"'"'/private/var/folders/j8/6zgt1c2x1jj3b5z5h29n38xw000w2g/T/pip-req-build-k9b19664/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /private/var/folders/j8/6zgt1c2x1jj3b5z5h29n38xw000w2g/T/pip-wheel-3sneaf0s
       cwd: /private/var/folders/j8/6zgt1c2x1jj3b5z5h29n38xw000w2g/T/pip-req-build-k9b19664/
  Complete output (4 lines):
  running bdist_wheel
  running build
  running build_py
  error: can't copy 'darshan/backend/apxc.py': doesn't exist or not a regular file
  ----------------------------------------
  ERROR: Failed building wheel for darshan

```

Actually, I suspect @jakobluettgau has a way to do this for building wheels already, but we should probably support this independent of any "special" process at some point.

tylerjereddy:
Note that this works "ok" in the absence of the autoperf options/modules being enabled at configure time for `darshan-util`

jakobluettgau:
I think that sounds like a good idea (just as a note, on pydarshan-devel and main with a fresh clone at least `python -m pip install .` worked just fine for me, but that doesn't include any autoperf yet.).

[...]

PR #504, opened 2021-09-30
CI: rm --enable-pydarshan

tylerjereddy:
* the `--enable-pydarshan` config flag is removed from
CI because this adds an extra/redundant `python setup.py install`
which could confuse attempts to install a clean wheel
build in the environment, etc.

* some related discussion in gh-476, although this doesn't
solve that apart from avoiding automatic use of `python setup.py install`


tylerjereddy:
Ok, CI is green and straigtforward CI cleanup so merging.

Issue #419, opened 2021-07-28
TST, CI, MAINT: min/max versions supported


tylerjereddy:
Establishing the min/max versions of various Python deps that are supported and testing them in CI to ensure compatibility may be useful. Otherwise, it can be easy to start using bleeding edge features. If we're ok with bleeding edge only, it may be less of a concern.

tylerjereddy:
`pyproject.toml` is often used for this kind of thing

tylerjereddy:
With Python 3.10 coming out Monday, it may be worth bringing this up again--to have a sane maintenance burden we may only want to support the 3 recent most recent stable versions of Python per NEP29: https://numpy.org/neps/nep-0029-deprecation_policy.html

tylerjereddy:
@kevin-harms thoughts? If your machines are anything like ours, Python gets updated slowly. That said, users can run the Python code locally easily enough in theory.

kevin-harms:
I think python 3.6[.9] is a safe minimum that is available on most platforms these days. We could probably argue about the maintenance release level.. If that is too old, we can discuss something newer.

tylerjereddy:
I made an attempt to add `3.6` to CI in the cross-linked PR, but it does look like it will already require a few shims in the code. It is also worth nothing that `3.6` will be EOL in a few months: https://devguide.python.org/#status-of-python-branches

PR #505, opened 2021-09-30
CI: add Python 3.6

tylerjereddy:
* there's a request from Kevin to support Python 3.6 for
now so add it to the CI (see: https://github.com/darshan-hpc/darshan/issues/419#issuecomment-930540108 )

* note that Python 3.6 will be EOL in just a few months:
https://devguide.python.org/#status-of-python-branches

* I suggest revisiting this in the near future, also
to keep our support surface manageable

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         platform: [ubuntu-latest,
                    macos-latest]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


tylerjereddy:
@kevin-harms @carns note the CI failures here--supporting Python 3.6 through 3.10 is going to be a bit of work, so definitely make sure we need it. 3.6 will be EOL in just a few months.

We can probably make it work if need though.

kevin-harms:
I took a look at some of the relevant systems we work on and 3.6 is the default on all systems so it would make a suitable baseline. If 3.6 becomes problematic, we could bump up to 3.8 as a minimum, but that means users on some systems will need to select a non-default python3.

tylerjereddy:
Ok, just wanted confirmation, thanks

tylerjereddy:
Ok, CI is passing with Python `3.6` added in. I will leave this open for comments/suggestions for a few days, then merge in to enforce the support requirements at Argonne. I'll also open an issue to remind us about what we can "restore" once we can ditch Python `3.6` support.

tylerjereddy:
Ok, no further comments.. the crippling of our type hints to support 3.6 is most unfortunate, but I'll merge this in to support HPC requirements.

PR #506, opened 2021-10-01
CI: build wheels in CI

### LINKED ISSUE #476: BLD: modernized build for pydarshan

tylerjereddy:
Fixes #476

* build wheels in CI instead of using the discouraged
`python setup.py install`

* not only is this the preferred way to do things these days
(because i.e., `setup.py` can have arbitrary code and the
wheel build is PEP compliant, etc.), but it also represents
how most of our end users are likely to install the package
if they use `pip`

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade pytest mypy pyflakes asv pytest-cov codecov
+          python -m pip install --upgrade pytest mypy pyflakes asv pytest-cov codecov wheel
           # matplotlib is pinned because of
           # gh-479
           python -m pip install matplotlib==3.4.3
@@ -46,8 +46,8 @@ jobs:
       - name: Install pydarshan
         run: |
           cd darshan-util/pydarshan
-          # TODO: use pip per gh-476
-          python setup.py install
+          make add-modules
+          python -m pip install .
       # shim for importlib.resources on Python < 3.9
       - if: ${{matrix.python-version < 3.9}}
         name: Install importlib_resources


tylerjereddy:
Looks like wheels are properly used now; MacOS and Linux CI entries both show output similar to this:

```
Building wheels for collected packages: darshan
  Building wheel for darshan (setup.py): started
  Building wheel for darshan (setup.py): finished with status 'done'
  Created wheel for darshan: filename=darshan-3.3.1.0-py2.py3-none-any.whl size=645662 sha256=ac4985456d52ece72c798291e62e18f7564f1c14353cfa35d95f381b4619eb73
```

Issue #507, opened 2021-10-01
BLD: use pyproject.toml


tylerjereddy:
We should consider using `pyproject.toml` for our wheel builds/install system to comply with [PEP518](https://www.python.org/dev/peps/pep-0518/), and to get in line with modern Python packages here.

tylerjereddy:
I'll see if I can draft this today

jakobluettgau:
Sounds good to me. Until recently we didn't really use many new API additions of numpy, pandas, or matplotlib, but i agree having minimum version prevents headaches in various situations beyond building.

Is there any experience for versions we might want to aim for good site/python compatibility?

shanedsnyder:
I think this is also mostly addressed by #969. `setup.py` is now pretty barebones, with only the `ext_modules` logic and the code for copying in AutoPerf python code (this will probably go away soon if we merge AutoPerf code into the Darshan repo).

PR #509, opened 2021-10-01
BLD: add simple pyproject.toml

tylerjereddy:
A small part of #507 standards adoption

* add a very simple `pyproject.toml` to move us closer
to adoption of PEP517 and PEP518

* if working properly, we should see output like this
for the `pydarshan` install in CI:
- `Building wheel for darshan (PEP 517)`
- instead of the old, less standards-compliant version:
`Building wheel for darshan (setup.py)`

* for an informal discussion of the advantages see:
https://snarky.ca/what-the-heck-is-pyproject-toml/

* I suspect some gradual cleanup of what we actually
need in `setup.py` would be needed to take real
advantage of the benefits here, but I've avoided messing
around there initially

### DIFF
--- a/darshan-util/pydarshan/pyproject.toml
+++ b/darshan-util/pydarshan/pyproject.toml
@@ -1,0 +1,5 @@
+[build-system]
+requires = [
+    "wheel",
+    "setuptools",
+]


tylerjereddy:
Alright, `PEP517` is indeed indicated in the wheel build in CI now. I'll go ahead and merge this, but defer further build system adjustments until there is more discussion there, if needed.

PR #511, opened 2021-10-04
CI: Update pydarshan install to install in same directory

nawtrey:
* Add flag `--use-feature=in-tree-build` to pydarshan install
command so it installs in current directory instead of using
a temporary one. This creates the autoperf symlinks correctly.

* Remove `make add-modules` since it is unnecessary


nawtrey:
I'll open an issue to remove this flag once `pip 21.3` is released since it will be the default at that point.

tylerjereddy (approved):
alright

Issue #489, opened 2021-09-22
TST: ensure summary report assets are distributed properly when installing from a wheel


nawtrey:
As Tyler discussed [here](https://github.com/darshan-hpc/darshan/pull/485#issuecomment-920504931), there are certain assets (i.e. `.html` and `.css` files) that may not be available when installing Darshan from a wheel. Similar issues have been brought up before, like [here](https://github.com/darshan-hpc/darshan/pull/386#discussion_r650141987) in Kevin's PR, where an adjustment to `MANIFEST.in` was necessary to keep the template code.

Tyler suggested the following:
- have a new CI matrix entry that tests "command line job summary generation from a wheel install"
- it may be most suitable to ensure this can be done inside a conventional regression test without having to use a subprocess--for example, that it is possible to call main() with appropriate arguments on each of the log files in the logs repo and make sure an .html artifact gets generated as expected in a temporary test directory (instead of the command line call proper)

tylerjereddy:
> I don't think we need a CI entry to address this issue since the CI is currently installing pydarshan using a wheel.

Yes, I fixed it myself in gh-506...

tylerjereddy:
Actually, I'm not convinced the testing is sufficient yet for the wheel *install*, I'll comment in one of the other PRs initially

Issue #488, opened 2021-09-22
TST: Add tests to cover summary report code from GH-485 


nawtrey:
Tests need to be added to cover the darshan summary report generation code from GH-485 (currently at 0% coverage). Many of the functions generate strings that could be at least regression tested.

Issue #502, opened 2021-09-29
BUG: summary report code fails to run for Python 3.9


nawtrey:
When executing the following command:

`python -m darshan summary /path/to/log.darshan`

in a Python 3.9 environment, I get the following error:
```
Traceback (most recent call last):
  File "/Users/nawtrey/opt/anaconda3/envs/pydarshan-dev/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/nawtrey/opt/anaconda3/envs/pydarshan-dev/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/nawtrey/darshan_install/darshan/darshan-util/pydarshan/darshan/__main__.py", line 3, in <module>
    main()
  File "/Users/nawtrey/darshan_install/darshan/darshan-util/pydarshan/darshan/cli/__init__.py", line 164, in main
    mod.main(args)
  File "/Users/nawtrey/darshan_install/darshan/darshan-util/pydarshan/darshan/cli/summary.py", line 218, in main
    with importlib_resources.path(darshan.cli, "") as path:
  File "/Users/nawtrey/opt/anaconda3/envs/pydarshan-dev/lib/python3.9/contextlib.py", line 117, in __enter__
    return next(self.gen)
  File "/Users/nawtrey/opt/anaconda3/envs/pydarshan-dev/lib/python3.9/importlib/resources.py", line 175, in _path_from_reader
    opener_reader = reader.open_resource(norm_resource)
  File "<frozen importlib._bootstrap_external>", line 995, in open_resource
IsADirectoryError: [Errno 21] Is a directory: '/Users/nawtrey/darshan_install/darshan/darshan-util/pydarshan/darshan/cli'
```

I've confirmed this on my own fork while testing the CLI for other reasons. It passes for our Python 3.7 and 3.8 runners, but only passes for 3.9 if we adopt the following change in `darshan/cli/summary.py`:

```python
with importlib_resources.path(darshan.cli, "base.html") as base_path:
    # load a template object using the base template
    template = Template(filename=str(base_path))
```

tylerjereddy:
Sounds like there should be tests for running the report in CI per #488 and #489. Also, Python 3.10 comes out next week, so testing will likely be expanded then. Anyway, definitely need tests running...

nawtrey:
Agreed, I discovered this issue while working on those issues.

[...]

PR #513, opened 2021-10-04
TST: Add tests for darshan.cli.summary.py

nawtrey:
* Add tests that probe `summary.main()`

* Add test that iterates over all logs in the darshan-logs
repo to ensure they run to completion

* Add `mako` dependency to `setup.py`

* Contributes to #488 and #489


nawtrey:
I  used `unittest.mock.patch` to input the command line arguments since it is pretty simple and doesn't add any new dependencies. 

This PR is going to fail the Python 3.9 runners initially, but I'll open a PR with a patch that fixes it shortly. I just wanted to demonstrate the issue before I fix it.

nawtrey:
Okay, Python 3.7 runner failed because of the missing `mako` dependency. I'll fix that shortly.

nawtrey:
Okay, now the Python 3.9 runner failed with the anticipated error from #502:
```
resource = ''

>   ???
E   IsADirectoryError: [Errno 21] Is a directory: '/home/runner/work/darshan/darshan/darshan-util/pydarshan/darshan/cli'

<frozen importlib._bootstrap_external>:1055: IsADirectoryError
```
This is fixed in gh-514.

tylerjereddy:
There we go (after some Python `3.6` shims): `E   FileNotFoundError: [Errno 2] No such file or directory: '/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/darshan/cli/style.css'`

That's one of the files missing from the wheel, which is the original error I would expect, so I'll push in the packaging fix above next..

tylerjereddy:
Ok, now the wheel seems to install the appropriate assets, and the original error is back, except that it is pointing to the appropriate path (the `site-packages` install dir, not the source dir) for the incorrect asset path:

`IsADirectoryError: [Errno 21] Is a directory: '/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/site-packages/darshan/cli'`

I'll just cherry-pick that over from the other PR now..

tylerjereddy:
Alright, I fixed up the `pytest`/codecov reports and reviewed the tests--the patch diff is 100 % now as expected. 

The tests could also check that there is some "sane" content in the html files, rather than just checking for their existence, but I'll merge this as an incremental improvement for now.

I was not super convinced that `main()` needed to accept both arguments and no arguments, nor that much thought was put into testing of install directories vs. source directories before my shims/test refinements were added in, but this should fix that up.

PR #516, opened 2021-10-05
(demo only) CI: show wheel portability failure (lack of data file install)

tylerjereddy:
(do not merge--for demo purposes only)

@nawtrey this is a demo-only PR intended to demonstrate a few things:

- the wheels are not portable, and this is not restricted to Python 3.9 (or any Python version, I've used 3.8 here to clearly show it is not just 3.9)
- an example of how you can test this in a clean environment, though it would be nice to have something more concise right in `pytest` (I don't have the bandwidth to come up with that at the moment..)

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         platform: [ubuntu-latest,
                    macos-latest]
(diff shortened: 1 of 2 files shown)


tylerjereddy:
Captured with a one-liner in gh-519, so closing.

Issue #408, opened 2021-06-29
MAINT: gitignore and pydarshan work


tylerjereddy:
The `pydarshan` subdirectory has its own `.gitignore`, which is useful, but when building out the C libs of `darshan-util` and so on in the same checkout, the tracking is a bit of a mess with object files we don't care about, and could easily be cleaned up without affecting the `.gitignore` settings of the other parent directories (actually, they don't have `.gitignore` anyway, but I'm mostly focused on `pydarshan` here):

`git status`

![image](https://user-images.githubusercontent.com/7903078/123828919-36857080-d8bf-11eb-9f47-6b72a178759d.png)

tylerjereddy:
This situation is still a mess on `pydarshan-devel` branch, even after we rebase. In some ways, it is even worse with the new autotools stuff causing changes to tracked files like `configure` as well

jakobluettgau:
Was there a particular reason not to include a .gitignore also at the repository root?

I think most of the object files and other temporary or compiled files can be covered with a couple of simple rules.

The configure file probably shouldn't be in pydarshan-devel (it is not in main).
For CI this might merely mean to call ./prepare.sh once and have autotools in the environment installed.

As a sidenote: I would still maintain a seperate .gitignore for the pydarshan directory. On the one hand because it keeps patterns simple and on the other hand because it allows for easier refactoring such as moving the pydarshan directory around.

tylerjereddy:
Yeah, a carefully crafted `.gitignore` should fix most of what I see I think.

Maybe `configure` wasn't tracked, you are probably right, but it was polluting my `git status` at least.

shanedsnyder:
Relatedly, it looks like `darshan-util/configure` is somehow tracked again on `pydarshan-devel`, maybe something weird happened as part of the rebase or some other merge. We will need to clear that stuff up at some point before merging back to `main`.

PR #524, opened 2021-10-14
MAINT: backport gitignore

tylerjereddy:
Backport gh-522 from Phil. Related to #408.

@carns this is what I have left in my `git status` after the backport of new `.gitignore`:

image


carns:
@tylerjereddy ok, I can help with that.  This is a good chance for me to learn a little more about how the git actions work too.

The change we need is simple: where it runs `./configure ...` it can just do something like `mkdir build && cd build && ../configure ...`. 

`../configure` from some subdir will put all of the resulting makefiles and compile artifacts in that subdir, pretty much as if you had done `cmake ../` in a CMake project.  Also like cmake you can do it in any directory, there is nothing special about make a "build" subdir, that's just what I tend to do.

What's the best github-ish way for me to suggest/test that change?  Do I open up a PR modifying the .yml, and if so will it execute the modified version in my PR so we can see if it worked?

PR #526, opened 2021-10-15
ENH: add infrastructure for adding figures, job i/o activity section

### LINKED ISSUE #463: job summary subsection: job i/o activity (heatmap)
Current goal is to make the first figure of the report be the heatmap figures that @nawtrey has been working on in #396.

Once we have a template to work with, we should figure out how we want to present this section. This section will always follow the high-level job metadata section described in #462 and will likely just be centered on the page, so nothing complicated there. Key things to figure out are:

1. What to display when DXT data is not available, since this plot currently relies on DXT data? I.e., warn about figure being tied to DXT for now, pointers to how to enable DXT, etc.
2. Come up with a helpful description of the plot so that users can quickly understand what they are looking at.

In this case, this is likely something the template engine simplifies, allowing either a figure to be embedded in the HTML in case of logs with DXT data, otherwise generating some code to display warnings about DXT not being enabled.

nawtrey:
* Add `ReportFigure` to `summary.py` for collecting
figure information to simplify figure registration process

* Add infrastructure to darshan summary report
generation code that allows for adding figures
via a new method `ReportData.register_figures`

* Add entries to `style.css` to correct
formatting for figures and their captions

* Add nested loop to `base.html` to create the
figure sections and insert the encoded images

* Fix #463

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/base.html
+++ b/darshan-util/pydarshan/darshan/cli/base.html
@@ -28,6 +28,27 @@
         ${report_data.module_table}
       </section>
 
+      <!-- Add sections and their figures  -->
+      % for sect_title in report_data.sections:
+          <section>
+          <h2>${sect_title}</h2>
+          % for fig_title, fig in report_data.sections[sect_title].items():
+            <h3>${fig_title}</h3>
+            <figure style="width: ${fig.fig_width}px;">
+            % if fig.img_str:
+              ${fig.img_str}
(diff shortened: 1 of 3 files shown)


tylerjereddy:
Ok, I did another round of reading through the code/comments and running locally. It seems like we should merge this and open an issue for allowing i.e., tables to added through the API as well.

tylerjereddy:
thanks Nik et al.

Issue #465, opened 2021-08-31
job summary subsection: per-module i/o characteristics


shanedsnyder:
Once we have a functional template we can build on, we should add a subsection related to I/O characteristics for different modules contained in the log. There are a number of characteristics that we have traditionally captured for a few Darshan modules (specifically, POSIX, MPI-IO, and STDIO), but it would likely be straightforward to extend the idea to other modules that aren't currently covered (HDF5 being an obvious one). 

I think as a starting point, we could try to reproduce the following types of graphs (note numbers at top) from our current job summary tool, probably starting with just POSIX and MPI-IO for now:
![image](https://user-images.githubusercontent.com/24571836/131564549-b49ff6eb-2e1a-418f-85f5-e9e605d5b5ab.png)
 
1. I/O cost graphs: Provide estimates on per-process I/O time spent doing read/write/metadata operations for the corresponding module API (POSIX, MPI-IO, etc.). This requires us re-implementing some performance estimate logic that is currently done in our darshan-parser C utility. Display using stacked bar graph. Important to display cost both in terms of percentage runtime and elapsed runtime in seconds.
2. I/O operation count graphs: Provide total number of different I/O operations executed using the module API. Things like opens, reads, writes, stats, seeks, etc. To generalize this plot, likely need an API here to allow modules to specify what operations they support and what counters are needed to access them (e.g., mapping "opens" to `POSIX_OPENS` counter for the POSIX module). Display using a bar graph, with each operation on the x-axis. Likely need log-scale support on y-axis for large ranges.
3. Access size histograms: Provide distributions of all read and write operation access sizes using histograms. A few modules capture this information in the same way (e.g., `POSIX_SIZE_WRITE_0_100`-`POSIX_SIZE_WRITE_1G_PLUS` counters for the POSIX module) so should be generalizable. 
4. Common access sizes: Darshan does capture information on the 4 most common access sizes issued to each file, so we could aggregate that information and summarize the most common ones for useful details that are too fine-grained to see in histogram described for plot 3.

The template engine would need to detect the presence of these modules, then generate the figures described above along with corresponding HTML code in different module sub-sections of the report as shown.

After completing this, we could experiment with extending these sorts of summaries to STDIO and HDF5 modules, and could also look at adding module-specific summaries in addition to the "common" ones described here. An example module-specific graph would be the figure in our current job summary tool that displays percentage of POSIX operations that are sequential or consecutive in files.

nawtrey:
[...]

Issue #492, opened 2021-09-22
MAINT: experimental vs. non-experimental namespaces


tylerjereddy:
Probably low priority, but we could think about moving some things from `experimental` to more prominent namespaces. This came up in a recent meeting with arguments for including some plots that are basically "mandatory" in the report anyway, and arguments against including providing at least a modest initial testing period for the new code.

New code is isolated o`pydarshan-devel` for now anyway. This is also somewhat of a duplication of the public (intended/ready for consumption) vs. private issue that was opened some time ago: gh-423

kevin-harms:
For some background, the 'experimental' name space has many analysis functions which were developed but not throughly tested and some of them only support a specific data type, like only numpy or only pandas. We considered removing all the experimental code but elected to leave it and document is as unsupported. If there is anything that gets adapted for use in the main code, it should be moved from experimental and make sure appropriate documentation is added.

PR #549, opened 2021-11-01
ENH: Add I/O cost figure module and tests

nawtrey:
* Add module for I/O cost figure

* Add testing module to cover I/O cost functions

* Add import to `__init__.py` to ease in importing all
plotting functions in summary report code

* Contributes to #465

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
@@ -1,0 +1,119 @@
+"""
+Module for creating the the I/O cost
+bar graph for the Darshan job summary.
+"""
+from typing import Any
+
+import numpy as np
+import pandas as pd
+import matplotlib.pyplot as plt
+import matplotlib.ticker as mtick
+
+import darshan
+
+
+def get_io_cost_df(report: darshan.DarshanReport, mod_key: str) -> Any:
+    """
+    Generates the I/O cost dataframe which contains the
+    raw data to plot the I/O cost stacked bar graph.
+
+    Parameters
+    ----------
+    report: a ``darshan.DarshanReport``.
+
+    mod_key: module to generate the I/O cost stacked
+    bar graph for (i.e. "POSIX", "MPI-IO", "STDIO").
+
+    Returns
+    -------
+    io_cost_df: a ``pd.DataFrame`` containing the
+    average read, write, and meta times.
+
+    """
+    # collect the records in dataframe form
+    recs = report.records[mod_key].to_df(attach=["rank"])
+    # correct the MPI module key
+    if mod_key == "MPI-IO":
+        mod_key = "MPIIO"
+    # filter out all except the following columns
+    cols = [
+        "rank",
+        f"{mod_key}_F_READ_TIME",
+        f"{mod_key}_F_WRITE_TIME",
+        f"{mod_key}_F_META_TIME",
+    ]
+    rd_wr_meta_df = recs["fcounters"].filter(cols, axis=1)
+    # locate any rows where "rank" == -1 and divide them by nprocs
+    nprocs = report.metadata["job"]["nprocs"]
+    rd_wr_meta_df.loc[rd_wr_meta_df["rank"] == -1] /= nprocs
+    # drop the "rank" column since it's no longer needed
+    rd_wr_meta_df.drop("rank", axis=1, inplace=True)
+    # rename the columns so the labels are automatically generated when plotting
+    name_dict = {cols[1]: "Read", cols[2]: "Write", cols[3]: "Meta"}
+    rd_wr_meta_df.rename(columns=name_dict, inplace=True)
+    # calculate the means and put the output `pd.Series` into a new dataframe
+    io_cost_df = pd.DataFrame({"Avg. Operation Time": rd_wr_meta_df.mean(axis=0)}).T
+    return io_cost_df
+
+
+def plot_io_cost(report: darshan.DarshanReport, mod_key: str) -> Any:
+    """
+    Creates a stacked bar graph illustrating the percentage of
+    runtime spent in read, write, and metadata operations.
+
+    Parameters
+    ----------
+    report: a ``darshan.DarshanReport``.
+
+    mod_key: module to generate the I/O cost stacked
+    bar graph for (i.e. "POSIX", "MPI-IO", "STDIO").
+
+    Returns
+    -------
+    io_cost_fig: a ``matplotlib.pyplot.figure`` object containing a
+    stacked bar graph of the average read, write, and metadata times.
+
+    Raises
+    ------
+    NotImplementedError: raised if the input module key is not "POSIX",
+    "MPI-IO", or "STDIO".
+
+    """
+    if mod_key not in ["POSIX", "MPI-IO", "STDIO"]:
+        # TODO: expand the scope of this function
+        # to include HDF5 module
+        raise NotImplementedError(f"{mod_key} module is not supported.")
+    # calculate the run time from the report metadata
+    runtime = report.metadata["job"]["end_time"] - report.metadata["job"]["start_time"]
+    if runtime == 0:
+        # for cases where runtime is < 1, just set it
+        # to 1 like the original perl code
+        runtime = 1
+    # get the I/O cost dataframe
+    io_cost_df = get_io_cost_df(report=report, mod_key=mod_key)
+    # generate a figure with 2 y axes
+    io_cost_fig = plt.figure(figsize=(4.5, 4))
+    ax_raw = io_cost_fig.add_subplot(111)
+    ax_norm = ax_raw.twinx()
+    # use the dataframe to plot the stacked bar graph
+    io_cost_df.plot.bar(stacked=True, rot=0, ax=ax_raw, legend=False, zorder=3)
+    ax_raw.grid(axis="y", zorder=0, alpha=0.6)
+    # set the y limits for both axes
+    ax_raw.set_ylim(0, runtime)
+    ax_norm.set_ylim(0, 100)
+    # convert the normalized axis y labels to percentages
+    ax_norm.yaxis.set_major_formatter(mtick.PercentFormatter())
+    # align both axes tick labels so the grid matches
+    # values on both sides of the bar graph
+    norm_yticks = ax_raw.get_yticks()
+    n_ticks = len(ax_norm.get_yticks())
+    yticks = np.linspace(norm_yticks[0], norm_yticks[-1], n_ticks)
+    ax_raw.set_yticks(yticks)
+    # add the legend and appropriate labels
+    ax_raw.set_ylabel("Runtime (s)")
+    handles, labels = ax_raw.get_legend_handles_labels()
+    ax_norm.legend(handles[::-1], labels[::-1], loc="upper left", bbox_to_anchor=(1.3, 1.02), title="Operation")
+    # adjust the figure to reduce white space
+    io_cost_fig.subplots_adjust(right=0.59)
+    io_cost_fig.tight_layout()
+    return io_cost_fig
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
@@ -1,5 +1,5 @@
 """
-Module for creating the the I/O cost
+Module for creating the I/O cost
 bar graph for the Darshan job summary.
 """
 from typing import Any
@@ -10,6 +10,46 @@ import matplotlib.pyplot as plt
 import matplotlib.ticker as mtick
 
 import darshan
+
+
+def get_by_avg_series(df: Any, mod_key: str, nprocs: int) -> Any:
+    """
+    Create the "by-average" series for the stacked
+    bar graph in the I/O cost figure.
+
+    Parameters
+    ----------
+    df: the dataframe containing the relevant data, typically the
+    "fcounter" data from a Darshan report.
+
+    mod_key: module to generate the I/O cost stacked
+    bar graph for (i.e. "POSIX", "MPI-IO", "STDIO").
+
+    nprocs: the number of MPI ranks used for the log of interest.
+
+    Returns
+    -------
+    by_avg_series: a ``pd.Series`` containing the
+    average read, write, and meta times.
+
+    """
+    # filter out all except the following columns
+    cols = [
+        "rank",
+        f"{mod_key}_F_READ_TIME",
+        f"{mod_key}_F_WRITE_TIME",
+        f"{mod_key}_F_META_TIME",
+    ]
+    df = df.filter(cols, axis=1)
+    # locate any rows where "rank" == -1 and divide them by nprocs
+    df.loc[df["rank"] == -1] /= nprocs
+    # drop the "rank" column since it's no longer needed
+    df.drop("rank", axis=1, inplace=True)
+    # rename the columns so the labels are automatically generated when plotting
+    name_dict = {cols[1]: "Read", cols[2]: "Write", cols[3]: "Meta"}
+    df.rename(columns=name_dict, inplace=True)
+    by_avg_series = df.mean(axis=0)
+    return by_avg_series
 
 
 def get_io_cost_df(report: darshan.DarshanReport, mod_key: str) -> Any:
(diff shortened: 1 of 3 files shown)


nawtrey:
My first crack at the I/O cost plot from #465 (plot 1). 

Examples
----------
I have attached a zipped file with some example outputs: [example_plots.tar.gz](https://github.com/darshan-hpc/darshan/files/7456727/example_plots.tar.gz). There are 6 plots in total, 3 for each log since they each have all 3 supported modules. 

Script
------
Reviewers should be able to generate figures using the following script.

<details>
<summary>Generate IO cost plot code</summary>

```python
import os
import argparse
import glob
import darshan
from darshan.experimental.plots import plot_io_cost

if __name__ == '__main__':

    # get the path to the file(s) to be read in
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "log_path",
        nargs="?",
    )
    args = parser.parse_args()
    log_path = args.log_path

    if os.path.isfile(log_path):
        # individual file case
        log_paths = [log_path]
    elif os.path.isdir(log_path):
        # directory case
        log_paths = glob.glob(os.path.join(log_path, "*.darshan"))

    for log_path in log_paths:
        report = darshan.DarshanReport(log_path, read_all=True)
        for mod in report.modules:
            try:
                fig = plot_io_cost(report=report, mod_key=mod)
            except NotImplementedError:
                continue
            else:
                if fig:
                    fname = os.path.basename(log_path)
                    fig.savefig(f"{fname}_{mod}.png", dpi=300)
                plt.close()
```

</details>

shanedsnyder:
Looks good to me!

I think probably the x-axis label could be dropped or at the very least clarified? There's only one bar in each case so not sure it's needed if the point can be communicated with a title and/or caption. I'm thinking the title is probably something like "I/O cost by average", with us having similar plots for other metrics (e.g., "I/O cost by slowest process") eventually like we've discussed in meetings.

Along those lines, it's probably worth making sure the code is generalized so you can re-use the plotting code with data generated using different cost metrics (average, slowest, etc.). Right now it looks like you'd have to re-write a new plotting routine and cost metric routine for each one, but you probably just need one plotting routine that can work with whatever cost metrics we end up supporting, if that makes sense.

shanedsnyder:
Also, at what point should we start moving things like this out of `experimental` namespace? I don't have anything strict in mind, but for sure once there are tests and the plots are used somewhere (i.e., the job summary tool), I think it makes sense. Are other figures in the job summary currently still in `experimental` or should we move this?

nawtrey:
> I think probably the x-axis label could be dropped or at the very least clarified? There's only one bar in each case so not sure it's needed if the point can be communicated with a title and/or caption. I'm thinking the title is probably something like "I/O cost by average", with us having similar plots for other metrics (e.g., "I/O cost by slowest process") eventually like we've discussed in meetings.

I've been avoiding putting titles because they will likely be placed by the summary report code. But I do think it is important to mention this is an average, so however you think that should get done is fine with me. I could remove the x-axis label and change the y-axis label to read "Avg. Runtime (s)" or something like that instead. 
 
> Along those lines, it's probably worth making sure the code is generalized so you can re-use the plotting code with data generated using different cost metrics (average, slowest, etc.). Right now it looks like you'd have to re-write a new plotting routine and cost metric routine for each one, but you probably just need one plotting routine that can work with whatever cost metrics we end up supporting, if that makes sense.

Okay, I guess my understanding of it was we would be changing this from "I/O cost by average" to "I/O cost by slowest", which would just be a matter of changing the keys used, I think. But if we want the option to have both I can work on generalizing the code a bit so the barrier is lower, if nothing else.

nawtrey:
> Also, at what point should we start moving things like this out of `experimental` namespace? I don't have anything strict in mind, but for sure once there are tests and the plots are used somewhere (i.e., the job summary tool), I think it makes sense. Are other figures in the job summary currently still in `experimental` or should we move this?

That discussion was started in #492, but at the moment all of the plotting functions are in `experimental`. I agree that some of them could be moved out at this point, but we haven't been concerned with moving them because it is a pretty low-cost change. My guess is we will probably wait until all of the plotting functions (and their respective tests) are merged in before going down that route.

shanedsnyder:
> > I think probably the x-axis label could be dropped or at the very least clarified? There's only one bar in each case so not sure it's needed if the point can be communicated with a title and/or caption. I'm thinking the title is probably something like "I/O cost by average", with us having similar plots for other metrics (e.g., "I/O cost by slowest process") eventually like we've discussed in meetings.
> 
> I've been avoiding putting titles because they will likely be placed by the summary report code. But I do think it is important to mention this is an average, so however you think that should get done is fine with me. I could remove the x-axis label and change the y-axis label to read "Avg. Runtime (s)" or something like that instead.
> 

Yeah, maybe just the y-axis label change then, with us adding additional details with a title and/or caption in the job summary tool. 

> > Along those lines, it's probably worth making sure the code is generalized so you can re-use the plotting code with data generated using different cost metrics (average, slowest, etc.). Right now it looks like you'd have to re-write a new plotting routine and cost metric routine for each one, but you probably just need one plotting routine that can work with whatever cost metrics we end up supporting, if that makes sense.
> 
> Okay, I guess my understanding of it was we would be changing this from "I/O cost by average" to "I/O cost by slowest", which would just be a matter of changing the keys used, I think. But if we want the option to have both I can work on generalizing the code a bit so the barrier is lower, if nothing else.

Ultimately I think both plots could be useful, so I was imaging we would show both. "Average time by slowest" is a more effective way of measuring the I/O time for applications doing collective I/O, where some processes are doing I/O operations while others are idle -- in that case the average skews things because so many processes aren't actually doing I/O. But for independent I/O workloads where there isn't as much synchronization across processes and all processes are contributing similar amounts of I/O, I still think average is probably a more appropriate metric. Maybe @carns can sanity check whether ultimately including both makes sense?

carns:
These graphs look great visually!

I do think in the long run both the avg (what we have now) and the "by slowest" (what we'll be able to generat4e in the future) will be nice to show, probably side-by-side.  It will be helpful at a glance to see when they roughly match vs. when they are very different.

The ones for which those graphs look a lot different will be of particular interest.  It is kind of like an at-a-glance indicator of how unbalanced the IO was.  We can also see that on the right side of the heatmap in more detail (so it's not urgent to have this side by side io cost view), but having it head to head in a simple bar graph, with runtime percentages, will be helpful in the long run I think.

nawtrey:
Okay, thanks for the clarification. I'll see what I can do about generalizing the code so it can be reused for the "by slowest" plot.

shanedsnyder:
Yeah, thinking about it more, maybe you could just generalize by allowing new bars to be added to a single plot based on different metrics (average, slowest), with the x-axis label mentioning the metric used ('by-average', 'by-slowest', or something like that). That way you can view them all on one plot as Phil mentions.

nawtrey:
> Yeah, thinking about it more, maybe you could just generalize by allowing new bars to be added to a single plot based on different metrics (average, slowest), with the x-axis label mentioning the metric used ('by-average', 'by-slowest', or something like that). That way you can view them all on one plot as Phil mentions.

Yeah I can do that. How do I go about calculating the "by-slowest" values? I see there is `*_F_SLOWEST_RANK_TIME`, but that isn't broken down by read/write/meta.

shanedsnyder:
Hmm, I've been thinking about it more, and I guess we don't technically have the right counters in place in Darshan currently to break down the slowest I/O time into metadata, read, or write components, at least for shared files (`rank == -1`).

This 'by_slowest' metric we've been discussing has been used in our `darshan-parser` tool for awhile to provide I/O time and bandwidth measurements, but those measurements aren't broken down into each of the separate components. As you noted, we only have SLOWEST counters for total I/O time -- I think we'd need to add some additional counters to help break that down into read/write/meta components.

Sorry about the distraction, but I'd say hold off on over generalizing the code here until we have something implemented in darshan-runtime that can help calculate these new metrics. In the meantime, we could think more about how to display the different I/O bandwidth measurements for each module (bandwidth by average, by slowest, etc.) in the reports, but I'd say let's focus on getting the remainder of the initial figures integrated before tackling that since it will require us to think about how to best expose the functions that do these bandwidth calculations from darshan-parser and libdarshan-util. I can open some issues to think more about those problems.

nawtrey:
I've already worked on this a bit, but I think most of what I've changed will be kept anyways. For now, could we add a second bar graph that shows the total (i.e. uses just `*_F_SLOWEST_RANK_TIME`), then correct it later once we have them broken down into read/write/meta? That's basically what I have at the moment, and the update would just have to update which keys are being used, basically.

darshan-util/pydarshan/tests/test_plot_io_cost.py
```diff
@@ -0,0 +1,139 @@
...
+@pytest.mark.parametrize("mod", ["POSIX", "MPI-IO", "STDIO"])
+@pytest.mark.parametrize(
+    "report, expected_yticks, expected_yticklabels", [
+        (
+            darshan.DarshanReport("examples/example-logs/ior_hdf5_example.darshan"),
+            [0.0, 0.2, 0.4, 0.6, 0.8, 1.0],
+            ["0.0", "0.2", "0.4", "0.6", "0.8", "1.0"],
+        ),
+        (
+            darshan.DarshanReport("examples/example-logs/sample.darshan"),
```
tylerjereddy, line 102: Here and elsewhere--I thought `sample.darshan` had buggy data in `STDIO_F_WRITE_TIME`? It may not matter, though I wonder about the long-term wisdom of leaning on it for regression tests that are added in new PRs? Or at least those that actually access that field at least.

Have issues been opened to carve a path forward on handling this log and its renamed duplicates? For example, if the Python machinery might eventually raise an exception via a lightweight sanity checker for input logs that might influence a decision to use the log (and its clones) for new testing.

darshan-util/pydarshan/tests/test_plot_io_cost.py
```diff
@@ -0,0 +1,139 @@
...
+            [0, 24, 48, 72, 96, 120],
+            ["0", "24", "48", "72", "96", "120"],
```
tylerjereddy, line 104: Instead of maintaining these typed out lists of identical values can't you do something like this to generate the expected labels for any expected tick list within the test proper:

`expected_yticklabels = [str(i) for i in expected_yticks]`

tylerjereddy (commented):
Things mostly look "ok," though I did add some questions/comments.

Also, is the team "ok" with small values being nearly invisible or do they want log-scaling or an inset that is log-scaled or something?

For `imbalanced-io.darshan` across 3 modules:

imbalanced-io darshan_STDIO
imbalanced-io darshan_POSIX
imbalanced-io darshan_MPI-IO

darshan-util/pydarshan/tests/test_plot_io_cost.py
nawtrey: Yes, it does have a negative value that shouldn't be there. I wasn't worried about it for this particular test, but I see what you mean.

I think you raise some good points, if a sanity checker is implemented this would cause a lot of issues for the current testing suite, and reusing this log is only going to make things worse. I *thought* we had agreed to keep 1 of the logs and just remove the duplicates, but looking back at the messages it seems no hard lines have been drawn. 

I think there are a few things to discuss/issues to open from that discussion:
 1. What to do with the duplicate logs
 2. Whether we should keep `sample.darshan` at all 
 3. Where we want the logs to be stored long-term (everything in the logs repo, a mix like we currently have, etc.)
 4. What we want to do with `/examples` (and its logs) long-term
 5. Whether or not we want to add some sanity checks for logs, and how we want to do that

shanedsnyder:
> I've already worked on this a bit, but I think most of what I've changed will be kept anyways. For now, could we add a second bar graph that shows the total (i.e. uses just `*_F_SLOWEST_RANK_TIME`), then correct it later once we have them broken down into read/write/meta? That's basically what I have at the moment, and the update would just have to update which keys are being used, basically.

I suppose we can at least see how that looks, but you're right it probably is useful to have this until we have more granular slowest time information.

The calculation we've been using in the C code for the 'by_slowest' metric is obviously a little more complicated compared to the average, but here's generally what we do:

1. Sum up the `_F_SLOWEST_RANK_TIME` counter for all shared files -- this gives us total time by slowest for access to shared files (rank == -1)
2. Sum up the cumulative I/O time (read+write+meta) for all independent files (rank != -1) on each rank, then find the rank with the highest cumulative I/O time -- this gives us the slowest rank for accessing independent files

If you sum those 2 values, you get a comprehensive slowest time value that includes both shared files and independent files. Hopefully that's somewhat clear, but I can help sanity check code that you come up with to make sure we're on the same page.

shanedsnyder:
> Things mostly look "ok," though I did add some questions/comments.
> 
> Also, is the team "ok" with small values being nearly invisible or do they want log-scaling or an inset that is log-scaled or something?
> 

My vote would be to avoid using log-scale in this case. I think it might be visually helpful for users to be able to straight up compare I/O time vs application runtime without having to consider how log axis might skew things visually. In cases where I/O time is negligible and doesn't show on these charts, it often means that users don't have pressing I/O issues to address (though that's not always true).

darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
```diff
@@ -0,0 +1,119 @@
...
+    if mod_key not in ["POSIX", "MPI-IO", "STDIO"]:
+        # TODO: expand the scope of this function
+        # to include HDF5 module
+        raise NotImplementedError(f"{mod_key} module is not supported.")
+    # calculate the run time from the report metadata
+    runtime = report.metadata["job"]["end_time"] - report.metadata["job"]["start_time"]
+    if runtime == 0:
+        # for cases where runtime is < 1, just set it
+        # to 1 like the original perl code
+        runtime = 1
```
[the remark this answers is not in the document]
shanedsnyder: Yeah, right, most practical cases would definitely run longer than a second, though I guess having start/end times at a second granularity probably poses other problems too, depending on what analysis the user is trying to do. I can't think of any reasons off the top of my head not to switch to something more granular (like a `struct timeval`, which has microsecond granularity), it's not really used for much other than to provide some brief metadata about when the job started/ended. Maybe this awkwardness is enough of an incentive to make this change, though.

At any rate, I think the analysis code will still have to be mindful that start_time and end_time could be identical, in which case I guess we should keep rounding up to 1 second wherever it makes sense.

darshan-util/pydarshan/tests/test_plot_io_cost.py
shanedsnyder: Yeah, I think it would be better to avoid using this log for most testing purposes. It may still be useful to have around as an example log with bad data (in case we come up with something to sanity check log data that this could be used as input for), but other than that it may be dangerous. I'd suggest we just rename it (`bad-counters.darshan` or something like that) and remove all duplicates, and switch tests that use it over to something less problematic.

shanedsnyder:
Just to recap what we discussed in a meeting today, but I think the best plan of action to move forward with minimal distractions here would be to just keep the I/O cost graphs based on the average calculations for now. We can think more about the best way to provide I/O cost measurements w.r.t. the slowest process later on once we've made more general progress on the v1 milestone.

carns:
> My vote would be to avoid using log-scale in this case. I think it might be visually helpful for users to be aMy vote would be to avoid using log-scale in this case. I think it might be visually helpful for users to be able to straight up compare I/O time vs application runtime without having to consider how log axis might skew things visually.ble to straight up compare I/O time vs application runtime without having to consider how log axis might skew things visually.

I definitely agree with this.  For the purposes of the summary tool its more important to quickly convey "this value is small" than it is to be able to see it's actual value.  Probably a user should move on to a different analysis tool (jupyter notebook or whatever) if they need to drill down.

darshan-util/pydarshan/tests/test_plot_io_cost.py
```diff
@@ -0,0 +1,139 @@
...
+    fig = plot_io_cost(report=report, mod_key=mod)
+    for i, ax in enumerate(fig.axes):
+        # there are only 2 axes, the first being the "raw" data
+        # and the second being the normalized data (percent)
+        actual_ylims = ax.get_ylim()
+        if i == 0:
+            assert_array_equal(actual_ylims, expected_ylims)
+        else:
+            # normalized data is always the same
+            assert_array_equal(actual_ylims, [0.0, 100.0])
```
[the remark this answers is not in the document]
nawtrey: I changed this to use `assert_allclose`

darshan-util/pydarshan/tests/test_plot_io_cost.py
nawtrey: I've purged `sample.darshan` from the tests here in lieu of `sample-badost.darshan`.

darshan-util/pydarshan/tests/test_plot_io_cost.py
nawtrey: I added this simplification in the latest commit(s).

nawtrey:
Here are the 3 plots (`MPI-IO`, `POSIX`, `STDIO`) for `ior_hdf5_example.darshan`, with the updated x-axis labels:

ior_hdf5_example darshan_MPI-IO
ior_hdf5_example darshan_POSIX
ior_hdf5_example darshan_STDIO

darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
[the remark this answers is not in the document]
tylerjereddy: That's true, though I'd guess the probability of `0` cases, even for quick and dirty debugging logs, should drop substantially if you improve the resolution by six orders of magnitude, assuming you can genuinely get microsecond resolution.

Indeed, we might even need a synthetically generated log to hit the `0` code path at that resolution. I guess the Python code would then need to use a log format version to decide on which resolution is in play.

tylerjereddy:
Ok, I read through the full code diff and the conversations one more time, and ran the latest version of the branch locally and it seems "ok" to me. The other reviewers also seem "ok" with this as a starting point, with more to come later with the "by-slowest" stuff. The patch diff coverage is also 100 % with CI passing.

So, I'll merge it, thanks.

As an aside, it seems to me like some extra shims for looping over modules and/or handling missing modules may be hard to avoid when this infrastructure is added to the job summary tool. Maybe good to think about the abstractions that might be helpful in the future for that API as more elements are added to the report, so that we can minimize "custom shims" in the central job summary code.

PR #553, opened 2021-11-04
TST: Add tests for `plot_opcounts` and `plot_access_histogram`

nawtrey:
* Add testing modules with some initial tests for
`plot_opcounts.plot_opcounts()` and
`plot_access_histogram.plot_access_histogram()`
since they will eventually be used in the Darshan
summary report


nawtrey:
It was mentioned recently that plotting functions in `darshan/experimental/plots`would be used for #465, and they need to be tested. This PR is a decent starting point, it includes some basic tests for the 2 modules we will need.

Here is the output of `pytest --cov-report term-missing --cov=darshan/experimental/plots`:
```
---------- coverage: platform darwin, python 3.9.7-final-0 -----------
Name                                                  Stmts   Miss  Cover   Missing
-----------------------------------------------------------------------------------
darshan/experimental/plots/__init__.py                    3      0   100%
darshan/experimental/plots/heatmap_handling.py           82      2    98%   12, 18
darshan/experimental/plots/plot_access_histogram.py      36      1    97%   29
darshan/experimental/plots/plot_dxt_heatmap.py          100      1    99%   11
darshan/experimental/plots/plot_dxt_heatmap2.py          71     65     8%   33-129
darshan/experimental/plots/plot_opcounts.py              58      8    86%   35, 100-107
-----------------------------------------------------------------------------------
TOTAL                                                   350     77    78%
```

The uncovered lines 100-107 for `plot_opcounts.py`  are for some unused code in the plotting function that can probably be removed. The remaining line is for an error that is raised if `darshan.enable_experimental()` is not called, and the same is true for the single line in `plot_access_histogram()`.

I think more tests will likely be added to these modules as we move forward. For example, there is an issue with both plotting functions where the labels are being cutoff for some cases, which a fix + regression test can be written for. Before I did that I wanted to make sure to nail down the bar graph heights so we are sure we aren't changing the actual results when we make changes.

nawtrey:
I think we will have to add tests for `darshan/experimental/aggregators/mod_agg_iohist.py ` and `darshan/experimental/aggregators/agg_ioops.py ` as well since the plotting functions rely on them. At the moment their coverage is good, but there aren't any tests targeting them specifically. 

Output for `pytest --cov-report term-missing --cov=darshan/experimental/aggregators`:
```
---------- coverage: platform darwin, python 3.9.7-final-0 -----------
Name                                                       Stmts   Miss  Cover   Missing
----------------------------------------------------------------------------------------
darshan/experimental/aggregators/__init__.py                   0      0   100%
darshan/experimental/aggregators/agg_ioops.py                 39      0   100%
darshan/experimental/aggregators/create_dxttimeline.py        43     41     5%   13-114
darshan/experimental/aggregators/create_sankey.py             43     41     5%   10-86
darshan/experimental/aggregators/create_time_summary.py       21     19    10%   9-50
darshan/experimental/aggregators/create_timeline.py           98     90     8%   8-27, 30-34, 37-51, 55-71, 75-117, 121-133, 150-219
darshan/experimental/aggregators/mod_agg_iohist.py            34      2    94%   17, 28
darshan/experimental/aggregators/name_records_summary.py      11      9    18%   14-26
darshan/experimental/aggregators/print_module_records.py       4      2    50%   15-141
darshan/experimental/aggregators/records_as_dict.py           22     20     9%   15-51
darshan/experimental/aggregators/summarize.py                  8      6    25%   14-26
----------------------------------------------------------------------------------------
TOTAL                                                        323    230    29%
```

tylerjereddy:
I think I've mentioned this before, but when you can easily split a PR with hundreds of lines into smaller PRs that can help me review a bit. I should be able to handle this one, but it is also fairly obvious how it could be split into two PRs according to the modules under test.

It is also worth noting that tests probably shouldn't be restricted to log files in the main repo out of mere convenience vs. using the logs repo files when appropriate. There are examples of iterating over all the logs in the logs repo and picking out individual logs from the logs repo floating around now.

nawtrey:
> I think I've mentioned this before, but when you can easily split a PR with hundreds of lines into smaller PRs that can help me review a bit. I should be able to handle this one, but it is also fairly obvious how it could be split into two PRs according to the modules under test.

I was going to split this one up but I decided to keep it in a single PR because the tests for each module are very similar. They are more or less the same tests with different data, but try to be more cautious next time.

> It is also worth noting that tests probably shouldn't be restricted to log files in the main repo out of mere convenience vs. using the logs repo files when appropriate. There are examples of iterating over all the logs in the logs repo and picking out individual logs from the logs repo floating around now.

Do you think this is a case where it would be appropriate to use the logs repo? I figured that since I got what I needed from the local files, it wasn't really necessary to use the logs repo.

darshan-util/pydarshan/tests/test_plot_opcounts.py
```diff
@@ -0,0 +1,112 @@
...
+                    actual_MPIIO_IND_vals.append(patch.get_height())
+                elif (i > 13) & (i <= 20):
+                    actual_MPIIO_COL_vals.append(patch.get_height())
+                elif i > 20:
+                    actual_STDIO_vals.append(patch.get_height())
+
+        assert_array_equal(actual_POSIX_vals, expected_POSIX_vals)
+        assert_array_equal(actual_MPIIO_IND_vals, expected_MPIIO_IND_vals)
+        assert_array_equal(actual_MPIIO_COL_vals, expected_MPIIO_COL_vals)
+        assert_array_equal(actual_STDIO_vals, expected_STDIO_vals)
```
tylerjereddy, line 112: I don't know if we gain much value in splitting the actual and expected values in to separate lists as opposed to unified lists. You might get a "label" on the module involved in a failure, though the global comparison would give you the global % difference across all fields/modules when the failure happens.

One reason I'm thinking about that is unification/deduplication of the two `test_bar_heights()` functions would likely be far easier if they just placed actual and expected values in single lists/arrays. The `fig.axes` conditionals would all disappear then, etc.

A little tricky maybe, though I also note that I don't see any cases for this "duplicated" test where two different modules actually produce different results for the same log file, at least at the moment.

darshan-util/pydarshan/tests/test_plot_access_histogram.py
```diff
@@ -0,0 +1,107 @@
...
+                "examples/example-logs/sample.darshan",
+                "POSIX",
+                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [4, 14, 0, 0, 0, 0, 0, 0, 16384, 0],
+            ),
+            (
+                "examples/example-logs/sample.darshan",
+                "MPI-IO",
+                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [4, 14, 0, 0, 0, 0, 0, 0, 16384, 0],
```
tylerjereddy, line 84: hmm, duplicating the same file with different module and identical expected result is perhaps a bit awkward--are the two modules guaranteed to always have the same result for this function?

If that isn't necessarily a guarantee that might be a dataset you don't have access to in the main repo--I know you mentioned having all that you needed in the main repo, so maybe the guarantee is true?

tylerjereddy (commented):
> They are more or less the same tests with different data

Isn't that the normal use case for parametrization? For the first "duplicated" test, the below seems to flush the same code path after deleting the other two in my hands. There would be two extra test runs we don't need, but could probably be filtered out with some creativity.


```diff
--- /dev/null
+++ b/darshan-util/pydarshan/tests/test_plot_exp_common.py
@@ -0,0 +1,50 @@
+import pytest
+from numpy.testing import assert_array_equal
+
+import darshan
+from darshan.experimental.plots.plot_opcounts import plot_opcounts
+from darshan.experimental.plots.plot_access_histogram import plot_access_histogram
+
+darshan.enable_experimental()
+
+@pytest.mark.parametrize(
+    "log_path, mod",
+    [
+        ("examples/example-logs/dxt.darshan", "POSIX"),
+        ("examples/example-logs/ior_hdf5_example.darshan", "POSIX"),
+        ("examples/example-logs/ior_hdf5_example.darshan", "MPI-IO"),
+        ("examples/example-logs/sample-badost.darshan", "POSIX"),
+        ("examples/example-logs/sample.darshan", "POSIX"),
+        ("examples/example-logs/sample.darshan", "MPI-IO"),
+    ],
+)
+@pytest.mark.parametrize(
+    "func, expected_xticks, expected_xticklabels",
+    [
+     (plot_access_histogram, range(10), 
+     ["0-100", "101-1K", "1K-10K", "10K-100K", "100K-1M",
+     "1M-4M", "4M-10M", "10M-100M", "100M-1G", "1G+"]),
+     (plot_opcounts, range(7), 
+      ['Read', 'Write', 'Open', 'Stat', 'Seek', 'Mmap', 'Fsync']),
+    ]
+    )
+def test_xticks_and_labels(log_path, func, expected_xticks, expected_xticklabels, mod):
+    # check the x-axis tick mark locations and
+    # labels
+    report = darshan.DarshanReport(log_path)
+
+    try:
+        fig = func(report=report, mod=mod)
+    except TypeError:
+        fig = func(report=report)
+
+    # retrieve the x-axis tick mark locations and labels
+    # from the output figure object
+    ax = fig.axes[0]
+    actual_xticks = ax.get_xticks()
+    actual_xticklabels = [tl.get_text() for tl in ax.get_xticklabels()]
+
+    # the expected x-axis tick marks and locations
+    # should be the same for all cases
+    assert_array_equal(actual_xticks, expected_xticks)
+    assert_array_equal(actual_xticklabels, expected_xticklabels)

```

tylerjereddy:
Regardless of what happens with the balance between duplication and over-architecture, the  code coverage increases do look good based on local double-checking.

darshan-util/pydarshan/tests/test_plot_opcounts.py
nawtrey: Yeah, I think it makes sense to merge the lists and the tests together.

darshan-util/pydarshan/tests/test_plot_access_histogram.py
nawtrey: > hmm, duplicating the same file with different module and identical expected result is perhaps a bit awkward--are the two modules guaranteed to always have the same result for this function?

No, I just happened to use ones that have the same result for both modules. I'll make sure to change that, it looks like `tests/input/sample-dxt-simple.darshan` has a slight difference between `POSIX` and `MPIIO`. Additionally `skew-app.darshan` and `imbalanced-io.darshan` from the logs repo have pretty notable differences between the modules.

nawtrey: I've added `imbalanced-io.darshan` (from the logs repo) and `sample-dxt-simple.darshan` to the list of logs being tested.

darshan-util/pydarshan/tests/test_plot_opcounts.py
nawtrey: I was able to merge the lists and tests together into a single module, and remove those conditionals.

nawtrey:
To use a single log from the logs repo, I borrowed the fixture from gh-397. The way I worked a single log (from the logs repo) into `test_bar_heights` seems a bit awkward to me, but it works.

darshan-util/pydarshan/tests/test_plot_exp_common.py
```diff
@@ -0,0 +1,172 @@
...
+            "examples/example-logs/sample.darshan",
+            None,
+            plot_opcounts,
+            [0, 16402, 2049, 0, 16404, 0, 0, 0, 18, 0, 0, 0, 0, 0,
+            0, 16384, 2048, 0, 0, 0, 0, 0, 74, 129, 0, 0, 0, 0],
+        ),
+    ],
+)
+@pytest.mark.skipif(not pytest.has_log_repo, # type: ignore
+                    reason="missing darshan_logs")
```
tylerjereddy, line 150: Why skip every single test case in `test_bar_heights()` when the logs repo is absent? I think the design we'd want is to only skip the cases with filepaths we don't have access to, but the in-repo files we currently always have access to.

`pytest` provides facilities for more fine-grained skipping, and also for imperative skipping in the test proper when that is more convenient.

darshan-util/pydarshan/tests/test_plot_exp_common.py
```diff
@@ -0,0 +1,172 @@
...
+@pytest.mark.parametrize("filename", ["imbalanced-io.darshan"])
+def test_bar_heights(log_path, mod, fig_func, expected_heights, select_log_repo_file):
+    # check bar graph heights
+
+    if log_path is None:
```
tylerjereddy, line 155: This doesn't seem quite like the right design yet, since this test has a complex/stacked parametrization pattern.

- how do you add another case from the logs repo sustainably now that you've used `None` as a placeholder for "file in logs repo" in parameters above, without bolting on tons of extra logic in the test proper?

Perhaps something like this might be more sustainable (just add more logs repo filenames to params list when you want); I didn't add the extra test skipping logic needed from comment above

```diff
--- a/darshan-util/pydarshan/tests/test_plot_exp_common.py
+++ b/darshan-util/pydarshan/tests/test_plot_exp_common.py
@@ -59,7 +59,7 @@ def test_xticks_and_labels(log_path, func, expected_xticks, expected_xticklabels
 
 
 @pytest.mark.parametrize(
-    "log_path, mod, fig_func, expected_heights",
+    "filename, mod, fig_func, expected_heights",
     [
         (
             "examples/example-logs/dxt.darshan",
@@ -103,14 +103,14 @@ def test_xticks_and_labels(log_path, func, expected_xticks, expected_xticklabels
         # more obvious case where "POSIX" and
         # "MPIIO" modules show different results
         (
-            None,
+            "imbalanced-io.darshan",
             "POSIX",
             plot_access_histogram,
             [18, 2492, 14679, 0, 50486, 186, 0, 0, 0,
             0, 43, 301, 2, 0, 50486, 0, 0, 0, 0, 0],
         ),
         (
-            None,
+            "imbalanced-io.darshan",
             "MPI-IO",
             plot_access_histogram,
             [11, 2492, 2, 0, 0, 0, 0, 410, 86, 0, 2526,
@@ -148,13 +148,15 @@ def test_xticks_and_labels(log_path, func, expected_xticks, expected_xticklabels
 )
 @pytest.mark.skipif(not pytest.has_log_repo, # type: ignore
                     reason="missing darshan_logs")
-@pytest.mark.parametrize("filename", ["imbalanced-io.darshan"])
-def test_bar_heights(log_path, mod, fig_func, expected_heights, select_log_repo_file):
+def test_bar_heights(filename, mod, fig_func, expected_heights, select_log_repo_file):
     # check bar graph heights
 
-    if log_path is None:
+    logs_repo_path = select_log_repo_file
+    if logs_repo_path is not None:
         # for logs repo cases
         log_path = select_log_repo_file
+    else:
+        log_path = filename
 
     report = darshan.DarshanReport(log_path)
     fig, ax = plt.subplots()
```

tylerjereddy: (note that `select_log_repo_file` will also fail catastrophically if the logs repo is absent, so careful with the skipping design on that point)

tylerjereddy: As another hint, since this may be challenging to do elegantly, using `pytest.has_log_repo` in the test body proper, as part of conditional logic, may prove helpful for a sustainable design, possibly in addition to skips applied elsewhere, because the presence of the logs repo and working with a log repo file are two different things to check for.

darshan-util/pydarshan/tests/test_plot_exp_common.py
nawtrey: Yeah this was a bit too coarse. I changed to marking the individual test cases to skip which seems much more appropriate. Locally it works for me with/without the logs repo available.

darshan-util/pydarshan/tests/test_plot_exp_common.py
```diff
@@ -0,0 +1,172 @@
...
+        ),
+        (
+            "examples/example-logs/sample-badost.darshan",
+            None,
+            plot_opcounts,
+            [0, 131072, 2048, 2048, 131072, 0, 2048, 0, 0, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 34816, 97, 6144, 0, 0, 0, 2056],
+        ),
+        (
+            "examples/example-logs/sample.darshan",
```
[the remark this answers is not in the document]
nawtrey: I've updated the test cases that used `sample.darshan` so it is no longer present in the test suite. They have been replaced by similar tests using `shane_macsio_id29959_5-22-32552-7035573431850780836_1590156158.darshan`.

darshan-util/pydarshan/tests/test_plot_exp_common.py
```diff
@@ -0,0 +1,176 @@
...
+)
+def test_bar_heights(filename, mod, fig_func, expected_heights, select_log_repo_file):
+    # check bar graph heights
+
+    logs_repo_path = select_log_repo_file
+    if logs_repo_path is not None:
+        # for logs repo cases
+        log_path = select_log_repo_file
+    else:
+        log_path = filename
```
tylerjereddy, line 161: Isn't this logic unnecessarily complex? My rough draft was mostly intended to inspire something much better.

```diff
--- a/darshan-util/pydarshan/tests/test_plot_exp_common.py
+++ b/darshan-util/pydarshan/tests/test_plot_exp_common.py
@@ -152,13 +152,7 @@ def test_xticks_and_labels(log_path, func, expected_xticks, expected_xticklabels
 )
 def test_bar_heights(filename, mod, fig_func, expected_heights, select_log_repo_file):
     # check bar graph heights
-
-    logs_repo_path = select_log_repo_file
-    if logs_repo_path is not None:
-        # for logs repo cases
-        log_path = select_log_repo_file
-    else:
-        log_path = filename
+    log_path = select_log_repo_file or filename
 
     report = darshan.DarshanReport(log_path)
     fig, ax = plt.subplots()

```

nawtrey: Yes, this is much simpler so I've made this adjustment with the latest commit.

tylerjereddy:
gh-566 describes an approach I was trying to inspire a bit; for now I'll just double check this in its current state..

PR #554, opened 2021-11-04
MAINT: Remove heatmap figure title

### LINKED ISSUE #537: MAINT: Remove title from heat map figure
As discussed in today's meeting, the figure title will be redundant in the context of a Darshan summary report since the summary report code will have its own title anyways. The only additional information (like the module) can be included there with ease, if need be, so let's just remove it on the figure-side of things. The only cases where a user would want a title it should be simple enough to add after generating the figure object anyways.

nawtrey:
* Fixes #537


tylerjereddy (approved):
LGTM

Issue #538, opened 2021-10-20
ENH: add `DXT_MPIIO` heat map figure to darshan summary report


nawtrey:
Once #536 is closed, we should be able to add another heatmap for the `DXT_MPIIO` module with ease, given the report has it. The error message logic will likely have to be changed so we don't get redundant "DXT tracing is not enabled" messages, and the figure titles (on the report side) should probably be updated to reflect which module is being used (i.e. `Heat Map: DXT_POSIX`).

nawtrey:
This issue was fixed via gh-570.

PR #560, opened 2021-11-09
ENH: Add `DXT_MPIIO` support to `plot_dxt_heatmap.plot_heatmap()`

### LINKED ISSUE #536: ENH: add `DXT_MPIIO` support to `plot_dxt_heatmap.plot_heatmap()`
As demonstrated in today's meeting, the heat map code already has the capability to plot `DXT_MPIIO` record data with some simple changes. It currently accepts a list of modules to aggregate (`mods`), but I think that can be removed/changed to specify which DXT module to plot, if available, since it doesn't look like we will be aggregating by module moving forward.

nawtrey:
`plot_heatmap()` changes:
    - Change "mods" parameter to "mod" to reflect the change to a single module key
    - Change "DXT" module check to allow for any module containing "DXT"  
    - Add check for input module to verify if it is in the input report

* Change "mods" parameter to "mod" for
`get_single_df_dict()` and `get_aggregate_data()`

* Update documentation for functions to reflect the parameter change

* Remove iteration over modules from
`heatmap_handling.get_single_df_dict()`

* Update documentation examples to reflect the new
data structures since aggregating data across
modules is no longer being considered

* Update `plot_heatmap()` and supporting function
call signatures to use "mod" parameter and single
key instead of list of module keys

* Correct data structures for expected
data in `test_get_rd_wr_dfs_no_write`

* Remove module key assertions in `test_get_rd_wr_dfs_no_write`
 since these are no longer stored

* Remove `KeyError` for `test_get_aggregate_data`
since it no longer applies for `DXT_MPIIO` cases

* Update `NotImplementedError` in `test_plot_heatmap`
to match the new error message

* Closes #536

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
@@ -320,15 +320,16 @@ def plot_heatmap(
     Raises
     ------
 
-    NotImplementedError: if a DXT module is not input (i.e. "DXT_POSIX") or if
-    the module is not in the input ``DarshanReport``.
+    NotImplementedError: if a DXT module is not input (i.e. "DXT_POSIX").
+
+    ValueError: if the input module is not in the ``DarshanReport``.
 
     """
     if "DXT" not in mod:
-        raise NotImplementedError(f"Only DXT modules are supported.")
+        raise NotImplementedError("Only DXT modules are supported.")
 
     if mod not in report.modules:
-        raise NotImplementedError(f"Module {mod} not found in DarshanReport.")
+        raise ValueError(f"Module {mod} not found in DarshanReport.")
 
     # aggregate the data according to the selected modules and operations
     agg_df = heatmap_handling.get_aggregate_data(report=report, mod=mod, ops=ops)
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
@@ -291,7 +291,7 @@ def adjust_for_colorbar(jointgrid: Any, fig_right: float, cbar_x0: float):
 
 def plot_heatmap(
     report: darshan.DarshanReport,
-    mods: Sequence[str] = ["DXT_POSIX"],
+    mod: str = "DXT_POSIX",
     ops: Sequence[str] = ["read", "write"],
     xbins: int = 200,
 ) -> Any:
@@ -303,8 +303,8 @@ def plot_heatmap(
 
     report: a ``darshan.DarshanReport``.
 
-    mods: a sequence of keys designating which Darshan modules to use for
-    data aggregation. Default is ``["DXT_POSIX"]``.
+    mod: the DXT module to do analysis for (i.e. "DXT_POSIX"
+    or "DXT_MPIIO"). Default is ``"DXT_POSIX"``.
 
     ops: a sequence of keys designating which Darshan operations to use for
     data aggregation. Default is ``["read", "write"]``.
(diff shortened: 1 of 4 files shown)


nawtrey:
An unfortunately messy PR, mostly due to a change in call signatures for multiple functions. This will allow us to plot a heatmap for `DXT_MPIIO` and add it to the Darshan summary report (issue #538), if it is available. It also removes any ability to aggregate data from multiple modules, which simplified some of the code and data structures being produced. 

There is a single line that is not covered by tests (line 328 in `plot_dxt_heatmap.py`) that can be addressed here or in a follow up. I should be able to add a test case that attempts a non-DXT module, which should trigger the appropriate error.

nawtrey:
I've corrected the CI failure which was due to a type checking issue. There was a complex data type that I failed to update, and it somehow eluded me despite many local runs of `mypy darshan`.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -320,16 +320,18 @@ def plot_heatmap(
...
+    NotImplementedError: if a DXT module is not input (i.e. "DXT_POSIX") or if
+    the module is not in the input ``DarshanReport``.
 
     """
-    if "DXT_POSIX" not in mods:
-        # TODO: for the moment reject any cases that don't input "DXT_POSIX"
-        # until we can properly aggregate the data
-        raise NotImplementedError("DXT_POSIX module is required.")
+    if "DXT" not in mod:
+        raise NotImplementedError(f"Only DXT modules are supported.")
```
tylerjereddy, line 328: nit: unused f-string

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -320,16 +320,18 @@ def plot_heatmap(
...
+
+    if mod not in report.modules:
+        raise NotImplementedError(f"Module {mod} not found in DarshanReport.")
```
tylerjereddy, line 331: Hmm, is something really not implemented here? It sound like maybe some data or values are not available in the report, so an error type reflecting that situation might be more appropriate?

tylerjereddy (commented):
Mostly seems "ok," I added some small comments. We can probably lean on the decent test coverage and the fact that you wrote the original source code.

Things that may be useful:
- bump up the coverage to 100 % if it only takes injecting the right log file somewhere
-  the tests actually changed less than I thought, which perhaps suggests to me that we don't have any cases where the MPIIO variant has a different dataset than the POSIX variant--this might be a good case/test to have to increase confidence that both modules are being handled correctly/separately as needed

tylerjereddy:
Maybe also edit the original comment to say this fixes gh-536, if I understand correctly. That will help me remember to close that issue when merging, since auto-close doesn't happen for `pydarshan-devel` merges.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
nawtrey: I agree that `NotImplementedError` isn't really appropriate. I think for a case like this a custom exception, like a Darshan-specific `ModuleNotFoundError`, would be nice to have. I'll change this to a `ValueError` for now I think.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
nawtrey: I fixed this with the latest round of revisions.

nawtrey:
> * bump up the coverage to 100 % if it only takes injecting the right log file somewhere

I've added to `test_plot_heatmap()` so that it covers as much of the module as we can. 

> * the tests actually changed less than I thought, which perhaps suggests to me that we don't have any cases where the MPIIO variant has a different dataset than the POSIX variant--this might be a good case/test to have to increase confidence that both modules are being handled correctly/separately as needed

I think the issue is that there are no tests running `DXT_MPIIO` at the moment. The ones that do (i.e. `test_plot_dxt_heatmap.test_plot_heatmap()`) are not checking the values in any way, so that would explain why there are no errors raised. `test_heatmap_handling.test_get_aggregate_data()` does test `DXT_MPIIO` but for the log file it is looking at, it doesn't have any data. So in those cases we don't really see anything. It would probably be a good idea to write a test (or add to a current one) that actually retrieves some of the `DXT_MPIIO` data and checks it.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
nawtrey: I changed this to a `ValueError` for now, but let me know if you think there is a better alternative.

tylerjereddy:
> It would probably be a good idea to write a test (or add to a current one) that actually retrieves some of the DXT_MPIIO data and checks it.

yup

nawtrey:
The latest commit adds a test case to `test_heatmap_handling.py::test_get_aggregate_data`, which covers `DXT_MPIIO` data.

tylerjereddy:
Alright the patch diff coverage is 100 % and CI is passing. The changes seems "ok" I think. Some of the testing is a getting a bit awkward/hard to read with the body of tests basically having to retrieve the name of the log file fed in, then make a decision based on the name of the log file, which to some extent defeats the purpose of parametrization.

I'll merge.. thanks

PR #565, opened 2021-11-17
BENCH: add better log file to `time_record_collection_to_df` for `LUSTRE` performance

nawtrey:
* Add `sample-badost.darshan` to `test_data_files_dxt` since it
contains `LUSTRE` records

* Add `sample-badost.darshan` to `time_record_collection_to_df`
benchmark so the `LUSTRE` record performance can be probed
more accurately

### DIFF
--- a/darshan-util/pydarshan/benchmarks/benchmarks/report.py
+++ b/darshan-util/pydarshan/benchmarks/benchmarks/report.py
@@ -2,7 +2,7 @@ import importlib
 import os
 
 example_logs = importlib.import_module("examples.example-logs")
-from tests.input import test_data_files_dxt
+from tests.input import test_data_files_dxt, test_data_files
 import darshan
 
 
@@ -86,6 +86,8 @@ class RecordCollectionToDF:
         filename = os.path.basename(darshan_logfile)
         if "examples" in darshan_logfile:
             self.logfile = example_logs.example_data_files_dxt[filename]
+        elif "sample-badost" in darshan_logfile:
+            self.logfile = test_data_files[filename]
         else:
             self.logfile = test_data_files_dxt[filename]
 
--- a/darshan-util/pydarshan/benchmarks/benchmarks/report.py
+++ b/darshan-util/pydarshan/benchmarks/benchmarks/report.py
@@ -67,6 +67,7 @@ class RecordCollectionToDF:
     params = [["examples/example-logs/ior_hdf5_example.darshan",
                "examples/example-logs/dxt.darshan",
                "tests/input/sample-dxt-simple.darshan",
+               "tests/input/sample-badost.darshan",
               ],
               ["POSIX",
                "STDIO",
--- a/darshan-util/pydarshan/tests/input/__init__.py
+++ b/darshan-util/pydarshan/tests/input/__init__.py
@@ -6,6 +6,8 @@ def installed_path(filename):
 
 test_data_files_dxt = {
     "sample-dxt-simple.darshan": installed_path("sample-dxt-simple.darshan"),
-    "sample-badost.darshan": installed_path("sample-badost.darshan")
 }
 
+test_data_files = {
+    "sample-badost.darshan": installed_path("sample-badost.darshan")
+}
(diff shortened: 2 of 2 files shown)


nawtrey:
This adds a log file that contains 2048 `LUSTRE` records because there is only 1 other log file that has them, and it only has 1 record. So with this log we should get a better performance comparison.

darshan-util/pydarshan/tests/input/__init__.py
```diff
@@ -6,5 +6,6 @@ def installed_path(filename):
 
 test_data_files_dxt = {
     "sample-dxt-simple.darshan": installed_path("sample-dxt-simple.darshan"),
+    "sample-badost.darshan": installed_path("sample-badost.darshan")
```
tylerjereddy, line 9: Why would we add a file without `DXT` records to a test dictionary of files that contain `DXT` records, as both the `test_data_files_dxt` name and its usage in the code implies?

darshan-util/pydarshan/benchmarks/benchmarks/report.py
```diff
@@ -67,6 +67,7 @@ class RecordCollectionToDF:
     params = [["examples/example-logs/ior_hdf5_example.darshan",
                "examples/example-logs/dxt.darshan",
                "tests/input/sample-dxt-simple.darshan",
+               "tests/input/sample-badost.darshan",
```
tylerjereddy, line 70: And how much longer does the benchmark take now? Otherwise, the reviewer will need to run it themselves to check.

darshan-util/pydarshan/tests/input/__init__.py
tylerjereddy: (there are better ways to add files to the benchmark suite for example, that don't involve masquerading as a DXT containing file)

tylerjereddy:
one other comment is that it may be good to check with the team re: why the file name has "bad" in it?

darshan-util/pydarshan/tests/input/__init__.py
nawtrey: That's a good point. I could create a new dictionary. It will complicate the setup logic a little but that's not a big deal.

darshan-util/pydarshan/benchmarks/benchmarks/report.py
nawtrey: The output for `time asv run -b "time_record_collection_to_df"`, ran locally on my Mac:
```
asv run -b "time_record_collection_to_df"  267.04s user 67.57s system 167% cpu 3:20.34 total
```

darshan-util/pydarshan/tests/input/__init__.py
nawtrey: The latest commit splits the newly added log into its own data file dictionary.

darshan-util/pydarshan/benchmarks/benchmarks/report.py
tylerjereddy: > how much longer

nawtrey: On `pydarshan-devel`:
```
asv run -b "time_record_collection_to_df"  202.47s user 52.02s system 162% cpu 2:36.31 total
```

So ~44 seconds longer.

tylerjereddy:
This seems "ok" now, thanks.

That  is indeed a big slowdown, normally more than one would allow for practical reasons. I'll let it through since we don't run the benchmarks that regularly yet and the code is still pretty slow in a lot of places.

`sample-badost.darshan` takes more than 60 ms on `POSIX` and > 20 ms on `LUSTRE`, which is already much slower than anything else in that specific benchmark.

You could also provide the file via the logs repo as Shane suggested on Slack. Indeed, you don't actually have to remove any files from the main repo to start experimenting with using those same files from the logs repo. For the benchmark suite, that would also likely mean starting to experiment with PyPI releases of the logs project so it can be pulled in that way, which may be a good learning experience (one of the more complicated tasks is release management). For the logs repo you only need a source distribution, which makes it a fair bit easier.

PR #571, opened 2021-12-01
WIP, ENH: Add I/O cost figure to Darshan Summary Report

nawtrey:
* Add I/O cost figure into new report
section "Cross-Module Comparisons"

* Add `plt.close()` to `plot_io_cost()`
so figures are properly closed

* Add checks to tests `test_summary.test_main_without_args()`
and `test_summary.test_main_all_logs_repo_files()` to ensure
I/O cost figures are being generated when the relevant modules
are present in the report

* Add check to `test_summary.test_main_without_args()` which
counts the number of `img` tags in the HTML report to regression
guard the figure counts in the reports

* Contributes to #465

### DIFF
--- a/darshan-util/pydarshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/tests/test_summary.py
@@ -111,11 +111,7 @@ def test_main_without_args(tmpdir, argv, expected_img_count):
                         assert "Heat map is not available for this job" in report_str
 
                     # check that expected number of figures are found
-                    actual_img_count = 0
-                    for txt in report_str.split():
-                        if "img" in txt:
-                            actual_img_count += 1
-                    assert actual_img_count == expected_img_count
+                    assert report_str.count("img") == expected_img_count
 
                     # check if I/O cost figure is present
                     for mod in report.modules:
--- a/darshan-util/pydarshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/tests/test_summary.py
@@ -69,15 +69,15 @@ def test_main_with_args(tmpdir, argv):
 
 
 @pytest.mark.parametrize(
-    "argv", [
-        [os.path.abspath("./tests/input/noposix.darshan")],
-        [os.path.abspath("./tests/input/noposix.darshan"), "--output=test.html"],
-        [os.path.abspath("./tests/input/sample-dxt-simple.darshan")],
-        [os.path.abspath("./tests/input/sample-dxt-simple.darshan"), "--output=test.html"],
-        [None],
+    "argv, expected_img_count", [
+        ([os.path.abspath("./tests/input/noposix.darshan")], 1),
+        ([os.path.abspath("./tests/input/noposix.darshan"), "--output=test.html"], 1),
+        ([os.path.abspath("./tests/input/sample-dxt-simple.darshan")], 3),
+        ([os.path.abspath("./tests/input/sample-dxt-simple.darshan"), "--output=test.html"], 3),
+        ([None], 0),
     ]
 )
-def test_main_without_args(tmpdir, argv):
+def test_main_without_args(tmpdir, argv, expected_img_count):
     # test summary.main() by running it without a parser
 
     with mock.patch("sys.argv", [""] + argv):
@@ -110,6 +110,17 @@ def test_main_without_args(tmpdir, argv):
                         # check that help message is present
                         assert "Heat map is not available for this job" in report_str
 
+                    # check that expected number of figures are found
+                    actual_img_count = 0
+                    for txt in report_str.split():
+                        if "img" in txt:
+                            actual_img_count += 1
+                    assert actual_img_count == expected_img_count
+
+                    # check if I/O cost figure is present
+                    for mod in report.modules:
+                        if mod in ["POSIX", "MPI-IO", "STDIO"]:
+                            assert "I/O Cost" in report_str
         else:
             # if no log path is given expect a runtime error
             # due to a failure to open the file
@@ -150,6 +161,11 @@ def test_main_all_logs_repo_files(tmpdir, log_repo_files):
                     else:
                         # check that help message is present
                         assert "Heat map is not available for this job" in report_str
+
+                    # check if I/O cost figure is present
+                    for mod in report.modules:
+                        if mod in ["POSIX", "MPI-IO", "STDIO"]:
+                            assert "I/O Cost" in report_str
 
 class TestReportData:
 
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -16,7 +16,7 @@ from mako.template import Template
 
 import darshan
 import darshan.cli
-from darshan.experimental.plots import plot_dxt_heatmap
+from darshan.experimental.plots import plot_dxt_heatmap, plot_io_cost
 
 darshan.enable_experimental()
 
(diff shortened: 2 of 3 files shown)


nawtrey:
Here are some example reports I've generated: [example_reports.tar.gz](https://github.com/darshan-hpc/darshan/files/7635518/example_reports.tar.gz)

If @shanedsnyder and @carns could comment on the caption I wrote that would be great, they might have other details in mind that I didn't think of. I also linked to the darshan-util documentation because users might be interested in what "meta" operations might mean for a specific module/API, but that might be unnecessary.

shanedsnyder:
Looks good!

Couple of quick comments:

- I'd drop the `by-average` x-axis label for now and probably just make sure we're clear in the description that the calculation is based on the average across all processes? I know you spent some time generalizing the code there to allow for other metrics, but since we can't technically provide them yet, it might be less confusing to users to avoid making the distinction -- hopefully it's easy enough to comment that out so we can easily add it back once we have that capability.
- The plots are pretty big, relative to the text. Simple enough to zoom in/out with a browser, but I think by default the figures should be reasonably-sized compared to the text, particularly in this case where there's not a whole lot visually to convey.

We should discuss in our meeting more today, but I wonder if these sorts of figures might be easier to to actually jam into a single I/O cost section... it would cut down on a lot of redundant captions, and will additionally give us one spot to really discuss the details of what the plots are meant to convey, caveats, etc. I still think per-module sections are useful, just maybe not ideal for this type of graph.

nawtrey:
> * I'd drop the `by-average` x-axis label for now and probably just make sure we're clear in the description that the calculation is based on the average across all processes? I know you spent some time generalizing the code there to allow for other metrics, but since we can't technically provide them yet, it might be less confusing to users to avoid making the distinction -- hopefully it's easy enough to comment that out so we can easily add it back once we have that capability.

I've opened issue #574 to address that separately. 

> * The plots are pretty big, relative to the text. Simple enough to zoom in/out with a browser, but I think by default the figures should be reasonably-sized compared to the text, particularly in this case where there's not a whole lot visually to convey.

This can be adjusted fairly easily, there is a "figure width" parameter for each figure that I can set to smaller. The default size is 500 pixels, so I can try something smaller. I anticipated we would need some adjustment there but hopefully we can find some nice defaults for each figure type that works. 

> We should discuss in our meeting more today, but I wonder if these sorts of figures might be easier to to actually jam into a single I/O cost section... it would cut down on a lot of redundant captions, and will additionally give us one spot to really discuss the details of what the plots are meant to convey, caveats, etc. I still think per-module sections are useful, just maybe not ideal for this type of graph.

To reiterate my comments in the meeting, I can try it both ways and see if we like it one way over the another, it's easy to keep them in a single section. I probably won't update anything here until we nail down a preference though.

tylerjereddy (commented):
Mostly seems "ok" to me. I'd suggest opening issues (deferring for now) for the other undecided reorganization suggestions from Shane as well.

I tested `snyder_acme.exe_id1253318_9-27-24239-1515303144625770178_2.darshan` and `darshan-apmpi-2nodes-64mpi.darshan` locally, which both produced boring/empty I/O cost plots, but that's working as intended I think.

`imbalanced-io.darshan` also seemed "ok" and does actually have some visible bars for the new sections at least

Similar to my suggestion in gh-570, since our testsuite already automatically generates an html report file for every single log file in the logs repo, it isn't hard to come up with ways to start regression testing the content a bit.

You might for example count the number of html image tags so you don't accidentally have someone removing a figure, or once again just sanity checking headers like this:

```diff
--- a/darshan-util/pydarshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/tests/test_summary.py
@@ -127,6 +127,14 @@ def test_main_all_logs_repo_files(tmpdir, log_repo_files):
                 # verify the HTML file was generated
                 assert os.path.exists(expected_save_path)
 
+                # spot check DXT heatmap presence where
+                # expected
+                with open(expected_save_path) as html_report:
+                    report_str = html_report.read()
+                    if "snyder_acme.exe" in log_filepath or "imbalanced" in log_filepath:
+                        for mod in ["POSIX", "MPI-IO", "STDIO"]:
+                            assert f"{mod} Characteristics" in report_str
+
```

I suspect we should just add some light testing and then merge.

tylerjereddy:
we should also keep an eye on the growth/complexity inside `register_figures()`, though I'm not panicked on that just yet

nawtrey:
Okay, after merging gh-581 I've made some updates here:
- Updated function call
- Removed module handling logic since that is now handled in `plot_io_cost.py`
- The new all-in-one figure is smaller now with a width of 350px instead of 500, which makes the figure text roughly the same size as the report text
- Added some light testing to check figure counts for the local files, and check if I/O cost header is found for cases with the appropriate modules

I've also updated the PR comment to reflect the new changes, and updated the examples I posted.

tylerjereddy:
@shanedsnyder reported an issue with the code on Slack--maybe try to reproduce that, add the log file to the logs repo + regression test, etc. (that should be possible to do separately of this PR since the basic plotting code is already merged)

tylerjereddy:
ok, I guess Shane is saying it is a false alarm; I'll try to use this log file during the next round of code review nonetheless...

tylerjereddy:
Ok, the reproducing log file should be most helpful in producing a regression test at least.

darshan-util/pydarshan/tests/test_summary.py
```diff
@@ -110,6 +110,17 @@ def test_main_without_args(tmpdir, argv):
                         # check that help message is present
                         assert "Heat map is not available for this job" in report_str
 
+                    # check that expected number of figures are found
+                    actual_img_count = 0
+                    for txt in report_str.split():
+                        if "img" in txt:
+                            actual_img_count += 1
+                    assert actual_img_count == expected_img_count
```
tylerjereddy, line 118: I think this is a fair bit clearer/faster/more idiomatic:

```diff
--- a/darshan-util/pydarshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/tests/test_summary.py
@@ -111,11 +111,7 @@ def test_main_without_args(tmpdir, argv, expected_img_count):
                         assert "Heat map is not available for this job" in report_str
 
                     # check that expected number of figures are found
-                    actual_img_count = 0
-                    for txt in report_str.split():
-                        if "img" in txt:
-                            actual_img_count += 1
-                    assert actual_img_count == expected_img_count
+                    assert report_str.count("img") == expected_img_count

```

Apart from that, the bug in the plotting code has a separate issue and I haven't seen any objections from Argonne folks re: the basic layout now, so close to ready I think.

shanedsnyder: Sorry, just now getting caught up on this and #581, but think the end result looks good. No objections here on merging once the bug I reported is fixed.

nawtrey: I've add the suggested change here.

tylerjereddy:
@nawtrey maybe just rebase/force push and I will do a final pass here/check a few plots locally after the CI runs again. It seems Shane was fine with merging once the division bug was fixed.

nawtrey:
Okay, should be ready review now.

darshan-util/pydarshan/tests/test_summary.py
```diff
@@ -151,6 +158,11 @@ def test_main_all_logs_repo_files(tmpdir, log_repo_files):
                         # check that help message is present
                         assert "Heat map is not available for this job" in report_str
 
+                    # check if I/O cost figure is present
+                    for mod in report.modules:
+                        if mod in ["POSIX", "MPI-IO", "STDIO"]:
+                            assert "I/O Cost" in report_str
```
tylerjereddy, line 164: If all 3 modules are present for example, you'll end up doing redundant checks here and above--you only need to check if this string is present once per report, since it only shows up once per report, regardless of the number of matching modules.

tylerjereddy (approved):
Apart from one testing inefficiency that I noted, this seems "ok," so I'll go ahead and merge based on:

- only about 8 seconds slower for single core full test run vs. `pydarshan-devel`
- checking the new summary reports for: `dxt.darshan`, `imbalanced-io.darshan`, and `nonmpi_partial_modules.darshan`; they seemed "ok" apart from the point noted in gh-594.

Issue #540, opened 2021-10-22
MAINT: Remove dtype dead code paths


jakobluettgau:
As discussed I'll prepare a pull request that removes dtype code paths which we can then merge for a an upcoming release that changes the minor version, in line with the conventions mentioned in https://github.com/darshan-hpc/darshan/issues/490#issuecomment-934835779

I'll break this into two pull requests:

- [x]  Pull request 1: Add deprecation warning when dtype is provided (see #572)
- [x]  Pull request 2: Purge of dtype code paths in DarshanReport

PR #572, opened 2021-12-01
Add dtype deprecations warnings to DarshanReport

jakobluettgau:
As discussed in #540 this pull request only adds deprecation warnings to discourage the use of dtype as an argument to pydarshan DarshanReports. In a second pull request changes that actually deprecate the feature are 

- [ ] Check testing
- [ ] Discuss when this actually lands and deprecates dtype
- [ ] Consider adding alternative recommendation (to_df, to_x) to warning message

### DIFF
--- a/darshan-util/pydarshan/darshan/report.py
+++ b/darshan-util/pydarshan/darshan/report.py
@@ -20,6 +20,8 @@ import pandas as pd
 
 import collections.abc
 
+import warnings
+
 import logging
 logger = logging.getLogger(__name__)
 
@@ -314,7 +316,11 @@ class DarshanReport(object):
         self.filename = filename
 
         # Behavioral Options
+        if dtype is not None:
+            warnings.warn('The use of dtype in reports is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
+
         self.dtype = dtype                                  # default dtype to return when viewing records
+
         self.automatic_summary = automatic_summary
         self.lookup_name_records = lookup_name_records
 
@@ -522,6 +528,9 @@ class DarshanReport(object):
             None
         """
 
+        if dtype is not None:
+            warnings.warn('The use dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
+
         self.read_all_generic_records(dtype=dtype)
         self.read_all_dxt_records(dtype=dtype)
         if "LUSTRE" in self.data['modules']:
@@ -545,6 +554,9 @@ class DarshanReport(object):
             None
         """
 
+        if dtype is not None:
+            warnings.warn('The use dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
+
         dtype = dtype if dtype else self.dtype
 
         for mod in self.data['modules']:
@@ -562,6 +574,9 @@ class DarshanReport(object):
         Return:
             None
         """
+
+        if dtype is not None:
+            warnings.warn('The use dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
 
         dtype = dtype if dtype else self.dtype
 
@@ -582,6 +597,10 @@ class DarshanReport(object):
             None
 
         """
+
+        if dtype is not None:
+            warnings.warn('The use dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
+
         unsupported =  ['DXT_POSIX', 'DXT_MPIIO', 'LUSTRE', 'APMPI', 'APXC']
 
         if mod in unsupported:
@@ -660,6 +679,10 @@ class DarshanReport(object):
             None
 
         """
+
+        if dtype is not None:
+            warnings.warn('The use dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
+
         if mod not in self.data['modules']:
             if warnings:
                 logger.warning(f" Skipping. Log does not contain data for mod: {mod}")
@@ -710,6 +733,11 @@ class DarshanReport(object):
             None
 
         """
+
+        if dtype is not None:
+            warnings.warn('The use dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
+
+
         if mod not in self.data['modules']:
             if warnings:
                 logger.warning(f" Skipping. Log does not contain data for mod: {mod}")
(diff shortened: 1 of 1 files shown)


tylerjereddy (commented):
@jakobluettgau Not sure if you forgot to push in the small fix, but something like below works just fine for now. This does indeed produce quite a few deprecation warnings in the full test suite (`247 passed, 2345 warnings in 124.68s (0:02:04)`).

```diff
--- a/darshan-util/pydarshan/darshan/report.py
+++ b/darshan-util/pydarshan/darshan/report.py
@@ -21,6 +21,9 @@ import pandas as pd
 import collections.abc
 
 import warnings
+# TODO: we shouldn't really be shadowing
+# standard library module names
+from warnings import warn
 
 import logging
 logger = logging.getLogger(__name__)
@@ -599,7 +602,7 @@ class DarshanReport(object):
         """
 
         if dtype is not None:
-            warnings.warn('The use dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
+            warn('The use dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
 
         unsupported =  ['DXT_POSIX', 'DXT_MPIIO', 'LUSTRE', 'APMPI', 'APXC']
 
@@ -789,7 +792,7 @@ class DarshanReport(object):
         """
 
         if dtype is not None:
-            warnings.warn('The use dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
+            warn('The use dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x', DeprecationWarning, stacklevel=2)
```

tylerjereddy:
As discussed, we could turn those warnings into errors with `pytest` configuration settings, but that should be for a later PR that does the cleanup on the usages anyway I think.

darshan-util/pydarshan/darshan/report.py
```diff
@@ -563,14 +576,17 @@ def read_all_dxt_records(self, reads=True, writes=True, dtype=None):
...
         dtype = dtype if dtype else self.dtype
 
         for mod in self.data['modules']:
-            self.mod_read_all_dxt_records(mod, warnings=False, reads=reads, writes=writes, dtype=dtype)
+            self.mod_read_all_dxt_records(mod, _warnings=False, reads=reads, writes=writes, dtype=dtype)
 
 
 
-    def mod_read_all_records(self, mod, dtype=None, warnings=True):
+    def mod_read_all_records(self, mod, dtype=None, _warnings=True):
```
tylerjereddy, line 589: Prefixing a parameter name with a `_` is perhaps a bit unusual, though it doesn't really matter. I suppose the convention to use `_` in connection with informal privacy might confuse some.

darshan-util/pydarshan/darshan/report.py
```diff
@@ -295,7 +297,8 @@ class DarshanReport(object):
 
 
     def __init__(self, 
-            filename=None, dtype='numpy', 
+            filename=None, 
+            dtype=None, # to be deprecated
```
tylerjereddy, line 301: I think once you add the warning, as you have here, it is already considered "deprecated," so it is effectively "to be removed" at this point

darshan-util/pydarshan/darshan/report.py
```diff
@@ -314,7 +317,11 @@ def __init__(self,
         self.filename = filename
 
         # Behavioral Options
-        self.dtype = dtype                                  # default dtype to return when viewing records
+        if dtype is not None:
+            warn(f"The use of dtype in reports is deprecated in v3.3.2 and will be removed in v3.4.x", DeprecationWarning, stacklevel=2)
```
tylerjereddy, line 321: I think here and below you can use a plain string since the f-substitution goes unused

darshan-util/pydarshan/darshan/report.py
```diff
@@ -660,15 +680,19 @@ def mod_read_all_apmpi_records(self, mod="APMPI", dtype=None, warnings=True):
             None
 
         """
+
+        if dtype is not None and _warnings:
+            warn(f"The use of the dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x", DeprecationWarning)
```
tylerjereddy, line 685: Why is the deprecation warning for `mod_read_all_apmpi_records` (and many others) gated on `_warnings`, but the deprecation warning for `mod_read_all_records` is not gated on `warnings`?

darshan-util/pydarshan/darshan/report.py
```diff
@@ -698,7 +722,7 @@ def mod_read_all_apmpi_records(self, mod="APMPI", dtype=None, warnings=True):
             self.update_name_records(mod=mod)
 
 
-    def mod_read_all_apxc_records(self, mod="APXC", dtype=None, warnings=True):
+    def mod_read_all_apxc_records(self, mod="APXC", dtype=None, _warnings=True):
```
tylerjereddy, line 725: I assume @shanedsnyder is happy with the backwards incompatible function signature change in support of the deprecation warnings for `dtype`.

tylerjereddy (commented):
This seems pretty close now.. I only needed a small patch to get the tests passing on this branch locally:

```diff
--- a/darshan-util/pydarshan/tests/test_report.py
+++ b/darshan-util/pydarshan/tests/test_report.py
@@ -90,7 +90,7 @@ def test_dxt_mod(caplog, mod: str, expected: str):
     """Invalid/unsupported dxt module cases"""
     report = darshan.DarshanReport("tests/input/sample-dxt-simple.darshan")
 
-    report.mod_read_all_dxt_records(mod, warnings=True)
+    report.mod_read_all_dxt_records(mod, _warnings=True)
     
     for record in caplog.records:
         assert expected in record.message

```

darshan-util/pydarshan/darshan/report.py
shanedsnyder: To be clear, you just mean the change of `warnings` to `_warnings`? Or something else?

If the former, I don't think it's a big deal, but not sure it's really needed (seems like you have some questions on this change above, too).

darshan-util/pydarshan/darshan/report.py
```diff
@@ -901,6 +940,11 @@ def mod_records(self, mod,
             None
 
         """
+
+        if dtype is not None and _warnings:
+            warn(f"The use of the dtype argument is deprecated in v3.3.2 and will be removed in v3.4.x", DeprecationWarning, stacklevel=2)
```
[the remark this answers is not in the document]
shanedsnyder: I would say `v3.3.2.0` for deprecation message (the last version number is for pydarshan to bump independently of Darshan proper releases). It's hard to really say what version number we'd remove in, though, since it kind of depends on our release schedule in Darshan proper. Given that, I'm a little hesitant to commit to a strict schedule -- can we just say a future release?

darshan-util/pydarshan/darshan/report.py
jakobluettgau: Hm, yeah fair point. I have revertet the change, although I think the intention of changing this parameter is actually for a user to stumble and maybe have a look at the documentation that would discourage them from populating this variable if they don't know what they are doing. We added this initially to allow the suppression of warnings by the report objects. But maybe this is in fact bad style to begin with.

darshan-util/pydarshan/darshan/report.py
jakobluettgau: Oh, good catch, yeah these are leftover from some debugging output. Fixed.

PR #573, opened 2021-12-01
MAINT: cleanup of `summary.py`

nawtrey:
* Remove f-string from `register_figures()` documentation

* Small whitespace fix


nawtrey:
Just some small things I noticed while working on the other recent PR's.

tylerjereddy (approved):
Seems "ok," there's probably a linter we could use for extraneous f-strings.

PR #581, opened 2021-12-08
MAINT: Plot multiple modules in single IO cost plot

### LINKED ISSUE #574: MAINT: Drop x-axis label from I/O cost plot
It was suggested in [this comment](https://github.com/darshan-hpc/darshan/pull/571#issuecomment-983982206) that we drop the x-axis label for the I/O cost plot. I had originally kept it in so that when the "by-slowest" category was added, we wouldn't have to make any adjustments to the "by-average" category, but removing/adding an axis label is simple enough so it's not a big deal either way.

nawtrey:
* Change `plot_io_cost()` to iterate over all supported modules
and plot them in a single figure

* Correct `plot_io_cost()` call signatures and expected data
for `test_plot_io_cost.py` tests `test_get_io_cost_df`,
`test_plot_io_cost_ylims`, and `test_plot_io_cost_y_ticks_and_labels`

* Replace `by-average` x-axis label with module names

* Fixes #574

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
@@ -52,7 +52,7 @@ def get_by_avg_series(df: Any, mod_key: str, nprocs: int) -> Any:
     return by_avg_series
 
 
-def get_io_cost_df(report: darshan.DarshanReport, mod_key: str) -> Any:
+def get_io_cost_df(report: darshan.DarshanReport) -> Any:
     """
     Generates the I/O cost dataframe which contains the
     raw data to plot the I/O cost stacked bar graph.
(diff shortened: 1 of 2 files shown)


nawtrey:
This addresses suggestions made by @shanedsnyder and @carns for the I/O cost figure. I've updated the code to generate a single figure for all 3 modules, if available. I think this takes priority over gh-571 since the call signature and testing structure will have to be changed for that PR. 

Here are some examples using the updated plotting function: [example_figures.tar.gz](https://github.com/darshan-hpc/darshan/files/7672375/example_figures.tar.gz)

tylerjereddy:
Hmm, probably helpful to actually encode their comments into actual issue(s) rather than me digging through comments in other PRs, etc--I only see the removal of a label in the matching issue so I had to re-read what Shane wrote in the other PR related to this analysis.

We would only need something like a 1 line change to close the issue referenced from this PR for example.

darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
```diff
@@ -133,6 +128,8 @@ def plot_io_cost(report: darshan.DarshanReport, mod_key: str) -> Any:
     ax_norm.set_ylim(0, 100)
     # convert the normalized axis y labels to percentages
     ax_norm.yaxis.set_major_formatter(mtick.PercentFormatter())
+    if runtime > 1e4:
+        ax_raw.yaxis.set_major_formatter(mtick.StrMethodFormatter('{x:.0e}'))
```
tylerjereddy, line 132: I think this change to scentific notation probably makes sense, though I'd generally suggest separating fixes into smaller bits when possible (and I don't think I saw the scientific notation explicitly requested yet, so probably could have been delayed for a smaller diff, etc.).

tylerjereddy (commented):
I don't see any major issues in the code. I'm not sure I can identify the exact target behavior desired from the issues and comments I've seen though. @shanedsnyder  or @carns can you confirm that you like the transitions below (if so, can probably merge):

# `imbalanced-io.darshan` before:
image
image
image

# `imbalanced-io.darshan` after:
test

# `noposix.darshan` before:

test


# `noposix.darshan` after:

test

darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
tylerjereddy: Actually, I don't like the scentific notation yet--look at the plots above--the rounding of the decimal places leads to deceptive apparent spacing AND identical values plotted for neighboring ticks, which to me just doesn't make sense.

darshan-util/pydarshan/tests/test_plot_io_cost.py
```diff
@@ -12,140 +12,109 @@
...
-        mods
-    ):
+def test_plot_io_cost_y_ticks_and_labels(report, expected_yticks):
     # check the y-axis tick marks are at the appropriate
     # locations and the labels are as expected
 
     # create the expected y-axis tick labels from the y ticks
-    expected_yticklabels = [str(i) for i in expected_yticks]
+    if np.max(expected_yticks) > 1e4:
+        expected_yticklabels = [f"{i:.0e}" for i in expected_yticks]
```
tylerjereddy, line 96: this leads to `2.0e+4` at both 40 and 60 % ticks, which doesn't make sense. Honestly, I'd just yank out the scientific notation stuff for now so we can focus on the initial comments from Shane and Phil (i.e., use an issue for that).

nawtrey: Yeah, I'll go ahead and take it out, it looks like I didn't take a close enough look at it. Probably better to address on an individual level, like you mentioned above.

darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
nawtrey: Yeah, I'll have to take a stab at it again in a different PR I think. For now I've stripped the scientific notation changes out of this PR.

darshan-util/pydarshan/tests/test_plot_io_cost.py
nawtrey: Okay, this has now been removed from the PR.

nawtrey:
With the scientific notation changes removed, I've updated the PR description and updated the examples.

tylerjereddy:
Ok,

- it seems to do part of what Shane was asking for in a comment in gh-571, so I'll make the decision that we should move forward and they are busy for now
- the CI is passing and patch diff coverage is 100 %
- I re-read the code diff and it seems sensible enough
- I re-ran the pertinent plotting with this feature branch with a bunch of different log files locally and the results seem "ok" I think

tylerjereddy:
merged, thanks

PR #582, opened 2021-12-08
Deprecate DarshanReport(dtype=...)

jakobluettgau:
As discussed in #540 this pull request remove the dtype-Feature of DarshanReports in favor of the lazy to_df/to_dict/... conversion mechanism.

- [ ] Purge dtype handling code 
- [ ] Update tests

### DIFF
--- a/darshan-util/pydarshan/darshan/report.py
+++ b/darshan-util/pydarshan/darshan/report.py
@@ -295,14 +295,13 @@ class DarshanReport(object):
 
 
     def __init__(self, 
-            filename=None, dtype='numpy', 
+            filename=None, 
             start_time=None, end_time=None,
             automatic_summary=False,
             read_all=True, lookup_name_records=True):
         """
         Args:
             filename (str): filename to open (optional)
-            dtype (str): default dtype for internal structures
             automatic_summary (bool): automatically generate summary after loading
             read_all (bool): whether to read all records for log
             lookup_name_records (bool): lookup and update name_records as records are loaded
@@ -314,7 +313,6 @@ class DarshanReport(object):
         self.filename = filename
 
         # Behavioral Options
-        self.dtype = dtype                                  # default dtype to return when viewing records
         self.automatic_summary = automatic_summary
         self.lookup_name_records = lookup_name_records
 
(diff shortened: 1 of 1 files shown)


darshan-util/pydarshan/darshan/report.py
```diff
@@ -844,45 +791,21 @@ def mod_read_all_lustre_records(self, mod="LUSTRE", dtype=None, warnings=True):
...
-                'counters': combined_c,
-                }]
-
-
-
 
 
-    def mod_records(self, mod, 
-                    dtype='numpy', warnings=True):
+    def mod_records(self, mod, dtype='numpy', warnings=True):
```
nawtrey, line 808: Do we have a reason to keep the `dtype` parameter here? I can't find an instance where this method is actually used, but it could make sense to remove this one as well.

nawtrey (commented):
I looked through `report.py` a bit and I only have a couple comments/suggestions. 

The failing test, `tests/test_cffi_misc.py::test_file_hash_type` can be "fixed" by just removing the last half of it:
```diff
diff --git a/darshan-util/pydarshan/tests/test_cffi_misc.py b/darshan-util/pydarshan/tests/test_cffi_misc.py
index a6b4798..f1b2c8a 100644
--- a/darshan-util/pydarshan/tests/test_cffi_misc.py
+++ b/darshan-util/pydarshan/tests/test_cffi_misc.py
@@ -53,17 +53,6 @@ def test_file_hash_type(log_path):
     assert rec["counters"]["id"].dtype == np.uint64
     assert rec["fcounters"]["id"].dtype == np.uint64
 
-    # additionally check that the dataframes
-    # generated are of the correct types
-    report = darshan.DarshanReport(log_path, read_all=True)
-    report.mod_read_all_records("POSIX", dtype="pandas")
-    rec_counters = report.records["POSIX"][0]["counters"]
-    rec_fcounters = report.records["POSIX"][0]["fcounters"]
-    # verify the records returned have the correct
-    # data type for the ids/hashes
-    assert rec_counters["id"].dtype == np.uint64
-    assert rec_fcounters["id"].dtype == np.uint64
-
 
 @pytest.mark.parametrize("dtype", ["numpy", "dict", "pandas"])
 def test_log_get_generic_record(dtype):
```

I'm not generally a fan of removing testing code, but with the accompanying changes in this PR I don't think we can effectively probe the dataframes targeted by this test. The alternative might be to change the test to use the `to_df()` method, but I don't think that would really be relevant since we wouldn't be executing the `dtype="pandas"` code.

Issue #577, opened 2021-12-03
Adjustments to `plot_access_histogram()` for Darshan Summary Report


nawtrey:
`plot_access_histogram` needs some final adjustments before it can be added to the Darshan Summary Report. I'll show some specific examples that I think are representative of some of the issues I'm seeing. 

## `noposixopens.darshan`:
![noposixopens darshan_POSIX](https://user-images.githubusercontent.com/46797896/144533954-3dd875eb-303a-4720-b623-e5ba7e0f6ed8.png)

The title has a misspelled word, but I think we will remove it like we have for other figures and just add an x-axis label instead. Based on prior conversations about other figures, I think we want to display values even if they are zero, but maybe setting the minimum y values will make this look a little cleaner.

## `theta-large-2021-bz2.darshan`:
![theta-large-2021-bz2 darshan_POSIX](https://user-images.githubusercontent.com/46797896/144533925-17cf67a0-ece8-414c-86a8-180f7e436d3f.png)

This one has some label overlap, and the highest bar label is slightly touching the top of the figure border. I'm not sure the best way to handle this. The best looking solution I've seen would be to make this a horizontal bar graph, and find a way to set the x-axis limits such that the labels never go outside of the figure border. Or maybe just removing the bar labels entirely would be fine, I'm not sure. It depends on what the team wants to convey with these plots. 

In summary:
 - [x] remove the title and add "access size" label to x-axis
 - [ ] set the y minimum so we never see negative tick values
 - [ ] set the y max so the labels are never touching the frame
 - [ ] switch to horizontal bar graphs

I think these cover the primary issues I've seen with the figures. I've listed some other changes that are more code-focused below:
- [x] `autolabel` function should be moved outside of plotting function to allow for testing
- [ ] Add type hints
- [ ] Update documentation

nawtrey:
After feedback from the team it seems we want to switch to horizontal bars. I've updated the checklist to reflect this.

shanedsnyder:
Tying up some loose ends on old issues, and I think the remaining items listed here are either handled in #907 (ylim of 0, removal of right/top axis spines to avoid annotation overlap) or are things we aren't going to do (horizontal orientation of the bars).

PR #584, opened 2021-12-08
MAINT: Cleanup `plot_access_histogram.py`

nawtrey:
* Remove unnecessary imports

* Move title into x-axis label

* Reformat longer lines to reduce length

* Simplify data handling code

* Contributes to #577

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_access_histogram.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_access_histogram.py
@@ -1,8 +1,5 @@
 # -*- coding: utf-8 -*-
 
-from darshan.report import *
-
-import matplotlib
 import matplotlib.pyplot as plt
 import numpy as np
 
@@ -26,15 +23,14 @@ def plot_access_histogram(report, mod, ax=None):
     if 'mod_agg_iohist' in dir(report):
         report.mod_agg_iohist(mod)
     else:
-        print("Can not create summary, mod_agg_iohist aggregator is not registered with the report class.")
+        print("Cannot create summary, mod_agg_iohist aggregator is not registered with the report class.")
 
     # defaults
     labels = ['0-100', '101-1K', '1K-10K', '10K-100K', '100K-1M', '1M-4M', '4M-10M', '10M-100M', '100M-1G', '1G+']
-    read_vals = [0, 0, 0, 0, 0,  0, 0, 0, 0, 0]
-    write_vals = [0, 0, 0, 0, 0,  0, 0, 0, 0, 0]
 
     agg = report.summary['agg_iohist'][mod]
-
+    # TODO: can simplify the read/write vals below after
+    # support for python 3.6 is dropped
     read_vals = [
         agg['READ_0_100'],
         agg['READ_100_1K'],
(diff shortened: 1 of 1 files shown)


darshan-util/pydarshan/darshan/experimental/plots/plot_access_histogram.py
```diff
@@ -26,41 +23,20 @@ def plot_access_histogram(report, mod, ax=None):
...
-        agg['WRITE_100M_1G'],
-        agg['WRITE_1G_PLUS']
+    labels = [
+        '0-100', '101-1K', '1K-10K', '10K-100K', '100K-1M',
+        '1M-4M', '4M-10M', '10M-100M', '100M-1G', '1G+',
     ]
 
+    agg_vals = list(report.summary['agg_iohist'][mod].values())
+    read_vals = agg_vals[:10]
+    write_vals = agg_vals[10:]
```
tylerjereddy, line 39: It looks like you're depending on the insertion order of the objects in the dictionary being preserved, but we need to support Python 3.6 for the moment, for which this is not considered good practice because it is an implementation detail not guaranteed until Python 3.7.

So, I would just add a small comment or TODO to the code and leave this alone for now until we're at higher minimum Python version. You could still delete the pointless double assignment to `read_vals` and `write_vals`.

darshan-util/pydarshan/darshan/experimental/plots/plot_access_histogram.py
```diff
@@ -26,41 +23,20 @@ def plot_access_histogram(report, mod, ax=None):
     if 'mod_agg_iohist' in dir(report):
         report.mod_agg_iohist(mod)
     else:
-        print("Can not create summary, mod_agg_iohist aggregator is not registered with the report class.")
+        print(
+            "Cannot create summary, mod_agg_iohist aggregator"
+            "is not registered with the report class."
+        )
```
tylerjereddy, line 29: I wouldn't spend much energy changing line lengths unless we start enforcing that in a linter/CI (probably not a good use of time for now). Even if the editor does it automatically I still have to parse it myself when reading a diff.

Also, as the CI shows, this line isn't covered by tests, and probably should be something other than a plain `print()`.

nawtrey: Okay, I've reverted the line length changes. 

Yeah, it should probably raise an error. I think if this is triggered it ends up raising an error anyways, so might as well. The same thing is present in `plot_opcounts.py`, so I'll open an issue for both cases.

darshan-util/pydarshan/darshan/experimental/plots/plot_access_histogram.py
nawtrey: Okay, I wasn't aware of that difference so I've made the recommended changes here.

darshan-util/pydarshan/darshan/experimental/plots/plot_access_histogram.py
nawtrey: > Also, as the CI shows, this line isn't covered by tests, and probably should be something other than a plain print().

I've opened issue #588 to address this.

tylerjereddy:
merged, thanks

PR #587, opened 2021-12-09
ENH: Add log path retrieval machinery

### LINKED ISSUE #566: ENH: on designing log file retrieval machinery
# Background

Something that I've been trying to encourage is a system for gracefully retrieving log files, wherever they might reside, in a sustainable/maintainable way.

There are a few problems with the current approaches:

- for the test suite in the main repo, there is a tendency to specify relative paths like the following for log files that are stored in the main repo: `examples/example-logs/dxt.darshan`; this unnecessarily ties the test suite to execution from a certain relative path, and requires developers to use a different approach to retrieve log files from the main repo vs. the logs repo
- at the moment, all current and proposed usage of log files in the main repo test suite requires some level of manual intervention; for example, manually marking those tests cases with a skip like in gh-553 (see below); is it really a useful exercise for developers to manually mark every single log repo file with a `skipif` directive and type out the same message about the logs repo missing over and over?

```python
        pytest.param(
            "imbalanced-io.darshan",
            "POSIX",
            plot_access_histogram,
            [18, 2492, 14679, 0, 50486, 186, 0, 0, 0,
            0, 43, 301, 2, 0, 50486, 0, 0, 0, 0, 0],
            marks=pytest.mark.skipif(not pytest.has_log_repo, # type: ignore
                                    reason="missing darshan_logs"),
```
- the current infrastructure depends on `pytest` fixtures, but the need to robustly handle retrieval of log files from different paths and two projects is shared by other infrastructure: the `asv` benchmarks won't be able to leverage a solution in a `pytest` fixture because it doesn't use `pytest` to run its suite; likewise, if you're working locally with small debug script, you probably also won't gain much convenience from the availability of a solution embedded in a `pytest` fixture
- the logs repo is using yet another approach, which currently requires separate maintenance, to retrieve log files from dictionaries defined the main project, but there is more than one dictionary, and a bunch of conditional logic in the setup process for those benchmarks to select the correct code path

# Proposed Solution

I think we want an approach that reduces duplication/maintenance burden and gracefully handles the presence/absence of the logs repo, and that can be used for both the test suite, the `asv` benchmark suite, and even local debugging scripts.

Here is one approach I was hoping to inspire a bit:

- a utility function that resides in `pydarshan`; can be imported by the test suite, the `asv` suite, and any other external scripts that `import darshan`
- the function accepts a single filename string for a log file and has logic that does the following:
  - search through all paths in the installed `darshan` project that can contain log file assets, and return the absolute path to the installed log file that matches (users no longer worry about something being in `examples` or in another folder, they just provide the log file name)
  - if nothing is found in the main repo for the log filename provided, make an attempt to import the darshan logs project
  - if the import of the darshan logs project fails, and it is detected that you are running `pytest`, use an imperative `pytest.skip(reason)` so that skipping logic, and the skipping message, are maintained in a single location and do not require developers to manually specify the skips/message [if you are not running `pytest`, instaed of using an imperative `pytest.skip`, you probably simply want to raise an exception of the appropriate type, like in the `asv` suite or your own external script trying to find a log file by name]
  - if the import of the darshan logs project succeeds, then return the absolute path to the log filename of interest
  - if there is still no match on the log filename, error out

# Some Considerations

- It seems likely that for the foreseeable future, we'll have logs spread between the main repo and the logs repo, so I don't see the problem disappearing immediately
- This doesn't really require any massive changes--all of the old/current approaches can remain in effect and be progressively removed at whatever pace is appropriate
- for the cases of duplicate log files, we can start to reduce the level of disruption because this function will search until it finds a particular log file, in a manner transparent to the developer (if you delete a duplicate log file, it will simply return the absolute path to the original log file)
- since the logic for file retrieval would be located in one place, modernizing the approach to use i.e., features of `importlib.resources` or whatever else comes along may be a fair bit less work

nawtrey:
### Description

* Add module `log_utils.py` containing new function
`get_log_path` for easy retrieval of logs from either local
sources or darshan-logs repo, using only the filename

* Fixes #566

### TODO
 - [ ] Decide on module location
 - [ ] Finish documentation for private functions 
 - [ ] Add tests


nawtrey:
I've opened this as a draft in the event that reviewers have strong feelings on any of the tasks in "TODO".

I have tested this for all of the various cases, and it seems to be working as intended. For example, changing some of the local tests and adding the appropriate import (`from darshan.logutils import get_log_path`), the following test still passes even with the addition of a test case from the darshan-logs repo:
```diff
diff --git a/darshan-util/pydarshan/tests/test_summary.py b/darshan-util/pydarshan/tests/test_summary.py
index 1860c25..30b2ddf 100644
--- a/darshan-util/pydarshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/tests/test_summary.py
@@ -70,11 +70,12 @@ def test_main_with_args(tmpdir, argv):

 @pytest.mark.parametrize(
     "argv", [
-        [os.path.abspath("./tests/input/noposix.darshan")],
-        [os.path.abspath("./tests/input/noposix.darshan"), "--output=test.html"],
-        [os.path.abspath("./tests/input/sample-dxt-simple.darshan")],
-        [os.path.abspath("./tests/input/sample-dxt-simple.darshan"), "--output=test.html"],
+        [get_log_path("noposix.darshan")],
+        [get_log_path("noposix.darshan"), "--output=test.html"],
+        [get_log_path("sample-dxt-simple.darshan")],
+        [get_log_path("sample-dxt-simple.darshan"), "--output=test.html"],
         [None],
+        [get_log_path("mpi-io-test.darshan")]
     ]
 )
```
 I also checked that this gracefully skips the test if the `darshan-logs` repo is not present (and it does). 

Additionally, I tested this on the following `asv` benchmark, and this seemed to work fine as well:
```diff
diff --git a/darshan-util/pydarshan/benchmarks/benchmarks/report.py b/darshan-util/pydarshan/benchmarks/benchmarks/report.py
index d48d3e0..a1b5665 100644
--- a/darshan-util/pydarshan/benchmarks/benchmarks/report.py
+++ b/darshan-util/pydarshan/benchmarks/benchmarks/report.py
@@ -4,6 +4,7 @@ import os
example_logs = importlib.import_module("examples.example-logs")
from tests.input import test_data_files_dxt, test_data_files
import darshan
+from darshan.log_utils import get_log_path

class ModReadAllRecords:
@@ -23,11 +24,7 @@ class ModReadAllRecords:

    def setup(self, dtype, darshan_logfile, mod, read_all):
        filename = os.path.basename(darshan_logfile)
-        if "examples" in darshan_logfile:
-            self.logfile = example_logs.example_data_files_dxt[filename]
-        else:
-            self.logfile = test_data_files_dxt[filename]
-
+        self.logfile = get_log_path(filename)
        self.report = darshan.DarshanReport(self.logfile, read_all=read_all)
```

nawtrey:
There seems to be a `mypy` issue with python 3.8:

```
darshan/log_utils.py:44: error: "Traversable" has no attribute "glob"  [attr-defined]
Found 1 error in 1 file (checked 39 source files)
```

I'll see if I can push a fix in for this.

tylerjereddy:
> I have tested this for all of the various cases,

How did you test the case where `asv` leverages the logs repo?

nawtrey:
> > I have tested this for all of the various cases,
> 
> How did you test the case where `asv` leverages the logs repo?

Oops, yeah that statement is incorrect. `asv` fails if you add a log from the darshan-logs repo because it fails to `import darshan_logs`:
```
File "/Users/nawtrey/darshan_install/darshan/darshan-util/pydarshan/benchmarks/env/6df5fb12e06d3834d1be558e6a895c97/lib/python3.9/site-packages/darshan/log_utils.py", 
line 94, in get_log_path raise FileNotFoundError(err_msg)
FileNotFoundError: File imbalanced-io.darshan not found locally and darshan-logs repo unavailable.
```

darshan-util/pydarshan/darshan/log_utils.py
```diff
@@ -0,0 +1,94 @@
...
+try:
+    import darshan_logs
+    has_log_repo = True
+except ImportError:
+    has_log_repo = False
+
+
+def _locate_local_log(filename: str) -> Union[str, None]:
+    """Locates a log if available in `tests/input` or `examples/example-logs`."""
+    pydarshan_path = pathlib.Path(__file__).parent.parent
```
tylerjereddy, line 30: It is best to avoid using `__file__` if possible (i.e., for reasons described in https://stackoverflow.com/a/58941536/2942522; I think I cheated and used it in a few places already, but it is sub-optimal). Plus the logs repo approach below is already using the desirable `importlib.resources` so matching that would make sense.

If that could be done, it would also enable more code reuse (see below). There is a big problem though--we can't actually use the data files from the `darshan` project install proper, because those files are never installed (!), which is one point I don't see discussed here or in the matching issue, so we wouldn't actually be unchaining ourselves from the need to run the testsuite from within a specific path in the local clone of the project.

It may be worth discussing with the team whether it would be acceptable to place the small number of files to be retained in the main repo in a standard installation for example. The drawback is likely size, which is probably why Jakob didn't include them by default.

```diff
--- a/darshan-util/pydarshan/darshan/log_utils.py
+++ b/darshan-util/pydarshan/darshan/log_utils.py
@@ -25,22 +25,9 @@ except ImportError:
     has_log_repo = False
 
 
-def _locate_local_log(filename: str) -> Union[str, None]:
-    """Locates a log if available in `tests/input` or `examples/example-logs`."""
-    pydarshan_path = pathlib.Path(__file__).parent.parent
-    local_dirs = ["tests/input", "examples/example-logs"]
-    for ldir in local_dirs:
-        ldir = os.path.join(pydarshan_path, ldir)
-        for log_path in glob.glob(os.path.join(ldir, "*.darshan")):
-            if filename in log_path:
-                return log_path
-    # if log is not found
-    return None
-
-
-def _locate_repo_log(filename: str) -> Union[str, None]:
-    """Locates a log in the darshan_logs repo."""
-    p = importlib_resources.files('darshan_logs') # type: Any
+def _locate_log(filename: str, project: str) -> Union[str, None]:
+    """Locates a log file in a project."""
+    p = importlib_resources.files(project) # type: Any
     darshan_logs_paths = [str(p) for p in p.glob('**/*.darshan')]
     for log_path in darshan_logs_paths:
         if filename in log_path:
@@ -74,12 +61,12 @@ def get_log_path(filename: str) -> str:
 
     """
     # try local paths
-    log_path = _locate_local_log(filename=filename)
+    log_path = _locate_log(filename=filename, project='darshan')
     if log_path:
         return log_path
     # try the logs repo
     if has_log_repo:
-        log_path = _locate_repo_log(filename=filename)
+        log_path = _locate_log(filename=filename, project='darshan_logs')
         if log_path:
             return log_path
         else:

```

tylerjereddy: We could also cheat a bit by modifying the above to combine both approaches (always using `importlib` and also moving up a few dirs so that we can scan for the adjacent data dirs), which doesn't avoid the need for a local clone of `darshan` but does reduce duplication:

```diff
-def _locate_repo_log(filename: str) -> Union[str, None]:
-    """Locates a log in the darshan_logs repo."""
-    p = importlib_resources.files('darshan_logs') # type: Any
+def _locate_log(filename: str, project: str) -> Union[str, None]:
+    """Locates a log file in a project."""
+    p = importlib_resources.files(project).parent.parent # type: Any
     darshan_logs_paths = [str(p) for p in p.glob('**/*.darshan')]
```

That seems to "just work", but it still makes me cringe a bit---moving up a few dirs relative to a given target package could go searching for other paths/logs we don't want to be picking up.

tylerjereddy (commented):
I did a first pass at reviewing/thinking deeply about some of the issues. It looks pretty close to what I had in mind except for the remaining issue about not being able to unchain us from running the tests right next to a local install.

tylerjereddy:
Come to think of it, since the `tests` folder is not installed, it really is currently hopeless to test an *installed* version of `darshan` without a local clone with i.e., `pytest --pyargs darshan` because the tests are not packaged in the install (even if we ignore the data files not being packaged) so one is forced to run the suite from the cloned repo path.

That may also be worth discussing a bit with the team, and is probably less controversial than packaging the data files. Then if a user is having issues with their `darshan` HPC cluster spack install, though could type i.e., `pytest --pyargs darshan` to make sure it is ok.

darshan-util/pydarshan/darshan/log_utils.py
tylerjereddy: (for now, could only move up the dirs if `project` == `darshan`, which would just be a more concise version of the original logic)

tylerjereddy:
I should also mention that the unguarded `import pytest` here is problematic--we don't want a mandatory `pytest` dependency for `darshan` at runtime, only at test time.

darshan-util/pydarshan/darshan/log_utils.py
```diff
@@ -0,0 +1,94 @@
...
+from typing import Union, Any
+
+try:
+    import darshan_logs
+    has_log_repo = True
+except ImportError:
+    has_log_repo = False
+
+
+def _locate_local_log(filename: str) -> Union[str, None]:
```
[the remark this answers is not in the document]
nawtrey: I've updated this to use `Optional[X]` over `Union`.

darshan-util/pydarshan/darshan/log_utils.py
nawtrey: Thanks for the recommendations, I went ahead and simplified the log locating functions down and removed the dependence on `pathlib`. 

I think the remaining concern here is moving up 2 directories, but I haven't figured out another way of doing this  that doesn't involve moving the files themselves. Based on our meeting yesterday it sounds like that is a possibility, so maybe that will have to be fixed before this is merged.

nawtrey:
> I should also mention that the unguarded `import pytest` here is problematic--we don't want a mandatory `pytest` dependency for `darshan` at runtime, only at test time.

I changed this so it only imports if it a current `pytest` session is running.

tylerjereddy:
Why do we need to move up two directories instead of just one? I can't reproduce that requirement locally, so a comment explaining that need would be helpful for me I think.

Testing on my Linux box, I can see that the current approach with `PYTEST_CURRENT_TEST` is not going to cut it, because that environment variable is not set during test collection time (when the pytest decorators get evaluated I think), only when actually running, and I see the following when trying to pull in a log file with this marchinery with the logs repo absent:

```
pytest tests/test_summary.py::test_main_without_args
```

```
================================================================================================================================= test session starts =================================================================================================================================
platform linux -- Python 3.8.9, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
rootdir: /home/tyler/github_projects/darshan/darshan-util/pydarshan, configfile: setup.cfg
plugins: cov-3.0.0, forked-1.3.0, xdist-2.4.0
collected 0 items / 1 error                                                                                                                                                                                                                                                           

======================================================================================================================================= ERRORS ========================================================================================================================================
_______________________________________________________________________________________________________________________ ERROR collecting tests/test_summary.py ________________________________________________________________________________________________________________________
tests/test_summary.py:78: in <module>
    ???
darshan/log_utils.py:91: in get_log_path
    raise FileNotFoundError(err_msg)
E   FileNotFoundError: File imbalanced-io.darshan not found locally and darshan-logs repo unavailable.
---------------------------------------------------------------------------------------------------------
```

```diff
--- a/darshan-util/pydarshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/tests/test_summary.py
@@ -9,6 +9,7 @@ from pandas.testing import assert_frame_equal #type: ignore
 
 import darshan
 from darshan.cli import summary
+from darshan.log_utils import get_log_path
 
 
 @pytest.mark.parametrize(
@@ -73,7 +74,8 @@ def test_main_with_args(tmpdir, argv):
         [os.path.abspath("./tests/input/noposix.darshan")],
         [os.path.abspath("./tests/input/noposix.darshan"), "--output=test.html"],
         [os.path.abspath("./tests/input/sample-dxt-simple.darshan")],
-        [os.path.abspath("./tests/input/sample-dxt-simple.darshan"), "--output=test.html"],
+        [get_log_path("sample-dxt-simple.darshan"), "--output=test.html"],
+        [get_log_path("imbalanced-io.darshan"), "--output=test.html"],
         [None],
     ]
 )

```

In fact, it seems that even if you do something a little more robust:

```diff
--- a/darshan-util/pydarshan/darshan/log_utils.py
+++ b/darshan-util/pydarshan/darshan/log_utils.py
@@ -2,6 +2,7 @@
 Module for log handling utilities.
 """
 
+import sys
 import platform
 python_version = platform.python_version_tuple()
 # shim for convenient Python 3.9 importlib.resources
@@ -16,7 +17,7 @@ import os
 import glob
 from typing import Optional, Any
 
-if "PYTEST_CURRENT_TEST" in os.environ:
+if "pytest" in sys.modules:
     # only import pytest if used in a testing context
     import pytest
     pytest_session = True
```

There is still a fundamental problem.

```
_______________________________________________________________________________________________________________________ ERROR collecting tests/test_summary.py ________________________________________________________________________________________________________________________
Using pytest.skip outside of a test is not allowed. To decorate a test function, use the @pytest.mark.skip or @pytest.mark.skipif decorators instead, and to skip a module use `pytestmark = pytest.mark.{skip,skipif}.

```

I'm a little confused because above it says the skipping happens gracefully for this very test?

tylerjereddy:
I would also actually add an example replacement in this PR that uses the logs repo (including a decorator case like above) so that I can see if the CI passes instead of trusting that the testing was actually done (the CI has jobs with/without the logs repo).

tylerjereddy:
> I also checked that this gracefully skips the test if the darshan-logs repo is not present (and it does).

I checked on my Mac too, still a failure to skip with `pytest` for the exact test you cite.

nawtrey:
> Why do we need to move up two directories instead of just one? 

Thanks for the further review. You're correct that we just need to move up a single directory, so I'll fix that. I think the diff above does show 2 `.parent`s but I just didn't vet it appropriately, so that's my mistake.

> I checked on my Mac too, still a failure to skip with `pytest` for the exact test you cite.

I was able to reproduce your results on my Mac with the identical error message you showed above. I also tried resetting my local branch back to my original commit which yielded no change, so at this point I'm not sure how I was able to get the tests to skip without `pytest` complaining. In the `pytest` documentation they recommend an [alternative solution](https://pytest.org/en/6.2.x/example/simple.html#detect-if-running-from-within-a-pytest-run), although they seem to recommend against it:
> Usually it is a bad idea to make application code behave differently if called from a test

I'll have to look into this in greater depth and see if I can produce the desired behavior. I remember it being surprising that I _didn't_ get the error (I had seen discussion about it on stackoverflow) but I do remember it skipping without error, and I ran through the various cases several times, initially. That's not to say I couldn't totally be wrong, but there might be a subtlety I have overlooked.

tylerjereddy:
The constraint of behaving properly both in a parametrization mark and in a test body is probably possible to satisfy with careful examination of pytest API docs, and possibly pytest source code, but I would anticipate some modest research is needed on how to "yield" or "inject" decorator skips from functions called during test *collection* instead of *execution* (and reliably detecting the difference between collection and execution).

My original design/issue described an imperative `pytest.skip`, which cannot be used in a parametrization mark, at least not without depending on a plugin or something. A solution that only has the constraint of skipping within the *body* of a test (parametrized on i.e., filename strings, with `get_log_path` in test body *only*) requires less research, and should be a largely trivial modification of the current branch (possibly no modification at all beyond some code cleanup, given that constraint), though is less versatile.

It isn't really application code--just a testing/benchmark shim not used in library code, so wouldn't worry too much about the caution.

tylerjereddy:
To nudge this forward a bit...

The changes below seem to work for the parametrized skip case via `pytest --import-mode=importlib tests/test_summary.py::test_main_without_args -rs -v `, but would still need substantial engineering and testing to work for broader incantations with i.e., `test/test_summary.py` and also a shim for the imperative in-test usage. One would have to study how `pytest` constructs the "nodes" of test cases more globally in `items`, and then modify `items` more robustly than below.

```diff
diff --git a/darshan-util/pydarshan/darshan/log_utils.py b/darshan-util/pydarshan/darshan/log_utils.py
index 507ca678..547761bf 100644
--- a/darshan-util/pydarshan/darshan/log_utils.py
+++ b/darshan-util/pydarshan/darshan/log_utils.py
@@ -3,6 +3,7 @@ Module for log handling utilities.
 """
 
 import platform
+import sys
 python_version = platform.python_version_tuple()
 # shim for convenient Python 3.9 importlib.resources
 # interface
@@ -16,10 +17,13 @@ import os
 import glob
 from typing import Optional, Any
 
-if "PYTEST_CURRENT_TEST" in os.environ:
+if 'pytest' in sys.modules:
     # only import pytest if used in a testing context
     import pytest
     pytest_session = True
+    class Skipper:
+        def __init__(self, err_msg):
+            self.err_msg = err_msg
 else:
     pytest_session = False
 
@@ -83,6 +87,6 @@ def get_log_path(filename: str) -> str:
     else:
         err_msg = f"File {filename} not found locally and darshan-logs repo unavailable."
         if pytest_session:
-            pytest.skip(err_msg)
+            return Skipper(err_msg)
         else:
             raise FileNotFoundError(err_msg)
diff --git a/darshan-util/pydarshan/tests/conftest.py b/darshan-util/pydarshan/tests/conftest.py
index 6ae8875f..e673a29a 100644
--- a/darshan-util/pydarshan/tests/conftest.py
+++ b/darshan-util/pydarshan/tests/conftest.py
@@ -1,3 +1,4 @@
+import darshan
 import platform
 python_version = platform.python_version_tuple()
 
@@ -21,6 +22,43 @@ except ImportError:
 def pytest_configure():
     pytest.has_log_repo = has_log_repo
 
+class CustomNode:
+    def __init__(self, mimic_item):
+        for attr in dir(mimic_item):
+            if attr not in ["own_markers", "iter_markers"] and "__" not in attr:
+                setattr(self, attr, getattr(mimic_item, attr))
+
+def pytest_collection_modifyitems(config, items):
+    new_items = []
+    skip_indices = []
+    for i, test_item in enumerate(items.copy()):
+        param_marker = test_item.own_markers[0]
+        num_args = int(len(param_marker.args) / 2)
+        for arg_num in range(num_args):
+            arg_name = param_marker.args[arg_num]
+            arg_values = param_marker.args[arg_num + 1]
+            for arg_val_num, arg_value in enumerate(arg_values[:]):
+                skipper_found = 0
+                for arg_sub_value in arg_value:
+                    if isinstance(arg_sub_value, darshan.log_utils.Skipper):
+                        err_msg = arg_sub_value.err_msg
+                        if arg_val_num not in skip_indices:
+                            skip_indices.append(arg_val_num)
+                            customnode = CustomNode(test_item)
+                            customnode.own_markers = pytest.mark.skip
+                            def iter_marks(name):
+                                yield pytest.mark.skip(reason=err_msg)
+                            customnode.iter_markers = iter_marks
+                            new_items.append(customnode)
+
+    for i, item in enumerate(items.copy()):
+        if i not in skip_indices:
+            new_items.append(item)
+
+    items.clear()
+    items.extend(new_items)
+
+
 @pytest.fixture
 def log_repo_files():
     # provide a convenient way to access the list
diff --git a/darshan-util/pydarshan/tests/test_summary.py b/darshan-util/pydarshan/tests/test_summary.py
index 1860c255..1fe1b3bf 100644
--- a/darshan-util/pydarshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/tests/test_summary.py
@@ -9,6 +9,7 @@ from pandas.testing import assert_frame_equal #type: ignore
 
 import darshan
 from darshan.cli import summary
+from darshan.log_utils import get_log_path
 
 
 @pytest.mark.parametrize(
@@ -74,6 +75,7 @@ def test_main_with_args(tmpdir, argv):
         [os.path.abspath("./tests/input/noposix.darshan"), "--output=test.html"],
         [os.path.abspath("./tests/input/sample-dxt-simple.darshan")],
         [os.path.abspath("./tests/input/sample-dxt-simple.darshan"), "--output=test.html"],
+        [get_log_path("imbalanced-io.darshan"), "--output=test.html"],
         [None],
     ]
 )

```

You'll also notice that the output show two `argv0` entries as result of some of the hacking, so even the isolated case is just a rough draft:
image

tylerjereddy:
I've also opened an upstream issue cross-linked above to see what the `pytest` folks think about the situation--when digging deeper on the machinery of an open source project, it doesn't hurt to ping the development team as long as you have evidence of having put in a resonable effort to solve the problem.

tylerjereddy:
If I don't see any further progress here soon, I'll probably proceed with refining what is here with my original suggestion above:

> A solution that only has the constraint of skipping within the body of a test (parametrized on i.e., filename strings, with get_log_path in test body only) requires less research, and should be a largely trivial modification of the current branch (possibly no modification at all beyond some code cleanup, given that constraint), though is less versatile.

nawtrey:
Oops, I thought I had commented here..

I've been prioritizing this lower than report-related enhancements/bugs, and I was still unsure of what to do after reading through the pytest devs responses, so I hadn't taken any action yet. I'm happy to review/test when you think this is good to go

tylerjereddy:
The example usage of `get_log_path()` to retrieve a logs repo file added here appears to be behaving normally, along with some expanded docs not to use it in a `pytest` decorator, etc:


- this PR:
  - Python 3.8 Linux (logs repo absent): 243 passed, 6 skipped, 2 warnings in 136.13s (0:02:16)
  - Python 3.10 Linux:  249 passed, 3 warnings in 282.85s (0:04:42)
- last merged PR: https://github.com/darshan-hpc/darshan/pull/602
  - Python 3.8 Linux (logs repo absent): 243 passed, 5 skipped, 2 warnings in 95.35s (0:01:35)
  - Python 3.10 Linux: 248 passed, 3 warnings in 221.19s (0:03:41)

Of course, some of the new code won't be covered by a `pytest` test because it is a `pytest` aware utility. We could add a debug env variable that allows us to force deactivation of `pytest` handling to test the non-`pytest` situations, but for now I've checked locally and it seems like the error messages are friendly enough if you do weird stuff.

Actually, it looks to me like the codecov report isn't quite right--I think the accumulation of coverage across different jobs may be not working properly, perhaps because codecov is reporting before all CI entries finish and upload their results (the Python 3.8/non-logs-repo CI runs are typically much faster of course).

Issue #585, opened 2021-12-08
Statically building with darshan fails at linker phase


bebosudo:
Hi,
I built darshan on Perlmutter at NERSC and it works fine for dynamically linked applications
```
$ module load darshan/3.3.1
$ cat hello.c
int main() {}
$ cc hello.c; echo $?
0
$ ldd a.out |grep darshan
	libdarshan.so => /global/common/software/nersc/pm-2021q4/sw/darshan/3.3.1/lib/libdarshan.so
```

But upon trying to build a dummy executable statically with darshan enabled, the link cannot find the fmpich, mpichcxx and lz libraries:
```
$ module load cray-pmi{,-lib}  # only required for static builds
$ cc -static hello.c
/usr/bin/ld: cannot find -lfmpich
/usr/bin/ld: cannot find -lmpichcxx
/usr/bin/ld: cannot find -lz
/usr/bin/ld: cannot find -lz
collect2: error: ld returned 1 exit status
```

Unloading darshan makes it compile fine.
Any idea what it could be? We don't officially support static builds, but it would be a nice thing to support

carns:
Thanks for the bug report @bebosudo .  I just activated my perlmutter account and confirmed the same thing.  It also happens if you enable static compilation with `export CRAYPE_LINK_TYPE=static`, which is something I was curious about.  I'll have a look and see if I can spot the problem.  It seems like either some library paths have been truncated or else they are in the wrong order for the Darshan library.

carns:
I don't have a solution yet, but a few interesting pieces of information:

1) I didn't look closely enough to see if the errors match.  For me, with the default perlmutter environment, I also get this when trying to statically compile, regardless of whether Darshan is loaded or not:
```
/usr/bin/ld: cannot find -lpmi
/usr/bin/ld: cannot find -lpmi2
```

2) The fmpich and mpichcxx errors are sensible enough from the linker's perspective, those libraries don't exist (though there is a dynamic version of the former):
```
$ ls /opt/cray/pe/mpich/8.1.11/ofi/nvidia/20.7/lib
libfmpich.so	     libmpifort_nvidia.so	  libmpi.so
libmpi.a	     libmpifort_nvidia.so.12	  libmpl.so
libmpich.a	     libmpifort_nvidia.so.12.0.0  libopa.so
libmpichf90.a	     libmpifort.so		  libtvmpich.a
libmpichf90.so	     libmpi_nvidia.a		  libtvmpich.so
[...]

PR #589, opened 2021-12-10
remove unecessary MPICH libraries from static link

carns:
Partially addresses #585 

This PR removes -lfmpich and -lmpicxx from the linker command  when statically compiling with Darshan support.  These libraries were previously important when Darshan used PMPI for instrumentation in C++ and Fortran codes, but that is no longer the case, and the library names may not be present in some MPICH builds.


roblatham00 (approved):
I like simplifying things.  great!

PR #592, opened 2021-12-15
BUG: Correct I/O cost figure y limits

### LINKED ISSUE #591: BUG: I/O cost figure sets incorrect y limits
All I/O cost figures should have a maximum y value of the stored `runtime`, but many of the available logs do not have this set correctly. At the moment it appears this is being set based on the `matplotlib` prescribed value, which is incorrect. 

For example, the following examples have noticeably different `runtime` values:
```
========================================
FILE: noposixopens.darshan
runtime:  1110
Y max:  1200.0
========================================
FILE: sample-badost.darshan
runtime:  779
Y max:  800.0
========================================
FILE: dxt.darshan
runtime:  1468
Y max:  1600.0
========================================
FILE: noposix.darshan
runtime:  39212
Y max:  40000.0
```

This should be simple to fix (i.e. `yticks = np.linspace(0, runtime, n_ticks)`), and should be accompanied by a regression test that at least covers the above cases.

nawtrey:
* Update y-axis ticks for I/O cost figure to have a maximum
value of the runtime calculated from the log metadata

* Correct and add new cases for test `test_plot_io_cost_ylims`
to regression guard against future y limit changes

* Correct test `test_plot_io_cost_y_ticks_and_labels`

* Fixes #591

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
@@ -130,9 +130,8 @@ def plot_io_cost(report: darshan.DarshanReport) -> Any:
     ax_norm.yaxis.set_major_formatter(mtick.PercentFormatter())
     # align both axes tick labels so the grid matches
     # values on both sides of the bar graph
-    norm_yticks = ax_raw.get_yticks()
     n_ticks = len(ax_norm.get_yticks())
-    yticks = np.linspace(norm_yticks[0], norm_yticks[-1], n_ticks)
+    yticks = np.linspace(0, runtime, n_ticks)
     ax_raw.set_yticks(yticks)
     # add the legend and appropriate labels
     ax_raw.set_ylabel("Runtime (s)")
--- a/darshan-util/pydarshan/tests/test_plot_io_cost.py
+++ b/darshan-util/pydarshan/tests/test_plot_io_cost.py
@@ -53,7 +53,19 @@ def test_get_io_cost_df(report, expected_df):
         ),
         (
             darshan.DarshanReport("examples/example-logs/sample-badost.darshan"),
-            [0.0, 800.0],
+            [0.0, 779.0]
+        ),
+        (
+            darshan.DarshanReport("examples/example-logs/dxt.darshan"),
+            [0.0, 1468.0],
+        ),
+        (
+            darshan.DarshanReport("examples/example-logs/noposix.darshan"),
+            [0.0, 39212.0],
+        ),
+        (
+            darshan.DarshanReport("tests/input/noposixopens.darshan"),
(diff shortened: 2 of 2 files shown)


tylerjereddy (commented):
Seems "ok" to me based on some local spot checking. I still note that the code (and comments) for placing 6 ticks/values between 0 and 100 % seems considerably more complex than is needed here.

For example, this passes tests and doesn't retrieve the numbers 0, 20, 40, 60, 80, 100 in a list and then calculate the length of the list instead of just i.e., `6`.

```diff
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
@@ -128,10 +128,7 @@ def plot_io_cost(report: darshan.DarshanReport) -> Any:
     ax_norm.set_ylim(0, 100)
     # convert the normalized axis y labels to percentages
     ax_norm.yaxis.set_major_formatter(mtick.PercentFormatter())
-    # align both axes tick labels so the grid matches
-    # values on both sides of the bar graph
-    n_ticks = len(ax_norm.get_yticks())
-    yticks = np.linspace(0, runtime, n_ticks)
+    yticks = np.linspace(0, runtime, 6)
     ax_raw.set_yticks(yticks)
     # add the legend and appropriate labels
     ax_raw.set_ylabel("Runtime (s)")
```

It looks like you don't actually have to normalize the data to display the normalized ticks on the plot either (one test is sensitive to this, but it is not actually detecting a problem as far as I can tell--as long as the display is normalized to 100%):

```diff
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
@@ -125,13 +125,9 @@ def plot_io_cost(report: darshan.DarshanReport) -> Any:
     ax_raw.grid(axis="y", zorder=0, alpha=0.6)
     # set the y limits for both axes
     ax_raw.set_ylim(0, runtime)
-    ax_norm.set_ylim(0, 100)
     # convert the normalized axis y labels to percentages
-    ax_norm.yaxis.set_major_formatter(mtick.PercentFormatter())
-    # align both axes tick labels so the grid matches
-    # values on both sides of the bar graph
-    n_ticks = len(ax_norm.get_yticks())
-    yticks = np.linspace(0, runtime, n_ticks)
+    ax_norm.yaxis.set_major_formatter(mtick.PercentFormatter(xmax=runtime))
+    yticks = np.linspace(0, runtime, 6)
     ax_raw.set_yticks(yticks)
     # add the legend and appropriate labels
     ax_raw.set_ylabel("Runtime (s)")
```

tylerjereddy:
merged, thx

PR #596, opened 2021-12-20
MAINT: Move `autolabel` function out of `plot_access_histogram`

nawtrey:
* Move `autolabel` function outside of `plot_access_histogram`
to allow for easier testing

* Add `ax` parameter to `autolabel` and correct call signatures

* Contributes to issue #577


nawtrey:
There will be changes to `autolabel` down the line that will be hidden in the diff if I move it at the same time, so pretty simple change here. At some point I think a unit test can be written for `autolabel`, but it's probably best to do after `plot_access_histogram` is updated to horizontal bar graphs since `autolabel` will have to be updated as well.

tylerjereddy (approved):
alright, thx

Issue #569, opened 2021-11-29
ENH: finalize `plot_op_counts()` for Darshan Summary Report


nawtrey:
`plot_opcounts.plot_opcounts()`, is a good starting point for the "I/O operation count" plot in issue #465, but there are some changes required before it will go in:
- [x] Change it to only plot (1) module's data 
- [x] Change the x-axis labels to better reflect their respective module's operations (give MPI-IO Independent and Collective values their own labels). 

Some other things I think we would have to address before adding to the report: 
- [x] Remove the title 
- [ ] Fix the bar graph value annotations and move the `autolabel` function, or remove the labels entirely (may not be necessary on a per-module basis)
- [x] Remove unused `as_csv()` function
- [ ] Add checks such that counters are only included in figure if they were enabled at runtime. It looks like MMap is the only one that is disabled by default, but I'll have to double-check

nawtrey:
I've added another task to the issue related to feedback from Phil. Basically we need to check to see if a given counter wasn't tracked, and omit them from the figure if so. This should give us a clear distinction between a counter with value 0 and a counter that wasn't tracked. Untracked counters should have a value of -1 for every record so it should be easy enough to catch.

nawtrey:
We are planning on moving to a horizontal bar graph in issue #577 because it will make the labels significantly easier to handle. I think it makes sense to do the same here.

Also, I think this issue along with #577 could go in the 3.4 milestone. I think they are probably low priority relative to other things, but it would be nice to have all of the report figures cleaned up for the Darshan release. Maybe @carns or @shanedsnyder can weigh in on that.

carns:
I agree with putting this in the 3.4 milestone.

shanedsnyder:
Closing this old issue. We don't currently plan to reorient the bar plots horizontally, annotation overlap with axes is addressed in #911, and we will use issue #904 to track the issue about removing counters with negative values from plots.

PR #598, opened 2021-12-20
MAINT: Refactor `plot_opcounts.py`

nawtrey:
* Abstract data handling code into new function `gather_count_data`

* Move `autolabel` function outside of `plot_opcounts`

* Add `ax` parameter to `autolabel` and correct call signatures

* Related to issue #569

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
@@ -4,19 +4,25 @@ import matplotlib.pyplot as plt
 import numpy as np
 
 
-def plot_opcounts(report, ax=None):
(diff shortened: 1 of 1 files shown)


nawtrey:
Similar to gh-596, I just want to move some of this code around before I start making changes to this figure so the future PR diffs are more clear.

A few notes:
1. The new function, `gather_count_data`, will be documented properly in a follow up PR
2. The `autolabel` function here is nearly identical to the one in `plot_access_histogram.py`, differing only by the `rotation` setting. It can be removed entirely with a simple modification to that function, but I figured that could be handled separately. We may not use it for this figure at all, so it may get removed entirely in one of the next PR's.

tylerjereddy (approved):
k thx

calling `autolabel` four times in a row probably means looping is in order

PR #599, opened 2021-12-22
MAINT: Single module `plot_opcounts`

nawtrey:
* Modify `plot_opcounts` to plot a single module

* Merge `MPIIO` collective and independent operations and update
the label names to better reflect the operations being used

* Remove unused `STDIO` "stat" and "mmap"
operations and change `Fsync` to `Flush`

* Correct bar annotation x-coordinate to
use bar midpoint in `autolabel`

* Remove the figure legend as there are no longer bar labels

* Correct tests in `test_plot_exp_common.py`:
  - Refactor parametrizations in `test_xticks_and_labels` to allow
  for plotting function and module specific x-axis tick labels
  - Fix `plot_opcounts` test cases in `test_bar_heights` by
  providing the module names and splitting up the original
  `expected_heights` arrays for their respective modules

* Addresses first 2 tasks in issue #569

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
@@ -10,7 +10,7 @@ def autolabel(ax, rects):
         height = rect.get_height()
         ax.annotate(
             '{}'.format(height),
-            xy=(rect.get_x() + rect.get_width() / 4 + rect.get_width(), height),
+            xy=(rect.get_x() + rect.get_width() / 2, height),
             xytext=(0, 3),  # 3 points vertical offset
             textcoords="offset points",
             ha='center',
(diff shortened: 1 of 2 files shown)


nawtrey:
This splits up the modules in `plot_opcounts` and plots them separately so they can be used in #465. At the same time, it updates the module operation names, combining the `MPIIO` names together so they are plotted together, and removing some that were not being used. 

Here's an example using `ior_hdf5_example.darshan`:

Before
-------
ior_hdf5_example darshan_before

After
-----
## `POSIX`
ior_hdf5_example darshan_POSIX
## `MPI-IO`
ior_hdf5_example darshan_MPI-IO
## `STDIO`
ior_hdf5_example darshan_STDIO

nawtrey:
I'm not sure if @shanedsnyder will have time to look this over, so here is his comment that inspired these changes:
> I was imagining splitting those up into per-module graphs. would be nice for this per-module thing to still support the ability to provide both MPI-IO independent vs collective calls, though

carns:
This looks great.  I like that the the plots have annotations with the exact value, that will particularly help when comparing across modules in cases where the scales differ.

tylerjereddy (approved):
Phil seems pretty happy with the appearance. I think the intersection of the value labels with the top border line might be worth fixing. Since the matching issue mentions "fixing" those labels I'll leave that to another PR.

tylerjereddy:
one other things to note is that the old report has roughly similar data plotted--can't really think of a good reason not to have a few logs repo log-based tests that enforce similarity with the old report (by inspection I think, the old report didn't label with numbers I don't think).

The same idea applies to at least one other plot I think, looking at the old perl reports..

PR #609, opened 2022-01-08
TST: tests for get_log_path()

tylerjereddy:
* add some basic regression tests for `get_log_path()`;
focuses on the usage in the `pytest` context, which is
the main use case

* no noticeable test suite slowdown measured locally with logs repo
present

### DIFF
--- a/darshan-util/pydarshan/tests/test_log_utils.py
+++ b/darshan-util/pydarshan/tests/test_log_utils.py
@@ -1,0 +1,55 @@
+import os
+
+import darshan
+from darshan.log_utils import get_log_path
+
+import pytest
+
+
+@pytest.mark.parametrize("log_filename", [
+    "noposixopens.darshan",
+    "sample-badost.darshan",
+    "sample-dxt-simple.darshan",
+    "noposix.darshan",
+    "sample-goodost.darshan",
+    "sample.darshan",
+    "pq_app_read_id71344_7-31-5658-2037904274838284930_55623.darshan",
+    "pq_app_readAB_writeC_id71326_7-31-5658-2037904274838284930_55623.darshan",
+    "pq_app_write_id71296_7-31-5657-2037904274838284930_55623.darshan",
+    "pq_app_write_id71310_7-31-5657-2037904274838284930_55623.darshan",
+    "pq_app_read_id71317_7-31-5657-2037904274838284930_55623.darshan",
+    "pq_app_write_id71303_7-31-5657-2037904274838284930_55623.darshan",
+    "dxt.darshan",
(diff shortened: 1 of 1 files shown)


tylerjereddy:
The failures are real--probably the `setup.py` and/or the lack of `__init__.py` are to blame for the specific path `./darshan-util/pydarshan/examples/darshan-graph/` to behave differently/not get searched/installed.

tylerjereddy:
I've added in an `__init__.py` at `pydarshan/examples/darshan-graph/__init__.py` so that `setuptools` automatically installs the `*.darshan` logs at that path as well. This should allow the tests to pass.

Note that pretty soon we'll have to stop installing `examples` and `tests` as their own Python packages--it isn't really reasonable for example for `darshan` to monopolize `import tests` and `import examples` solely for the purpose of the convenience of searching log files. I'll try to achieve this using `get_log_path()` in the test suite, followed by moving those two folders into `darshan` proper, but that is a refactor out of scope for this PR.

darshan-util/pydarshan/tests/test_log_utils.py
```diff
@@ -0,0 +1,55 @@
...
+    "sample-dxt-simple.darshan",
+    "noposix.darshan",
+    "sample-goodost.darshan",
+    "sample.darshan",
+    "pq_app_read_id71344_7-31-5658-2037904274838284930_55623.darshan",
+    "pq_app_readAB_writeC_id71326_7-31-5658-2037904274838284930_55623.darshan",
+    "pq_app_write_id71296_7-31-5657-2037904274838284930_55623.darshan",
+    "pq_app_write_id71310_7-31-5657-2037904274838284930_55623.darshan",
+    "pq_app_read_id71317_7-31-5657-2037904274838284930_55623.darshan",
+    "pq_app_write_id71303_7-31-5657-2037904274838284930_55623.darshan",
```
nawtrey, line 21: Do we want to include these files here? We *could* manipulate the glob search to remove them since they all start with the same name: `p.glob('**/[!pq_app_]*.darshan')`, but not a big deal either way.

nawtrey (commented):
I added a single comment, albeit minor. 

Tests pass locally, and they seem sensible. LGTM

darshan-util/pydarshan/tests/test_log_utils.py
tylerjereddy: I suspect if we really don't want to glob them in (i.e., they have no possible use, and they may not?), we should just delete them. I'll merge this as is then and if the team wants to dump those files after that's fine by me.

Issue #597, opened 2021-12-20
BLD: spack package adjustments for pydarshan


tylerjereddy:
@tylerjereddy should expand the current utils spackage (see: `spack info darshan-util`) to have `pydarshan` variants for the Python component of the package. Probably to be done locally on a branch of `spack`, and later deployed in our own custom "spack repo" before getting upstreamed.

tylerjereddy:
@carns using branch in https://github.com/tylerjereddy/spack/pull/2, I currently encounter some issues, detailed there, when I try to add on the Python part of the spackage. One of the dangers is that we end up needing custom logic to get the Python install/test working I think (?), all of which is normally handled by the `spack` `PythonPackage` class, while `darshan-util` is fundamentally a `AutotoolsPackage` class/process.

Current incantation is i.e., `spack install darshan-util@pydarshan-devel+apmpi+apxc+pydarshan % gcc@7.4.0`

I suspect we may be better off, short term, with a separate spackage with `spack` `PythonPackage` class usage for `pydarshan`, even if we don't use that solution long-term, so we can avoid our own hacks of `PYTHONPATH` or whatever. If we do that, then it might be something like:

`spack install py-pydarshan@pydarshan-devel+apmpi+apxc % gcc@7.4.0`

tylerjereddy:
That said, the only way to know is probably to try both ways, and see which one is worse. Maybe I'll get to that later today.

carns:
Ah, I see.  I'm in favor of whatever seems most expedient for now.

Possibly in the longer run we could consider a package refactoring, so that darshan-util essentially becomes a meta package that pulls in pydarshan and darshan-c-utils (that's probably not the right name, but you see what I mean) so that you can still get everything by installing the same darshan-util package name as ever, but it's really pulling sub-packages with appropriate build systems for each piece.

tylerjereddy:
@carns using https://github.com/tylerjereddy/spack/pull/3 I can get a basic install to work on a Linux box:

`spack install py-darshan % gcc@7.4.0`

Details on what does/doesn't work is over in that PR/branch. It might be worth testing on one of your machines to see if you reproduce my problem (that kind of `gfortran` symbol error can be flaky, so you may just luck out and succeed on one of your machines). That won't be fun to fix.. SciPy is also moving to `meson` as a new build system, so the team probably wouldn't prioritize a fix, though I could dig deeper.

[...]

PR #608, opened 2022-01-08
BUG, BENCH: fix time_plot_heatmap_builtin_logs

tylerjereddy:
Fixes #586 

* fix the call signature of `plot_heatmap()` in
`time_plot_heatmap_builtin_logs()` so that the
benchmark no longer fails

* locally, ` asv run -e -b "time_plot_heatmap_builtin_logs"`:

```
· Creating environments
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[  0.00%] · For pydarshan commit 6d0a48e0 <pydarshan-devel>:
[  0.00%] ·· Benchmarking virtualenv-py3.9-cffi-numpy-pytest
[ 50.00%] ··· Running (dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs--).
[100.00%] ··· dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs                                                                                                                                                                                                                                              ok
[100.00%] ··· ================================================ ========= ========== ============
              --                                                             xbins              
              ------------------------------------------------ ---------------------------------
                              darshan_logfile                      10       100         1000    
              ================================================ ========= ========== ============
               examples/example-logs/ior_hdf5_example.darshan   354±2ms   510±30ms   1.20±0.03s 
                     examples/example-logs/dxt.darshan          352±6ms   471±3ms    1.19±0.01s 
                   tests/input/sample-dxt-simple.darshan        307±4ms   416±6ms    1.10±0.03s 
              ================================================ ========= ========== ============
```

PR #612, opened 2022-01-09
MAINT: refactor assets

tylerjereddy:
Supports the testing of gh-597 on HPC installs with `pytest --pyargs darshan`
from any directory (HPC users should not have to clone `darshan` to test
an already-installed HPC version).

* move `examples` and `tests` folders inside
`darshan` package so they can be easily
installed as elements of the package, and prevent us
from having to search more broadly on the filesystem
to locate `*.darshan` file assets (and to avoid monopolizing
`import tests`, `import examples`, which is crazy)

* adjust setup.py accordingly--there is now only a
`darshan` package to `install`, not separate `tests`
and `examples` packages

* migrate the test suite over to `get_log_path()` as needed
to empower a portable approach to running tests (i.e.,
`pytest --pyargs darshan` from any directory)

* `_locate_log()` no longer needs to go searching around
on the filesystem outside of our own packages, so get
rid of that..

* on this branch, in some random directory,
`pytest --pyargs darshan -n 8` produces:
`255 passed in 88.93s (0:01:28)`

* this is in line with latest `pydarshan-devel` run
of the full suite before migration to `get_log_path()`
and reorganization:
`pytest --import-mode=importlib -n 8`
`255 passed in 87.78s (0:01:27)`

* enforce support for `pytest --pyargs darshan` from any
dir in the CI (using a random temp dir there)

* things reviewers could do to help here:
  - check the large diff for potential errors
  - check/fix up the `asv` benchmark suite for the
    effects of this branch
  - check if the coverage report stuff is still working
    "ok" with new `pytest` portability/incantation


tylerjereddy:
`mypy` says `example-logs is not a valid Python package name`, which I agree with, but I'll see if I can override for now...

tylerjereddy:
The benchmark suite should probably just use `get_log_path()` at this point--for now, I've just adjusted the current infrastructure there to compensate for the folder movements (that doesn't mean I've checked the `asv` suite runs properly beyond `asv check`, so item for reviewers above remains valid).

tylerjereddy:
Well, CI is finally passing. I'll leave the rest up to reviewers, including the codecov stuff (changing paths may require a merge to re-enable codecov, but I'll let reviewers investigate that I guess).

nawtrey:
I'm gonna take a look at this today.

tylerjereddy:
feel free to ping `@thomasrockhu` on the codecov team if you get stuck or it gets too time consuming. It could be as simple as changing a path or even just gets turned back on after merging a base report with the new paths.

nawtrey (commented):
I've added some comments related to the CI. I can say that in all of the `get_log_path`-related changes I didn't find anything out of place. 

I was able to successfully run `pytest --pyargs darshan -n 8` in a random dir, so I can confirm that is working as intended for me as well, including when `darshan-logs` unavailable. `mypy darshan` also seems to work, and covers all files now with a single command, which is nice.

I would recommend updating `test_plot_io_cost.py::test_issue_590` to use `get_log_path` because it is the last test using `select_log_repo_file`. With the changes in the diff below, I was able to remove `conftest.select_log_repo_file` and run the test suite without error.
```diff
diff --git a/darshan-util/pydarshan/darshan/tests/test_plot_io_cost.py b/darshan-util/pydarshan/darshan/tests/test_plot_io_cost.py
index 636b850..fa72e09 100644
--- a/darshan-util/pydarshan/darshan/tests/test_plot_io_cost.py
+++ b/darshan-util/pydarshan/darshan/tests/test_plot_io_cost.py
@@ -192,8 +192,6 @@ def test_get_by_avg_series(mod_key, input_df, expected_series):
     assert_series_equal(actual_series, expected_series)


-@pytest.mark.skipif(not pytest.has_log_repo, # type: ignore
-                    reason="missing darshan_logs")
 @pytest.mark.parametrize(
     "filename, expected_df",
     [
@@ -209,10 +207,10 @@ def test_get_by_avg_series(mod_key, input_df, expected_series):
             ),
         ),
     ])
-def test_issue_590(select_log_repo_file, expected_df):
+def test_issue_590(filename, expected_df):
     # regression test for issue #590
     # see: https://github.com/darshan-hpc/darshan/issues/590
-    log_path = select_log_repo_file
+    log_path = get_log_path(filename)
     report = darshan.DarshanReport(log_path)
     actual_df = get_io_cost_df(report=report)
     assert_frame_equal(actual_df, expected_df)
```

As for the benchmarks, I ran `asv run` locally and the only one that is failing for me is the benchmark addressed in gh-608. So I think if that gets merged, that should fix the issue there. 

Lastly, I'm not sure what the best way to check the `setup.py` changes, but I believe the CI is setup in such a way that this is being tested. So as long as our tests are passing there, I think we are good?

tylerjereddy:
I think I've pushed in commits to address all of Nik's suggestions so far. I'm working rather quickly without testing locally first though.

For the codecov stuff, I've added a small amount of debug code to print out some paths--maybe we can iterate to specify the path manually for now based on what gets printed out in the logs (this will also help us confirm if we're actually in a temporary directory or not when running `pytest ..`).

.github/workflows/main_ci.yml
```diff
@@ -90,5 +90,5 @@ jobs:
       - name: codecov check
         uses: codecov/codecov-action@v2
         with:
-          files: coverage.xml
+          files: $RUNNER_TEMP/coverage.xml,/home/runner/work/_temp/coverage.xml,/Users/runner/work/_temp/coverage.xml
```
tylerjereddy, line 93: @thomasrockhu Any idea why I have to specify the absolute paths here instead of `codecov` action being able to expand the `$RUNNER_TEMP` variable? For now, I'll just use this, but would be more maintainable if we could use the first path in this list only.

tylerjereddy:
Ok, I think I've addressed all reviewer comments and codecov has been restored to CI thanks to absolute paths for report files. That's not super maintainable, but I'd like to get this in as apart of the infrastructure improvements to describe in the report.

tylerjereddy:
Folks may now need to resolve some merge conflicts in their PRs, and of course be vigilant for the next little while since this was a big change.

.github/workflows/main_ci.yml
thomasrockhu: yeah, that's pretty confusing to me. Are you able to run in verbose mode? I think we'd get a better understanding there.

Issue #550, opened 2021-11-02
ENH: probably support addition of tables in the new report API


tylerjereddy:
The Argonne team seems interested in being able to use the new report API to add tables on top of (matplotlib) figures eventually per discussion in gh-526.

tylerjereddy:
fixed by gh-618

PR #618, opened 2022-01-13
ENH: Add dataframe/table figure support

nawtrey:
* Update method `generate_img` name to more general `generate_fig`
and update documentation to reflect the changes

* Generalize `generate_fig` to support figure
functions that generate pandas dataframes

* Add test `test_generate_fig_unsupported_fig_type` to
check appropriate error is raised when input function
does not generate the supported figure types

* Add object `DarshanReportTable` to `plot_common_access_table.py`
for storing tables in dataframe and html format. Should set a
standard that allows access for testing but also puts the burden
on the figure function to generate a `DarshanReportTable` that
contains the desired html table for the report.

* Change `plot_common_access_table` to return
`DarshanReportTable` and update documentation

* Fix `test_common_access_table` due to above change

* Fix issue #550

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -13,12 +13,14 @@ from typing import Any, Union, Callable
 
 import pandas as pd
 from mako.template import Template
-import matplotlib
-import seaborn
 
 import darshan
 import darshan.cli
-from darshan.experimental.plots import plot_dxt_heatmap, plot_io_cost
+from darshan.experimental.plots import (
+    plot_dxt_heatmap,
+    plot_io_cost,
+    plot_common_access_table,
+)
 
 darshan.enable_experimental()
 
@@ -90,20 +92,16 @@ class ReportFigure:
         # generate the figure using the figure's
         # function and function arguments
         fig = self.fig_func(**self.fig_args)
-        supported_mpl_fig_types = [matplotlib.figure.Figure, seaborn.axisgrid.JointGrid]
-        if type(fig) in supported_mpl_fig_types:
+        if hasattr(fig, "savefig"):
             # encode the matplotlib figure
             encoded = self.get_encoded_fig(mpl_fig=fig)
             # create the img string
             self.fig_html = f"<img src=data:image/png;base64,{encoded} alt={self.fig_title} width={self.fig_width}>"
-        elif isinstance(fig, pd.DataFrame):
-            # use built-in pandas utility to convert table to html
-            self.fig_html = fig.to_html(index=False, border=0)
+        elif isinstance(fig, plot_common_access_table.DarshanReportTable):
+            # retrieve html table from `DarshanReportTable`
+            self.fig_html = fig.html
         else:
-            err_msg = (
-                f"Figure of type {type(fig)} not supported. \n"
-                f"Supported figure types: {supported_mpl_fig_types}"
-            )
+            err_msg = f"Figure of type {type(fig)} not supported."
             raise NotImplementedError(err_msg)
 
 class ReportData:
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -13,6 +13,8 @@ from typing import Any, Union, Callable
 
 import pandas as pd
 from mako.template import Template
+import matplotlib
+import seaborn
 
 import darshan
 import darshan.cli
@@ -58,9 +60,9 @@ class ReportFigure:
         # temporary handling for DXT disabled cases
         # so special error message can be passed
         # in place of an encoded image
-        self.img_str = None
+        self.fig_html = None
         if self.fig_func:
-            self.generate_img()
+            self.generate_fig()
 
     @staticmethod
     def get_encoded_fig(mpl_fig: Any):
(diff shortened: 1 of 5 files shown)


darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -81,17 +83,28 @@ def get_encoded_fig(mpl_fig: Any):
...
+        # generate the figure using the figure's
         # function and function arguments
-        mpl_fig = self.fig_func(**self.fig_args)
-        # encode the matplotlib figure
-        encoded = self.get_encoded_fig(mpl_fig=mpl_fig)
-        # create the img string
-        self.img_str = f"<img src=data:image/png;base64,{encoded} alt={self.fig_title} width={self.fig_width}>"
+        fig = self.fig_func(**self.fig_args)
+        supported_mpl_fig_types = [matplotlib.figure.Figure, seaborn.axisgrid.JointGrid]
+        if type(fig) in supported_mpl_fig_types:
```
nawtrey, line 94: I'm not ecstatic about this bit because the _real_ requirement is that the figure has a `.savefig()` method, so this might be overly strict. But I think we do want to catch cases where incorrect figure types are generated instead of failing at `.savefig()` downstream..

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -81,17 +83,28 @@ def get_encoded_fig(mpl_fig: Any):
...
+            # encode the matplotlib figure
+            encoded = self.get_encoded_fig(mpl_fig=fig)
+            # create the img string
+            self.fig_html = f"<img src=data:image/png;base64,{encoded} alt={self.fig_title} width={self.fig_width}>"
+        elif isinstance(fig, pd.DataFrame):
+            # use built-in pandas utility to convert table to html
+            self.fig_html = fig.to_html(header=False, border=0)
```
nawtrey, line 101: This line is currently not covered by the CI since there are no tests for it. This will have to be merged before we can add the "common access table" figure, but with that will come tests that will certainly cover this.

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -81,17 +83,28 @@ def get_encoded_fig(mpl_fig: Any):
...
+        fig = self.fig_func(**self.fig_args)
+        supported_mpl_fig_types = [matplotlib.figure.Figure, seaborn.axisgrid.JointGrid]
+        if type(fig) in supported_mpl_fig_types:
+            # encode the matplotlib figure
+            encoded = self.get_encoded_fig(mpl_fig=fig)
+            # create the img string
+            self.fig_html = f"<img src=data:image/png;base64,{encoded} alt={self.fig_title} width={self.fig_width}>"
+        elif isinstance(fig, pd.DataFrame):
+            # use built-in pandas utility to convert table to html
+            self.fig_html = fig.to_html(index=False, border=0)
```
nawtrey, line 101: I changed this from `header=False` to `index=False`, since that is what works best with the common access size table, but this obviously won't work for every case (i.e. the metadata and module tables use `header=False`). We will likely have to come up with a way to choose which one to remove, because so far the tables have always needed one of these turned off.

nawtrey:
Alright, rebased/force pushed after the merge of gh-612.

darshan-util/pydarshan/darshan/cli/summary.py
tylerjereddy: Why not use the real requirement then? This will let things that quack like a duck through.

```diff
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -13,8 +13,6 @@ from typing import Any, Union, Callable
 
 import pandas as pd
 from mako.template import Template
-import matplotlib
-import seaborn
 
 import darshan
 import darshan.cli
@@ -90,8 +88,7 @@ class ReportFigure:
         # generate the figure using the figure's
         # function and function arguments
         fig = self.fig_func(**self.fig_args)
-        supported_mpl_fig_types = [matplotlib.figure.Figure, seaborn.axisgrid.JointGrid]
-        if type(fig) in supported_mpl_fig_types:
+        if hasattr(fig, "savefig"):
             # encode the matplotlib figure
             encoded = self.get_encoded_fig(mpl_fig=fig)
             # create the img string
@@ -101,8 +98,7 @@ class ReportFigure:
             self.fig_html = fig.to_html(index=False, border=0)
         else:
             err_msg = (
-                f"Figure of type {type(fig)} not supported. \n"
-                f"Supported figure types: {supported_mpl_fig_types}"
+                f"Figure of type {type(fig)} not supported."
             )
             raise NotImplementedError(err_msg)
```

darshan-util/pydarshan/darshan/cli/summary.py
tylerjereddy: Maybe leave this to the author of the table code for any given case? You could mandate a small custom class/object is returned for tables that gives you a method/attribute or similar that retrieves the html (which is expected to be formatted properly, otherwise it is up to the author to fix it).

```python
elif isinstance(fig, DarshanReportTable):
    # the function returning DarshanReportTable is
    # is responsible for formatting the html
    self.fig_html = DarshanReportTable.html
```

At the moment you could even get away with just providing the raw html without a custom class, but this may not be the best design if we plan on supporting more custom things later, like a class/object for traffic light heuristics, etc.

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: Thanks for the suggestion. It was sort of an afterthought that I forgot to circle back to, but I'll implement these changes.

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: Thanks, I like that idea. I'll see if I can implement that on the sibling PR along with your example.

nawtrey:
I think I've made all of the suggested changes here, including adding a `DarshanReportTable` object. `DarshanReportTable` could probably be in a more accessible location, but for now it is only leveraged by `plot_common_access_table()` so I figured it was fine to place it in the same module.

tylerjereddy:
Ok, I'll put it in, thanks.

Note that `plot_common_access_table` no longer returns a type of `Any` since you've just explicitly defined a new type that it returns, so that can be made more specific for `mypy`.

`DarshanReportTable` should be in a general data structure-related module somewhere most likely, I'll leave that up to you to discuss with the team.

PR #619, opened 2022-01-13
ENH: Add common access size table to darshan summary report

nawtrey:
* Add "Per-Module Stats" section to report
containing the common access size table (to start)

* Extend `test_main_without_args` to check for number of tables
generated as a regression guard

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -389,6 +389,28 @@ class ReportData:
         io_cost_fig = ReportFigure(**io_cost_params)
         self.figures.append(io_cost_fig)
 
+        ################################
+        ## Per-Module Statistics
+        ################################
+        for mod in self.report.modules:
+            if mod in ["POSIX", "MPI-IO", "H5D"]:
+                if mod == "MPI-IO":
+                    com_acc_tbl_description = (
+                        "NOTE: MPI-IO accesses are given in "
+                        "terms of aggregate datatype size."
+                    )
+                else:
+                    com_acc_tbl_description = ""
+                com_acc_tbl_fig = ReportFigure(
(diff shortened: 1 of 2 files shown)


darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -378,6 +382,28 @@ def register_figures(self):
...
+            if mod in ["POSIX", "MPI-IO", "H5D"]:
+                if mod == "MPI-IO":
+                    com_acc_tbl_description = (
+                        "NOTE: MPI-IO accesses are given in "
+                        "terms of aggregate datatype size."
+                    )
+                else:
+                    com_acc_tbl_description = ""
+                com_acc_tbl_fig = ReportFigure(
+                    section_title=f"Per-Module Stats: {mod}",
```
nawtrey, line 398: I've titled the new section "Per-Module Stats: MOD", but based on issue [#465](https://github.com/darshan-hpc/darshan/issues/465), maybe this should be "Per-Module Characteristics: MOD"?

nawtrey:
Alright, I rebased/force-pushed after the merge of gh-618

tylerjereddy (approved):
For `sample-dxt-simple` I think I slightly prefer the compact single-table representation of the old report vs. separate tables. I'll assume the Argonne folks asked for what you did here vs. the old one though, or that you've at least touched base with them about the change.

That said, I'm not really that picky about the html aesthetics, and the code changes look "ok", so I'll merge thanks.

image

vs. old report:

image

Similar delta for the `snyder_acme..` file:

image

image

carns:
I'm happy for them to be separate now.  I do agree that the older-style combined table looks nicer from an aesthetic perspective, but I like that the new approach is going to be more modular regardless of what APIs are in use.

Issue #576, opened 2021-12-02
BUG: heatmap anomalies for `snyder_acme.exe_id1253318_9-27-24239-1515303144625770178_2.darshan`


nawtrey:
@treddy highlighted some interesting behavior when generating the DXT heat maps for `snyder_acme.exe_id1253318_9-27-24239-1515303144625770178_2.darshan` (shown below). His original [comment](https://github.com/darshan-hpc/darshan/pull/570#pullrequestreview-820567121). 

It appears we have a few features to explain:
 - Why do the horizontal sums appear to be disproportionate to the activity shown in the heatmap? (Tyler's original question). Specifically, why, in the `DXT_POSIX` plot, do we see 2 ranks with almost identical activity but different horizontal sums?
 - Why are the rank counts different between the plots? 

![image](https://user-images.githubusercontent.com/46797896/144519634-164e7809-1228-45f8-803b-aab49c21536e.png)

tylerjereddy:
gh-622 should have fixed the rank counts, but not likely the resolution issues

tylerjereddy:
The issue is called `DXT heatmap vertical bar graph inconsistency` and seems to focus on the vertical bar issues, but I've added a comment to encapsulate the horizontal issues as well for the case that originally concerned me: https://github.com/darshan-hpc/darshan/issues/630#issuecomment-1035610509

I'll close this and then then that issue can be used to address the issue along both axes then

Issue #575, opened 2021-12-02
DXT heat map plotting code updates


nawtrey:
There were a few changes suggested in gh-570 for the DXT heatmap plotting code that I'll describe here. These will likely be handled in separate PR's since some will likely be difficult enough on their own. 

Suggested changes:
- [ ] Use stored `end_time` to set the x-axis maximum value, to make comparisons across modules easier (and integrate easier with darshan-runtime heatmap eventually). See comments [1](https://github.com/darshan-hpc/darshan/pull/570#pullrequestreview-820567121) and [2](https://github.com/darshan-hpc/darshan/pull/570#issuecomment-983963683).
- [ ]  ~Find a way to scale the heatmap colorbar such that a single order of magnitude has a noticeable change in color.~
- [x] Use `nprocs - 1` to set the y axis maximum 
- [ ] ~Find a way to label the vertical & horizontal bar graphs such that a user can easily get an idea of how much data is in a given rank. Phil suggested adding some text that explains what the largest value is and what rank it corresponds to (i.e. something like "rank 3 read/wrote the most data @ 200 MiB").~
- [x] Make sure it is compliant with the new runtime heatmap data

I'll probably stave off the majority of these changes until the runtime-heatmap data is accessible, as discussed in today's meeting.

tylerjereddy:
Are there issues open for the anomalies in the heatmaps for the snyder log (weird horizontal sums/rank activity patterns?). It looks like those may be indirectly referenced in comment links above, but only described as a visual matter rather than a potential bug somewhere, so you may want to split those off due to potential severity/clarity.

nawtrey:
That's a good point, I'll open a separate issue for those since they appear to be a bug.

nawtrey:
Out of the remaining tasks, I think tasks 1 and 5 (and arguably 4) are Darshan 3.4 material:
- Task 1 should allow for easier comparison with the runtime heatmaps
- Task 5 is likely going to be handled when the runtime heatmaps get added to the report, if not before
- Task 4 probably isn't too bad to check off, but probably isn't a requirement

If we agree on which tasks we want I can move the non-milestone task(s) to separate issues and mark this one as 3.4.

carns:
Agreed.  I would suggest just putting tasks for 1 and 5 into the 3.4 milestone, and capturing the others outside of a milestone for now but just labeled with the pydarshan tag.

nawtrey:
[...]

PR #622, opened 2022-01-14
BUG: Change DXT heatmap to use `nprocs` for y-axis scale and correct bar graph anomalies

nawtrey:
* Use `nprocs` to scale y-axis for heatmap figure
instead of number of ranks with IO activity

* Add `nprocs` parameter to `get_heatmap_df`
and update documentation

* Update `get_heatmap_df` to return dense dataframe based on `nprocs`

* Correct tests in `test_plot_dxt_heatmap.py` for cases 
where not all ranks had non-zero bins

* Correct tests in `test_heatmap_handling.py` to reflect
dense heatmap dataframe output

* Fixes issue #576

* Contributes to issue #575

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -273,7 +273,7 @@ def get_aggregate_data(
     return agg_df
 
 
-def get_heatmap_df(agg_df: pd.DataFrame, xbins: int) -> pd.DataFrame:
+def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFrame:
     """
     Builds an array similar to a 2D-histogram, where the y data is the unique
     ranks and the x data is time. Each bin is populated with the data sum
@@ -287,6 +287,8 @@ def get_heatmap_df(agg_df: pd.DataFrame, xbins: int) -> pd.DataFrame:
     by the input modules and operations.
 
     xbins: the number of x-axis bins to create.
+
+    nprocs: the number of MPI ranks/processes used at runtime.
 
     Returns
     -------
@@ -379,33 +381,5 @@ def get_heatmap_df(agg_df: pd.DataFrame, xbins: int) -> pd.DataFrame:
     cats = cats.mul(agg_df["length"], axis=0)
     cats.index = agg_df["rank"]
     hmap_df = cats.groupby("rank").sum()
+    hmap_df = hmap_df.reindex(index=range(nprocs), fill_value=0.0)
     return hmap_df
-
-
-def get_filled_hmap_df(hmap_df: Any, nprocs: int) -> Any:
-    """
-    Takes a dataframe containing a subset of ranks and injects them into
-    a new dataframe with a row for each rank.
-
-    Parameters
-    ----------
-
-    hmap_df: dataframe with time intervals for columns and rank
-    index (0, 1, etc.) for rows, where each element contains the data
-    read/written by the corresponding rank in the given time interval.
-
-    nprocs: the number of MPI ranks/processes used at runtime.
-
-    Returns
-    -------
-
-    filled_df: dataframe of shape ``nprocs x xbins`` with same intervals as
-    `hmap_df`, all data from `hmap_df`, but rows that span from 0 to `nprocs-1`.
-
-    """
-    # make an empty dataframe of size `nprocs x xbins`
-    # and fill it with the data from above dataframe
-    filled_df = pd.DataFrame(np.zeros((nprocs, hmap_df.shape[1])), columns=hmap_df.columns)
-    filled_df.index.name = "rank"
-    filled_df.update(hmap_df)
-    return filled_df
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -380,3 +380,32 @@ def get_heatmap_df(agg_df: pd.DataFrame, xbins: int) -> pd.DataFrame:
     cats.index = agg_df["rank"]
     hmap_df = cats.groupby("rank").sum()
     return hmap_df
+
+
+def get_filled_hmap_df(hmap_df: Any, nprocs: int) -> Any:
+    """
+    Takes a dataframe containing a subset of ranks and injects them into
+    a new dataframe with a row for each rank.
+
+    Parameters
+    ----------
+
+    hmap_df: dataframe with time intervals for columns and rank
+    index (0, 1, etc.) for rows, where each element contains the data
+    read/written by the corresponding rank in the given time interval.
+
+    nprocs: the number of MPI ranks/processes used at runtime.
+
+    Returns
+    -------
+
+    filled_df: dataframe of shape ``nprocs x xbins`` with same intervals as
+    `hmap_df`, all data from `hmap_df`, but rows that span from 0 to `nprocs-1`.
+
+    """
+    # make an empty dataframe of size `nprocs x xbins`
+    # and fill it with the data from above dataframe
+    filled_df = pd.DataFrame(np.zeros((nprocs, hmap_df.shape[1])), columns=hmap_df.columns)
+    filled_df.index.name = "rank"
+    filled_df.update(hmap_df)
+    return filled_df
--- a/darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
@@ -457,7 +457,7 @@ def test_plot_heatmap(filepath, mod, ops):
         assert jgrid.ax_joint.has_data()
         if "dxt.darshan" in filepath:
             # verify the horizontal bar graph does not contain data since there
-            # are is only 1 rank for this case
+            # is only 1 rank for this case
             assert not jgrid.ax_marg_y.has_data()
         else:
             # verify the horizontal bar graph contains data for multirank cases
--- a/darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
@@ -242,10 +242,10 @@ def test_get_y_axis_tick_labels(
         # check that if we ask for more y-axis labels than
         # available, we still get the same output
         ("dxt.darshan", 4, [0.5], [0]),
-        ("sample-dxt-simple.darshan", 2, [0.5], [0.0]),
+        ("sample-dxt-simple.darshan", 2, [0.5, 15.5], [0.0, 15.0]),
         # check that if we ask for more y-axis labels than
         # available, we still get the same output
-        ("sample-dxt-simple.darshan", 4, [0.5], [0.0]),
+        ("sample-dxt-simple.darshan", 4, [0.5, 5.5, 10.5, 15.5], [0.0, 5.0, 10.0, 15.0]),
     ],
 )
 def test_set_y_axis_ticks_and_labels(
@@ -266,16 +266,14 @@ def test_set_y_axis_ticks_and_labels(
 
     # x-axis bins are arbitrary
     xbins = 100
+    nprocs = report.metadata["job"]["nprocs"]
 
     # generate the heatmap data
     data = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=xbins)
-
-    # use the unique ranks to get the number of y-axis bins to use
-    unique_ranks = np.unique(agg_df["rank"].values)
-    ybins = unique_ranks.size
+    data = heatmap_handling.get_filled_hmap_df(hmap_df=data, nprocs=nprocs)
 
     # generate a joint plot object, then add the heatmap to it
-    jointgrid = sns.jointplot(kind="hist", bins=[xbins, ybins])
+    jointgrid = sns.jointplot(kind="hist", bins=[xbins, nprocs])
     sns.heatmap(data, ax=jointgrid.ax_joint)
 
     # set the x-axis ticks and tick labels
@@ -356,55 +354,53 @@ def test_adjust_for_colorbar(filepath):
     assert hmap_positions.x0 == 0.1
     assert hmap_positions.y0 == 0.15000000000000002
     assert hmap_positions.y1 == 0.774390243902439
-    if "ior_hdf5_example.darshan" in filepath:
-        # since `ior_hdf5_example.darshan` has 4 ranks, it has
-        # different x max values because it needs room for
+    if "dxt.darshan" in filepath:
+        # since `dxt.darshan` has 1 rank, it has
+        # different x max values because it doesn't need room for
         # the colorbar on the outside of the horizontal bar graph
-        assert hmap_positions.x1 == 0.7158709677419354
-    else:
         assert hmap_positions.x1 == 0.7824516129032258
+    else:
+        assert hmap_positions.x1 == 0.7158709677419354
 
     # get vertical bar graph positions
     vert_bar_positions = jgrid.ax_marg_x.get_position()
     assert vert_bar_positions.x0 == 0.1
     assert vert_bar_positions.y0 == 0.7780487804878049
     assert vert_bar_positions.y1 == 0.9
-    if "ior_hdf5_example.darshan" in filepath:
-        # since `ior_hdf5_example.darshan` has 4 ranks, the vertical
-        # bar graph has a different x max value because it needs room for
+    if "dxt.darshan" in filepath:
+        # since `dxt.darshan` has 1 rank, the vertical
+        # bar graph has a different x max value because it doesn't need room for
         # the colorbar on the outside of the horizontal bar graph
-        assert vert_bar_positions.x1 == 0.7158709677419354
-    else:
         assert vert_bar_positions.x1 == 0.7824516129032258
+    else:
+        assert vert_bar_positions.x1 == 0.7158709677419354
 
     # get horizontal bar graph positions
     horiz_bar_positions = jgrid.ax_marg_y.get_position()
     assert horiz_bar_positions.y0 == 0.15000000000000002
     assert horiz_bar_positions.y1 == 0.774390243902439
-    if "ior_hdf5_example.darshan" in filepath:
-        # since `ior_hdf5_example.darshan` has 4 ranks, the horizontal
-        # bar graph has different x min/max values because it has to
+    if "dxt.darshan" in filepath:
+        # since `dxt.darshan` has 1 rank, the horizontal
+        # bar graph has different x min/max values because it doesn't need to
         # make room for the colorbar
-        assert horiz_bar_positions.x0 == 0.7206451612903225
-        assert horiz_bar_positions.x1 == 0.84
-    else:
         assert horiz_bar_positions.x0 == 0.7877419354838711
         assert horiz_bar_positions.x1 == 0.92
+    else:
+        assert horiz_bar_positions.x0 == 0.7206451612903225
+        assert horiz_bar_positions.x1 == 0.84
 
     # get the colorbar positions
     cbar_positions = jgrid.fig.axes[-1].get_position()
     assert cbar_positions.y0 == 0.15000000000000002
     assert cbar_positions.y1 == 0.774390243902439
-    if "ior_hdf5_example.darshan" in filepath:
-        # since `ior_hdf5_example.darshan` has 4 ranks, the colorbar has
-        # to go closer to the edge of the figure so the horizontal bar
-        # graph can fit in the panel
-        assert cbar_positions.x0 == 0.85
-        assert cbar_positions.x1 == 0.8716135084427767
-    else:
+    if "dxt.darshan" in filepath:
+        # since `dxt.darshan` has 1 rank, the colorbar doesn't have
+        # to go closer to the edge of the figure
         assert cbar_positions.x0 == 0.82
         assert cbar_positions.x1 == 0.8416135084427767
-
+    else:
+        assert cbar_positions.x0 == 0.85
+        assert cbar_positions.x1 == 0.8716135084427767
 
 @pytest.mark.parametrize(
     "filepath",
@@ -460,13 +456,13 @@ def test_plot_heatmap(filepath, mod, ops):
         # horizontal bar graph does not exist
         assert jgrid.ax_marg_x.has_data()
         assert jgrid.ax_joint.has_data()
-        if "ior_hdf5_example.darshan" in filepath:
-            # verify the horizontal bar graph contains data since there
-            # are multiple ranks for this case
-            assert jgrid.ax_marg_y.has_data()
-        else:
-            # verify the horizontal bar graph does not contain data
+        if "dxt.darshan" in filepath:
+            # verify the horizontal bar graph does not contain data since there
+            # are is only 1 rank for this case
             assert not jgrid.ax_marg_y.has_data()
+        else:
+            # verify the horizontal bar graph contains data for multirank cases
+            assert jgrid.ax_marg_y.has_data()
 
         # check that the axis labels are as expected
         assert jgrid.ax_joint.get_xlabel() == "Time (s)"
--- a/darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
@@ -269,8 +269,7 @@ def test_set_y_axis_ticks_and_labels(
     nprocs = report.metadata["job"]["nprocs"]
 
     # generate the heatmap data
-    data = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=xbins)
-    data = heatmap_handling.get_filled_hmap_df(hmap_df=data, nprocs=nprocs)
+    data = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=xbins, nprocs=nprocs)
 
     # generate a joint plot object, then add the heatmap to it
     jointgrid = sns.jointplot(kind="hist", bins=[xbins, nprocs])
--- a/darshan-util/pydarshan/benchmarks/benchmarks/dxt_heatmap.py
+++ b/darshan-util/pydarshan/benchmarks/benchmarks/dxt_heatmap.py
@@ -60,4 +60,4 @@ class GetHeatMapDf:
     def time_get_heatmap_df(self, unique_ranks, bin_count):
         # benchmark get_heatmap_df() handling of variable
         # numbers of unique ranks/bins
-        heatmap_handling.get_heatmap_df(self.agg_df, xbins=bin_count)
+        heatmap_handling.get_heatmap_df(self.agg_df, xbins=bin_count, nprocs=unique_ranks)
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
@@ -374,7 +374,7 @@ def plot_heatmap(
     for _, spine in hmap.spines.items():
         spine.set_visible(True)
 
-    # if there is more than 1 processes,
+    # if there is more than 1 process,
     # create the horizontal bar graph
     if nprocs > 1:
         jgrid.ax_marg_y.barh(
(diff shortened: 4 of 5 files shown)


nawtrey:
This is targeted at the second task in #575 (setting y-axis based on `nprocs`), but the end result appears to fix the anomalies highlighted in #576. 

The changes here have changed the outcomes for some other plots as well since we are now plotting a heatmap row for each processes, regardless of its presence in the DXT module data. For example, for logs such as `sample-dxt-simple.darshan` where we previously only listed a single rank, we now list all 16 ranks despite 15 of them containing no data. I think this makes a little more sense so users can see the number of processes they are expecting, and see what each one is doing, even if it is nothing. If we wanted we could even make that configurable later if users wanted to "zoom in" on only the ranks that have data, although I did find it difficult to fix this issue without filling in the blanks.

### `snyder_acme.exe_..._2.darshan` Before:
## `DXT_POSIX`:
snyder_acme exe_id1253318_9-27-24239-1515303144625770178_2 darshan_DXT_POSIX

## `DXT_MPIIO`:
snyder_acme exe_id1253318_9-27-24239-1515303144625770178_2 darshan_DXT_MPIIO

### `snyder_acme.exe_..._2.darshan` After: 
## `DXT_POSIX`:
snyder_acme exe_id1253318_9-27-24239-1515303144625770178_2 darshan_DXT_POSIX

## `DXT_MPIIO`:
snyder_acme exe_id1253318_9-27-24239-1515303144625770178_2 darshan_DXT_MPIIO

## Some considerations:

* We can now access the `nprocs x xbins` dataframe, OR the original
  dataframe which will be useful for comparisons to the runtime heatmap

* This does increase the test suite runtime (125.2 -> 139.7 s on my Mac).
If I had to guess, `df.update()` is the culprit, especially for cases like
`snyder_acme.exe_id1253318_9..._2.darshan` where there are 1000's of ranks with no data.

* I still have concerns that probably deserve their own issue. Namely, for both `DXT_POSIX` and `DXT_MPIIO` modules in `snyder_acme.exe_..._2.darshan`, we get vertical histograms that don't 100% agree with the heatmap. As you can see above, there are several instances where there is a non-zero value for the vertical histogram but no colored tile in the heatmap (in the 500-1000s range). I checked and the values are certainly in the dataframe, and for 7 ranks there are values of ~5e6, so these should show up orange-red.

nawtrey:
CI failed while installing dependencies. I'll close/reopen to retrigger..

carns:
This is great!

tylerjereddy:
maybe finish up gh-626, then deal with the merge conflicts in this and your other PRs before I start reviewing them

nawtrey:
Alright, I rebase/force pushed here. I combed through `test_plot_dxt_heatmap.py` pretty thoroughly since it had to be merged manually, but I'm pretty confident the current changes are what I originally intended.

tylerjereddy (commented):
Let's start off reviewing this by reading the docstring for `get_heatmap_df()` (emphasis mine). 

> Builds an array similar to a 2D-histogram, where the y data is the <ins>unique ranks</ins> and the x data is time...

What does it return? 

>    hmap_df: dataframe with time intervals for columns and rank
>    index (0, 1, etc.) for rows, where each element contains the data
>    read/written by the corresponding rank in the given time interval.

Does it say anywhere that the data is sparse or filtered to only include active rows? No.
So where is the bug? In `get_heatmap_df()`. Since it is trivial to filter out all-zero rows, I don't really see it as a useful feature to have the function defy its docstring and return a "sparse" representation.

Ok, the PR also describes some performance issues. Let's see if `asv` can find them, rather than relying on anecdotal reports.

`asv continuous pydarshan-devel nawtrey_issue_575_update_ymax -b "time_plot_heatmap_builtin_logs"`

```
       before           after         ratio
     [ed46e931]       [b0c75970]
     <pydarshan-devel>       <nawtrey_issue_575_update_ymax>
+      1.11±0.02s       1.78±0.01s     1.60  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('tests/input/sample-dxt-simple.darshan', 1000)
+      1.19±0.02s       1.81±0.02s     1.51  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('examples/example-logs/ior_hdf5_example.darshan', 1000)
+      1.23±0.04s       1.80±0.03s     1.47  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('examples/example-logs/dxt.darshan', 1000)
+        430±10ms          549±6ms     1.28  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('tests/input/sample-dxt-simple.darshan', 100)
+         316±9ms          372±3ms     1.18  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('tests/input/sample-dxt-simple.darshan', 10)
+         480±6ms         535±10ms     1.11  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('examples/example-logs/ior_hdf5_example.darshan', 100)

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
PERFORMANCE DECREASED.
```

Why not report the statistically relevant results from our standard benchmark suite, which was specifically designed to track heatmap performance?

Ok, what else is weird here? Well, we have a fundamental data structure problem, and not a single test has changed in `test_heatmap_handling.py`. Does it make sense that if there is a fundamental data structure problem, the only way to detect it is to run the *plotting*-focused tests in `test_plot_dxt_heatmap.py` and not the data structure tests? Likely not.

Ok, so now let's see if it was worth vendoring an entire new function, `get_filled_hmap_df()`, that effectively implements a slow and untested version of `pandas.DataFrame.reindex()`:

Compared to the `pydarshan-devel`, the performance hit is substantially lower when I fix the problem right at the source function using a pandas built-in solution (see diff below):

`asv continuous pydarshan-devel treddy_pr622_cleanup -b "time_plot_heatmap_builtin_logs"`

```
       before           after         ratio
     [ed46e931]       [66a86bbe]
     <pydarshan-devel>       <treddy_pr622_cleanup>
+        326±20ms          370±3ms     1.13  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('tests/input/sample-dxt-simple.darshan', 10)
+        441±20ms          494±6ms     1.12  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('tests/input/sample-dxt-simple.darshan', 100)

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
PERFORMANCE DECREASED.

```

If we compare the changes in the diff below directly with this PR feature branch:

`asv continuous nawtrey_issue_575_update_ymax treddy_pr622_cleanup -b "time_plot_heatmap_builtin_logs"`

```
       before           after         ratio
     [b0c75970]       [66a86bbe]
     <nawtrey_issue_575_update_ymax>       <treddy_pr622_cleanup>
-        540±20ms          481±8ms     0.89  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('examples/example-logs/ior_hdf5_example.darshan', 100)
-      1.86±0.04s       1.22±0.03s     0.66  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('examples/example-logs/ior_hdf5_example.darshan', 1000)
-      1.88±0.03s       1.23±0.03s     0.65  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('examples/example-logs/dxt.darshan', 1000)
-      1.83±0.05s       1.19±0.04s     0.65  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('tests/input/sample-dxt-simple.darshan', 1000)

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
PERFORMANCE INCREASED.
```

So, to summarize, by reading the docstring of main dxt dataframe processing function and carefully checking the `pandas` API we can:
- avoid maintaining a slow version of a functionality that `pandas` has already tested and optimized
- have our data structure tests actually sensitive to issues in the dxt heatmap data structure
- not misrepresent the docstring of `get_heatmap_df()`

```diff
diff --git a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
index 157be6f5..dc8ee89a 100644
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -273,7 +273,7 @@ def get_aggregate_data(
     return agg_df
 
 
-def get_heatmap_df(agg_df: pd.DataFrame, xbins: int) -> pd.DataFrame:
+def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFrame:
     """
     Builds an array similar to a 2D-histogram, where the y data is the unique
     ranks and the x data is time. Each bin is populated with the data sum
@@ -379,33 +379,6 @@ def get_heatmap_df(agg_df: pd.DataFrame, xbins: int) -> pd.DataFrame:
     cats = cats.mul(agg_df["length"], axis=0)
     cats.index = agg_df["rank"]
     hmap_df = cats.groupby("rank").sum()
+    hmap_df = hmap_df.reindex(index=range(nprocs),
+                              fill_value=0.0)
     return hmap_df
-
-
-def get_filled_hmap_df(hmap_df: Any, nprocs: int) -> Any:
-    """
-    Takes a dataframe containing a subset of ranks and injects them into
-    a new dataframe with a row for each rank.
-
-    Parameters
-    ----------
-
-    hmap_df: dataframe with time intervals for columns and rank
-    index (0, 1, etc.) for rows, where each element contains the data
-    read/written by the corresponding rank in the given time interval.
-
-    nprocs: the number of MPI ranks/processes used at runtime.
-
-    Returns
-    -------
-
-    filled_df: dataframe of shape ``nprocs x xbins`` with same intervals as
-    `hmap_df`, all data from `hmap_df`, but rows that span from 0 to `nprocs-1`.
-
-    """
-    # make an empty dataframe of size `nprocs x xbins`
-    # and fill it with the data from above dataframe
-    filled_df = pd.DataFrame(np.zeros((nprocs, hmap_df.shape[1])), columns=hmap_df.columns)
-    filled_df.index.name = "rank"
-    filled_df.update(hmap_df)
-    return filled_df
diff --git a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
index 9407f8ff..71c77dfd 100644
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
@@ -335,10 +335,8 @@ def plot_heatmap(
     agg_df = heatmap_handling.get_aggregate_data(report=report, mod=mod, ops=ops)
 
     # get the heatmap data array
-    hmap_df = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=xbins)
-
     nprocs = report.metadata["job"]["nprocs"]
-    hmap_df = heatmap_handling.get_filled_hmap_df(hmap_df=hmap_df, nprocs=nprocs)
+    hmap_df = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=xbins, nprocs=nprocs)
 
     # reverse order of rows so the ranks are in ascending order when plotted
     hmap_df = hmap_df.loc[::-1]
diff --git a/darshan-util/pydarshan/darshan/tests/test_heatmap_handling.py b/darshan-util/pydarshan/darshan/tests/test_heatmap_handling.py
index 85f0e1eb..0bbc8a46 100644
--- a/darshan-util/pydarshan/darshan/tests/test_heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/tests/test_heatmap_handling.py
@@ -332,19 +332,24 @@ def test_get_aggregate_data(log_file, expected_agg_data, mod, ops):
             "sample-dxt-simple.darshan",
             1,
             ["read", "write"],
-            np.array([[4040]]),
+            np.array([[4040, 0, 0, 0, 0, 0, 0, 0, 0,
+                       0, 0, 0, 0, 0 , 0 ,0]]).reshape(16, 1),
         ),
         (
             "sample-dxt-simple.darshan",
             4,
             ["read", "write"],
+            np.vstack((
             np.array([[0, 0, 0, 4040]]),
+            np.zeros((15, 4)))),
         ),
         (
             "sample-dxt-simple.darshan",
             10,
             ["read", "write"],
+            np.vstack((
             np.array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 4040]]),
+            np.zeros((15, 10)))),
         ),
         # `dxt.darshan` is complex enough to warrant changing the
         # selected operations
@@ -555,17 +560,18 @@ def test_get_heatmap_df(
     # generate the report and use it to obtain the aggregated data
     filepath = get_log_path(filepath)
     report = darshan.DarshanReport(filepath)
+    nprocs = report.metadata["job"]["nprocs"]
     agg_df = heatmap_handling.get_aggregate_data(
         report=report, mod="DXT_POSIX", ops=ops
     )
     # run the aggregated data through the heatmap data code
-    actual_hmap_data = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=xbins)
+    actual_hmap_data = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=xbins, nprocs=nprocs)
 
     if "sample-dxt-simple.darshan" in filepath:
         # check the data is conserved
         assert actual_hmap_data.values.sum() == 4040
         # make sure the output array is the correct shape
-        assert actual_hmap_data.shape == (1, xbins)
+        assert actual_hmap_data.shape == (16, xbins)
         # make sure the output data contains identical values
         assert_allclose(actual_hmap_data.values, expected_hmap_data)
 
diff --git a/darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py b/darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
index 04691c1e..7d26fa1c 100644
--- a/darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
@@ -269,8 +269,7 @@ def test_set_y_axis_ticks_and_labels(
     nprocs = report.metadata["job"]["nprocs"]
 
     # generate the heatmap data
-    data = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=xbins)
-    data = heatmap_handling.get_filled_hmap_df(hmap_df=data, nprocs=nprocs)
+    data = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=xbins, nprocs=nprocs)
 
     # generate a joint plot object, then add the heatmap to it
     jointgrid = sns.jointplot(kind="hist", bins=[xbins, nprocs])
```

It is also worth pointing out that this PR describes some other "potential issues--" I'd encourage you to open those as concisely written separate issues with minimal reproducers so I can focus on what is going on here as opposed to the N other bugs that might be present (don't wait for this to be merged, better to reduce my reading overhead here).

nawtrey:
Thanks for the review! I have some questions/comments:

> So where is the bug? In `get_heatmap_df()`. Since it is trivial to filter out all-zero rows, I don't really see it as a useful feature to have the function defy its docstring and return a "sparse" representation.

You make a good point, it should be clear in the dosctring that `hmap_df` contains only the active rank data.

I'm also confused by the wording here -- are you saying we _should_ return a sparse dataframe so it is consistent with the documentation? We only need the sparse representation for plotting, and the current `hmap_df` is comparable to what we should get from the runtime-heatmap module, so I would almost lean towards updating the docstring to say it only contains the active ranks (and reindexing in `plot_heatmap()`). But either way is fine with me.

>I'd encourage you to open those as concisely written separate issues with minimal reproducers so I can focus on what is going on here 

I opened gh-630 to address the issue I noted above.

tylerjereddy:
> I'm also confused by the wording here -- are you saying we should return a sparse dataframe so it is consistent with the documentation? 

No, and I'm not sure how there can be confusion when I pasted the full diff of my intended changes, which already pass the unit tests.

> You make a good point, it should be clear in the dosctring that hmap_df contains only the active rank data.

No, it should return the dense data--why else would I make it dense in my diff if I didn't want that?

> We only need the sparse representation for plotting

Then why are you making it dense before plotting in this PR? Please don't add changes that are not needed. If you don't need the zeros, don't add them in the PR. If you *do* need the zeros present explicitly, then all of my above arguments seem fine.

> reindexing in plot_heatmap()

Again, the full data structure needed for plotting should be dealt with in *tested* heatmap handling code before it gets fed to the plotting routines. There's no need to start stacking a bunch of post-processing functions on the plotting side when we have excellent tools to return the appropriate data structures (and test their structure) on the backend.

> the current hmap_df is comparable to what we should get from the runtime-heatmap module,

I don't see what that has to do with anything--it still isn't a reason to add an untested processing function in the plot code instead of on the tested backend data structure code. The runtime heatmap data will probably have its own backend handling function that is thoroughly tested, and we should strive to have the plotting code *plot* and the data handling backend do as much processing as it can before hand-off.

tylerjereddy:
> I think we are using different definitions of "sparse". When I refer to a matrix as "sparse" I'm using the following defintion

> a sparse matrix or sparse array is a matrix in which most of the elements are zero.

By that definition, if 9/10 ranks are fully active the whole time, and 1 rank is completely inactive, there is no way to produce a sparse representation ("needed" for plotting) because you can't exceed 10 % zeros.

Commonly accepted sparse representations in Python community tend to "hide" zero data by excluding it with speical indexing constructions--see: https://docs.scipy.org/doc/scipy/reference/sparse.html and https://sparse.pydata.org/en/stable/#motivation

> To save space we often <ins>avoid storing</ins> these arrays in traditional dense formats, and instead choose different data structures. Our choice of data structure can significantly affect our storage and computational costs when working with these arrays.

The data becomes dense when you use methods like [`todense`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csc_matrix.todense.html#scipy.sparse.csc_matrix.todense) to transform from implicit sparse representation to explicit dense representation. You can have a dense representation of a sparse matrix--it means to have all the values present. This way you can productively have a conversation about whether an array is "dense" (all rows and columns are present) or sparse (you save memory with various tricks like not explicitly storing the zeros).

nawtrey:
Okay, thanks for the info. That clears things up, so I'll go ahead and fix this up then.

nawtrey:
Okay, I think I've addressed all of the comments here so far. I updated the PR comment to reflect the current changes.

I've also reverted back to using `ax.invert_yaxis()` since the difference we see as a result is really related to gh-630. With those changes reverted the changes here are not enough to remove all of the anomalies described in gh-576 (current figure plotted below), specifically the bullet point about the `DXT_POSIX` horizontal bar graph. But I think that will be addressed by gh-630.

### `snyder_acme.exe_id1253318_9-27-24239-1515303144625770178_2.darshan`(`DXT_POSIX`):
snyder_acme exe_id1253318_9-27-24239-1515303144625770178_2 darshan_DXT_POSIX

tylerjereddy:
It may be helpful to read through some of the pandas User Guide as well. For example, from the sparse tutorial: https://pandas.pydata.org/pandas-docs/stable/user_guide/sparse.html

> pandas provides data structures for efficiently storing sparse data. These are not necessarily sparse in the typical “mostly 0”. Rather, you can view these objects as being “compressed” where any data matching a specific value (NaN / missing value, though any value can be chosen, including 0) is omitted. The compressed values are not actually stored in the array.

Proactively going through content like that may help us communicate more effectively in the future, given that `pandas` is core to `darshan` and understanding the basic data structures `pandas` supports is likely to be quite useful.

This could also lead to a nicer solution long-term, like using a `Sparse[float64, 0.0]` data type, which then specifically encodes that the data is sparse in its type, along with the methods/conversions conveniences that come with it.

darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
```diff
@@ -460,13 +455,13 @@ def test_plot_heatmap(filepath, mod, ops):
...
         assert jgrid.ax_joint.has_data()
-        if "ior_hdf5_example.darshan" in filepath:
-            # verify the horizontal bar graph contains data since there
-            # are multiple ranks for this case
-            assert jgrid.ax_marg_y.has_data()
-        else:
-            # verify the horizontal bar graph does not contain data
+        if "dxt.darshan" in filepath:
+            # verify the horizontal bar graph does not contain data since there
+            # are is only 1 rank for this case
```
tylerjereddy, line 460: May need some proofreading on the comments here I think.

I think I see what is happening with the test changes here, though it is a bit hard to read because file names have changed and there is no parametrization association, so two files can be associated with 1 value that got flipped, which does make it harder for me to read.

More concerning is this change locally for multi-core testing:

- `pydarshan-devel`: `pytest --pyargs darshan -n 6`
  - `298 passed in 92.99s (0:01:32)`
- this branch: `pytest --pyargs darshan -n 6`
  - I gave up after 5.5 minutes

Single core is a bit more reasonable. Nonetheless, it makes me a bit hesitant to move forward here just yet. I may have to at least check if using a genuine sparse data structure and taking advantage of `seaborn.heatmap` feature of `Cells with missing values are automatically masked` would do the same thing faster/with less sustained memory usage.

nawtrey: Okay, I'll go through the comments again and push in more revisions. 

I've been having issues running `pytest --pyargs darshan -n 6` as well, but it seems to work fine with 4 cores, or if I run without using `xdist`:
- `pytest --pyargs darshan -n 4`: `297 passed in 102.48s (0:01:42)`
- `pytest --pyargs darshan`: `297 passed in 134.63s (0:02:14)`

Also I found that after running `pytest --pyargs darshan -n 6 --cache-clear` once I can now run `pytest --pyargs darshan -n 6` fine.

tylerjereddy: Right, a single core is probably using way less memory; if you run on N cores and generate N large reports (incl. DXT figs) simultaneously, there may be some issues, so that's where I have pause.

nawtrey: I tried converting the heatmap dataframe to `SparseDtype` before the reindex:
```diff
diff --git a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
index a917242..e79bed0 100644
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -381,5 +381,6 @@ def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFram
     cats = cats.mul(agg_df["length"], axis=0)
     cats.index = agg_df["rank"]
     hmap_df = cats.groupby("rank").sum()
+    hmap_df = hmap_df.astype(pd.SparseDtype("float", fill_value=0.0))
     hmap_df = hmap_df.reindex(index=range(nprocs), fill_value=0.0)
     return hmap_df
```

Printing out the memory usage with `df.info(memory_usage="deep")`, this change appears to reduce the memory usage:
```
FILE: snyder_acme.exe_id1253318_9-27-24239-1515303144625770178_2.darshan
==============================
mod:  DXT_POSIX
hmap_df (original) memory usage: 12.6 MB
hmap_df (SparseDtype) memory usage: 274.4 KB
hmap_df (SparseDtype, reindexed) memory usage: 210.5 KB
==============================
mod:  DXT_MPIIO
hmap_df (original) memory usage: 48.7 KB
hmap_df (SparseDtype) memory usage: 28.9 KB
hmap_df (SparseDtype, reindexed) memory usage: 28.8 KB
```

The tests still pass with this change (`pytest --pyargs darshan -n 6`: `298 passed in 102.51s (0:01:42)`), and the heatmaps look fine. 

The performance isn't better, of course, but maybe that's tolerable if the memory is of greater concern:

`asv continuous pydarshan-devel nawtrey_issue_575_update_ymax -b "time_plot_heatmap_builtin_logs"`
```
       before           after         ratio
     [536cdce1]       [cba5828d]
     <pydarshan-devel>       <nawtrey_issue_575_update_ymax>
+        906±20ms       1.38±0.06s     1.52  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('tests/input/sample-dxt-simple.darshan', 1000)
+      1.01±0.01s       1.42±0.05s     1.41  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('examples/example-logs/dxt.darshan', 1000)
+      1.01±0.02s       1.41±0.03s     1.41  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('examples/example-logs/ior_hdf5_example.darshan', 1000)
+         262±4ms         335±10ms     1.28  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('tests/input/sample-dxt-simple.darshan', 10)
+         363±4ms         459±10ms     1.26  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('tests/input/sample-dxt-simple.darshan', 100)
+        402±10ms         460±20ms     1.14  dxt_heatmap.PlotDXTHeatMapSmall.time_plot_heatmap_builtin_logs('examples/example-logs/ior_hdf5_example.darshan', 100)

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
PERFORMANCE DECREASED.
```

tylerjereddy: Not sure yet--if memory were chosen over performance then `asv` (peak) memory benchmarks may also be in order: https://asv.readthedocs.io/en/stable/benchmarks.html#benchmark-types 

Deeper question may be if sparsity should start with data structures that are lower down, near the CFFI layer. The code is so slow that `asv` benchmarks for larger logs would be painful, on top of still not having access to the logs repo in `asv`...

Many of those sparse slowdowns are still not as bad as the original version of the PR though. Adding the `snyder..` file to the benchmarks with a custom timeout to allow the benchmarks to run for enough replicates may shed some light on it, but would be painful.

tylerjereddy: Of course, sparse data structures take less memory, though it is not immediately obvious to me what happens when `seaborn` encounters the sparse data structure and has to plot it--maybe it does something clever, though it is certainly quite possible that it produces the dense zeros eventually.

nawtrey: > Not sure yet--if memory were chosen over performance then asv (peak) memory benchmarks may also be in order: https://asv.readthedocs.io/en/stable/benchmarks.html#benchmark-types

Yeah, if we want to go this route then that's probably a requirement. If the sparse representation lowers a relative max instead of the global peak then it's probably not worth the trouble. 

> Of course, sparse data structures take less memory, though it is not immediately obvious to me what happens when `seaborn` encounters the sparse data structure and has to plot it--maybe it does something clever, though it is certainly quite possible that it produces the dense zeros eventually.

I tried tracing the path through `seaborn.matrix.heatmap()`, and it appears that it effectively sends `hmap_df.values` directly into `matplotlib.pyplot.pcolormesh()`, which then runs it through `np.asanyarray()`, and finally through [this function](https://github.com/matplotlib/matplotlib/blob/710fce3df95e22701bd68bf6af2c8adbc9d67a79/lib/matplotlib/cbook/__init__.py#L700-L711) (array is `x`) where it uses `np.array()`. I'm not sure exactly what numpy does with sparse arrays when it encounters them, but my understanding is it doesn't recognize that they're sparse. So it's not clear what is happening yet, but I'll try to walk through those steps and see what happens. It might be as simple as retrieving the plot data from one of the plotting classes, but I'll have to see.

tylerjereddy: asarray() will guarantee dense representation yeah

nawtrey: I pushed in a fix for the typo mentioned here.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -273,7 +273,7 @@ def get_aggregate_data(
     return agg_df
 
 
-def get_heatmap_df(agg_df: pd.DataFrame, xbins: int) -> pd.DataFrame:
+def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFrame:
```
tylerjereddy, line 276: when you change a function signature in the main library, it is important to grep the codebase for other places it is used like the `asv` benchmarks, which this breaks..

I'll try to fix this now

tylerjereddy: Ok, I pushed in the fix and rebased as well

PR #632, opened 2022-01-25
BUG: correct section closing tag in `base.html`

nawtrey:
* Relocate section closing tag to outside of figure loop

* Add check to tests `test_main_without_args` and `test_main_all_logs_repo_files` to verify the number of close/open section tags match

### DIFF
--- a/darshan-util/pydarshan/darshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/darshan/tests/test_summary.py
@@ -178,6 +178,10 @@ def test_main_all_logs_repo_files(tmpdir, log_repo_files):
                         if mod in ["POSIX", "MPI-IO", "STDIO"]:
                             assert "I/O Cost" in report_str
 
+                    # check the number of opening section tags
+                    # matches the number of closing section tags
+                    assert report_str.count("<section>") == report_str.count("</section>")
+
 class TestReportData:
 
     @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


nawtrey:
Basically we have unnecessary `</section>` tags after each figure, which throws everything out of wack. This small change closes the sections at the appropriate time. The reports don't look any different as a result, but this solves a problem that will show itself if we move towards tiling figures within each section.

nawtrey:
I'll push in some changes that check the number of closing `/section` tags match the number of sections as a regression guard.

darshan-util/pydarshan/darshan/tests/test_summary.py
```diff
@@ -118,6 +118,10 @@ def test_main_without_args(tmpdir, argv, expected_img_count):
                     # check that expected number of figures are found
                     assert report_str.count("img") == expected_img_count
 
+                    # check the number of opening section tags
+                    # matches the number of closing section tags
+                    assert report_str.count("<section>") == report_str.count("</section>")
```
tylerjereddy, line 123: Maybe add this to `test_main_all_logs_repo_files()` below as well, after rebasing to deal with current merge conflicts.

Since this test now covers at least one log repo file, the need for `test_main_all_logs_repo_files()` may become a point of contention, but for now, that test still remains useful for PRs to the logs repo, since it will scrape in every log to sanity check report generation.

nawtrey: > Maybe add this to `test_main_all_logs_repo_files()` below as well, after rebasing to deal with current merge conflicts.

Thanks, I've rebased and added the same check to `test_main_all_logs_repo_files()`. 

> the need for `test_main_all_logs_repo_files()` may become a point of contention

We could probably remove it and move some test cases to `test_main_without_args` if we think they are important to keep. Then add write a new test that initializes a `DarshanReport` for each log in the repo and maybe perform a few checks on the log data.

tylerjereddy: For now, making sure that every log file in the log repo can successfully be used generate a report file proper makes sense to me.

tylerjereddy:
Ok, I don't see any problems so I'll merge thx

Issue #563, opened 2021-11-10
Remove duplicate logs


nawtrey:
It was discovered that `tests/input/sample.darshan` shares a `jobid` with `example.darshan` and `example2.darshan`. `sample.darshan` also appears in `examples/example-logs`. It seems the consensus is to simply purge the redundant logs from `pydarshan`.

PR #635, opened 2022-01-29
MAINT: purge example2.darshan

tylerjereddy:
Another small part of gh-563

* `example2.darshan` is a duplicate of `example.darshan`, which
itself is a duplicate of `sample.darshan`

* for now, just purge `example2.darshan` as a small cleanup; I
have not touched the two Jupyter notebooks that used `example2.darshan`,
but I suspect Jakob is "ok" with that for now


jakobluettgau:
Sounds good to me too.

I'll create an issue and go over the notebooks (fix paths, doublecheck that dtype is not used etc).

tylerjereddy:
Ok, reviews are positive, CI is passing, and this is fairly trivial, so in it goes.

PR #636, opened 2022-01-29
MAINT: rebase pydarshan-devel on latest main

tylerjereddy:
* rebase `pydarshan-devel` on latest `main` to facilitate incorporation of runtime heatmap Python/CFFI bindings into `pydarshan-devel` (see: https://github.com/darshan-hpc/darshan/pull/615#issuecomment-1022587999)

* probably good for reviewers to scan through the (substantial) diff and make sure it looks "ok" for things from `main` added into `pydarshan-devel`, and that CI is passing of course


tylerjereddy:
hmm, the diff is wrong.. let me check..

tylerjereddy:
Ok, it was easier to switch strategies and merge `main` into `pydarshan-devel`. The diff looks cleaner now, but still merits a skim over by reviewers given the size.

carns (approved):
Looks good to me; I think it covers everything.  I forgot that this was going to include the wrapper performance optimizations too, but that's good to have it all in sync now.

The "merge to feature branch" approach is fine.  There has been more activity on the pydarshan-devel branch than the main branch over this time frame so its probably a safer option in this case at any rate.

tylerjereddy:
ok, I'll merge then so Jakob can chip away at the bindings/simplify the diff there

tylerjereddy:
Devs should note that rebuilding the C library is probably advisable after this update to `pydarshan-devel`, in case that is not obvious..

PR #637, opened 2022-01-31
ENH: Add access size histogram to summary report

nawtrey:
* Add access size histogram figure from
`plot_access_histogram` to summary report

* Update expected `img` tag counts in `test_main_without_args`
to account for new access size histogram figures

* Rework table counting in `test_main_without_args` to ignore
extraneous instances of the word "table", such as the instance
in the figure caption for the access size histogram

* Contributes to issue #465

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -20,6 +20,7 @@ from darshan.experimental.plots import (
     plot_dxt_heatmap,
     plot_io_cost,
     plot_common_access_table,
+    plot_access_histogram,
 )
 
 darshan.enable_experimental()
@@ -393,6 +394,20 @@ class ReportData:
         ## Per-Module Statistics
         ################################
         for mod in self.report.modules:
+            if mod in ["POSIX", "MPI-IO"]:
+                access_hist_description = (
+                    "Read/write operations grouped by access size. Most frequent "
+                    "access sizes are featured in the <i>Common Access Sizes</i> table."
+                )
+                access_hist_fig = ReportFigure(
+                    section_title=f"Per-Module Stats: {mod}",
+                    fig_title="Access Sizes",
+                    fig_func=plot_access_histogram,
+                    fig_args=dict(report=self.report, mod=mod),
+                    fig_description=access_hist_description,
+                    fig_width=350,
+                )
+                self.figures.append(access_hist_fig)
             if mod in ["POSIX", "MPI-IO", "H5D"]:
                 if mod == "MPI-IO":
                     com_acc_tbl_description = (
--- a/darshan-util/pydarshan/darshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/darshan/tests/test_summary.py
@@ -74,9 +74,9 @@ def test_main_with_args(tmpdir, argv):
     "argv, expected_img_count, expected_table_count", [
         (["noposix.darshan"], 1, 2),
         (["noposix.darshan", "--output=test.html"], 1, 2),
-        (["sample-dxt-simple.darshan"], 3, 4),
-        (["sample-dxt-simple.darshan", "--output=test.html"], 3, 4),
-        (["nonmpi_partial_modules.darshan"], 2, 3),
+        (["sample-dxt-simple.darshan"], 5, 4),
+        (["sample-dxt-simple.darshan", "--output=test.html"], 5, 4),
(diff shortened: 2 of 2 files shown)


nawtrey:
Adding the figure as-is despite some prevailing cosmetic issues (see gh-577).

Some example reports generated on this branch: [report_examples.tar.gz](https://github.com/darshan-hpc/darshan/files/7974617/report_examples.tar.gz)

Maybe @carns can comment on the figure caption, I just figured it was worth mentioning the connection between this figure and the common access sizes table.

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -393,6 +394,20 @@ def register_figures(self):
         ## Per-Module Statistics
         ################################
         for mod in self.report.modules:
+            if mod in ["POSIX", "MPI-IO"]:
```
tylerjereddy, line 397: Why does the access table include modules that the plots do not? For example, `ior_hdf5_example.darshan` is missing an access plot for `H5D`. Well, I can think of a reason, but I don't see it mentioned here or in the matching issue, so that seems like something obvious to clarify, and what you plan to do about it. Probably want to check with the team about this module, since the old report seems to include neither the table nor the plot for it.

image

paralepsis:
The old report predates the HDF module. — Rob

darshan-util/pydarshan/darshan/cli/summary.py
nawtrey: Thanks, that's a good question. And thanks for the comment @paralepsis 

I was going off the "supported_mods" listed in `experimental/aggregators/mod_agg_iohist.py`, but it looks like we can add `H5D` without much effort:

```diff
diff --git a/darshan-util/pydarshan/darshan/cli/summary.py b/darshan-util/pydarshan/darshan/cli/summary.py
index f7da02d..32c4fae 100644
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -394,7 +394,7 @@ class ReportData:
         ## Per-Module Statistics
         ################################
         for mod in self.report.modules:
-            if mod in ["POSIX", "MPI-IO"]:
+            if mod in ["POSIX", "MPI-IO", "H5D"]:
                 access_hist_description = (
                     "Read/write operations grouped by access size. Most frequent "
                     "access sizes are featured in the <i>Common Access Sizes</i> table."
@@ -408,7 +408,6 @@ class ReportData:
                     fig_width=350,
                 )
                 self.figures.append(access_hist_fig)
-            if mod in ["POSIX", "MPI-IO", "H5D"]:
                 if mod == "MPI-IO":
                     com_acc_tbl_description = (
                         "NOTE: MPI-IO accesses are given in "
diff --git a/darshan-util/pydarshan/darshan/experimental/aggregators/mod_agg_iohist.py b/darshan-util/pydarshan/darshan/experimental/aggregators/mod_agg_iohist.py
index d9bb9ee..e9fcf35 100644
--- a/darshan-util/pydarshan/darshan/experimental/aggregators/mod_agg_iohist.py
+++ b/darshan-util/pydarshan/darshan/experimental/aggregators/mod_agg_iohist.py
@@ -12,7 +12,7 @@ def mod_agg_iohist(self, mod, mode='append'):
     """
 
     # sanitation and guards
-    supported = ["POSIX", "MPI-IO"]
+    supported = ["POSIX", "MPI-IO", "H5D"]
     if mod not in supported:
         raise Exception("Unsupported mod_name for aggregated iohist.")
``` 


And it seems to generate the figure just fine:


Screen Shot 2022-02-01 at 1 05 01 PM

For the "Common Access Sizes" table, when I was figuring out how to retrieve the module data for `POSIX` and `MPI-IO`, I found that `H5D` didn't require any extra work to support. So that's why it has some extra module coverage at the moment. 

As for a solution, based on the diff above I think we can add support for `H5D` fairly easily. We will have to increase the scope of the current tests of course. I could open a separate PR for that.

tylerjereddy:
Ok, maybe open an issue to add support/tests for `H5D` in the plots. That's mostly what I'm looking for--an issue with an explanation for the absence of the plot and a plan to deal with it so I can move forward with review.

nawtrey:
Alright I created gh-639 to address that issue.

tylerjereddy:
Ok merged -- I already cross-checked with a bunch of the old perl reports last night.

Issue #517, opened 2021-10-06
ENH: add layout controls to darshan summary report  


nawtrey:
In the last few meetings we've discussed giving users the option to easily configure their summary report by selecting which plots they want and what sections they want to place them in, but it would be nice to also have the ability to place figures adjacent to each other as well (instead of vertically stacking them). 

The real advantage to this is the `pydarshan` plotting functions would only ever need to generate a single `matplotlib` figure, we wouldn't need to create figures with multiple subfigures because the HTML would be handling the formatting/object placement on its own.

This is lower priority than any of the current pydarshan-job-summary-v1 milestone objectives, but I'm opening an issue now so we can track it moving forward.

shanedsnyder:
I think this is done, as well. We can re-open if we have more specific wants.

PR #645, opened 2022-02-03
ENH: Add grid formatting to darshan summary report

nawtrey:
* Add and implement new CSS class to allow for grid
formatting of figures in each summary report section

* Contributes to issue #517

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/base.html
+++ b/darshan-util/pydarshan/darshan/cli/base.html
@@ -32,7 +32,16 @@
       % for sect_title in report_data.sections:
           <section>
           <h2>${sect_title}</h2>
+
+          % if len(report_data.sections[sect_title]) > 1:
+            <!-- if more than 1 figure in section, use grid wrapper class -->
+            <div class="grid-wrapper">
+          % else:
+            <div>
+          % endif
+
           % for fig_title, fig in report_data.sections[sect_title].items():
+            <div>
             <h3>${fig_title}</h3>
             <figure style="width: ${fig.fig_width}px;">
             % if fig.fig_html:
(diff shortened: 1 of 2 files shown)


nawtrey:
A first attempt at adding some formatting to the report via `display: grid`. At the moment each section is blocked together, and the elements in that block are grouped into a `N x 2` grid, where `N` is the number of rows (determined by the number of figures). If a section contains only a single figure, it will just be put in a block by itself (no grid). It is also set to auto scale the widths of the figures in a section, so if a figure is much wider than another it will dedicate more space to that figure. This isn't really highlighted in the examples because the "common access sizes" table is deceptively wide, but it's worth mentioning.  

I think we could still improve on this by centering the figure titles, figures, and figure captions in their grid cells, but that's my personal preference. If there is interest in that I can open a separate PR for that. I would also be interested in finding a way to do this without so much white space between figures, but that will require some more digging I think. 

#### Examples
Some example reports highlighting the grid structure: [example_reports.tar.gz](https://github.com/darshan-hpc/darshan/files/7998743/example_reports.tar.gz)

carns:
This is great.  Just a quick observation (this is beside the point of the grid, but it's the kind of thing that a x2 layout helps show), but in acme example the two heatmaps look like they have slightly different y ranges.  Is that expected/known?  I think it's Ok if the x range varies a little but for comparison we ought to be able to lock in to a fixed y range since the number of processes is constant.

tylerjereddy:
> the two heatmaps look like they have slightly different y ranges. Is that expected/known? 

Yes, there is a fairly complicated discussion in gh-622, which in turn aims to deal with gh-576. I still haven't decided which solution I like on the dense/sparse data structure spectrum just yet. I'll try to decide soon..

carns:
> I'll try to decide soon..

Ok, no problem.  Just making sure we had that captured somewhere.

tylerjereddy (approved):
LGTM, thx

Issue #641, opened 2022-02-02
ENH: general feedback on pydarshan summary text


carns:
Hopefully this is legible; it's easier for me to review the report as a whole than offer feedback on individual components at a time:

https://acrobat.adobe.com/link/track?uri=urn:aaid:scds:US:fc9ed5ea-c0fe-48b0-9502-b24208a1742d

The only organizational change is a suggestion to reorg the top two sections slightly so that all of the generic job metadata is in one section and all of the Darshan instrumentation metadata is in another section.  The rest of the edits are just wording suggestions.

The re-write of the caption for the I/O cost figure is an attempt to mitigate (for now) the concern that the average values can be misleading, by referring back to the heatmap which can show imbalance.  We might want to change this again later once we have the tooling to show the slowest process in addition to the average process.

Any of this is open for discussion.

tylerjereddy:
Thanks, we could probably edit your issue message to contain a markdown checklist for the items, then have granular PRs to address the items individually (to make it easier for review of changes).

nawtrey:
Thanks for the feedback! I think this covers Phil's comments:

#### TODO: 
- [ ] "Log Details" -> "Job Summary"
- [ ] Change runtime data type to integer
- [ ] In "Log Details", "Command" -> "Command Line"
- [ ] Break up log metadata/module info such that "Log Filename", "Runtime Library Version" and "Log Format Version" are combined with the "Module Breakdown" metadata, with the module data on top
- [ ] "Module Breakdown" -> "Darshan Log Information"
- [ ] Suffix each module index with "Module Data" (i.e. `POSIX (ver=4)` -> `POSIX (ver=4) Module Data`)
- [ ] Update partial flag message to "Module data incomplete due to runtime memory or record count limit"
- [ ] "I/O Operations" -> "I/O Summary"
- [ ] DXT heatmap caption:
   * "vertical bar graph" -> "top edge bar graph"
   * Remove "all" from "across all ranks"
   * "show the total I/O" -> "aggregate I/O volume"
   * "horizontal bar graph" -> "right edge bar graph"
   * "all I/O events for each rank to illustrate how th eI/O was distributed across ranks" -> "each rank across time slices t show I/O distribution across ranks"
- [ ] Update I/O cost caption to "Average (across all ranks) amount of run time that each process spent performing I/O, broken down by access type. See the right edge bar graph on heat maps in preceding section to indicate if I/O activity was balanced across processes."
[...]

PR #647, opened 2022-02-04
ENH: Update summary report figure captions

nawtrey:
* Update captions for DXT heat map, I/O cost, and Access Size
figures according to suggestions from Phil

* Contributes to issue #641

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -330,10 +330,10 @@ class ReportData:
         hmap_description = (
             "Heat map of I/O (in bytes) over time broken down by MPI rank. "
             "Bins are populated based on the number of bytes read/written in "
-            "the given time interval. The vertical bar graph sums each time "
-            "slice across all ranks to show the total I/O over time, while the "
-            "horizontal bar graph sums all I/O events for each rank to "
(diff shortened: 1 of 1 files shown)


nawtrey:
Here is an example report using `ior_hdf5_example.darshan`: [ior_report.tar.gz](https://github.com/darshan-hpc/darshan/files/7999124/ior_report.tar.gz)

carns:
Slight correction from looking at the example: "broken down by process type" should probably be "broken down by access type" in the I/O Cost graph caption.

nawtrey:
> Slight correction from looking at the example: "broken down by process type" should probably be "broken down by access type" in the I/O Cost graph caption.

Thanks, I'll fix this real quick

tylerjereddy:
One point of caution--check the supported Python versions for the latest `pytest` release--we are bound to supporting some pretty old versions so that is something to consider before making changes, or to carefully guard changes based on Python version.

tylerjereddy:
That said, test-time dependencies should be fair game to have a latest stable release--it is not really the same kind of burden as a bleeding-edge runtime dep.

tylerjereddy (approved):
Ok, I did a spot check of a report generated locally and Phil seems happy with the changes.

I checked off the corresponding items in the issue markdown list.

tylerjereddy:
thanks

PR #654, opened 2022-02-11
ENH: Add operation counts figure to summary report

nawtrey:
* Add operation counts figure from
`plot_opcounts.py` to summary report

* Update expected `img` tag counts in `test_main_without_args`
to account for new operation count figures

* Fix `img` tag string such that extraneous instances
of "img" are not included in total count

* Close issue #465


nawtrey:
I've spot checked the old report "I/O Operation Counts" figure against the reports below, and all of the values seem consistent. It is also a bit different, since the names have been changed and the figure has been split up into several figures, but they are easy to compare. 

Example reports: [example_reports.tar.gz](https://github.com/darshan-hpc/darshan/files/8045249/example_reports.tar.gz)

nawtrey:
@carns Sounds good, I'll fix that up today.

nawtrey:
I've updated the figure caption and update the example reports to reflect the new changes.

carns:
looks great, thanks @nawtrey

tylerjereddy (approved):
LGTM

When I spot checked `treddy_mpi-io-test_id4373053_6-2-60198-9815401321915095332_1.darshan` it seemed a bit jarring that `STDIO` had an opcount plot but no access size plot, while other modules tend to have both.

The old report is consistent with that, so I can't really fault it. Maybe related to avoiding duplication or limitations in the underlying data.

carns:
That's correct that the stdio module doesn't produce this information.

It's a lower priority there since all of the data that passes through the stdio interface is explicitly buffered, and there isn't much information available about how/when it maps to underlying storage system or storage service access.

PR #659, opened 2022-02-16
MAINT: Update darshan summary report metadata table runtime datatype

nawtrey:
* Change method `ReportData.get_runtime()`
to use integers instead of floats

* Correct tests `test_metadata_table` and
`test_get_runtime` to use integer values

* Fixes task "Change runtime data type to integer" in gh-641

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -174,11 +174,11 @@ class ReportData:
 
         """
         # calculate the run time
-        runtime_val = float(
+        runtime_val = int(
             report.metadata["job"]["end_time"] - report.metadata["job"]["start_time"]
         )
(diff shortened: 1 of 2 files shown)


tylerjereddy (approved):
Ok, I spot checked a report locally and looks good.

I'm perhaps a bit less convinced that this is a useful long-term change since Shane already mentioned potential buy-in for using higher-resolution timers in the future, and microsecond resolution would be pretty normal these days I'd say.

Nonetheless, this is what Phil asked for and that is the current resolution we have I suppose.

tylerjereddy:
I checked off the TODO item

carns:
Thanks guys. Ideally the report would simply show as much decimal precision as the values in the log (in case as @tylerjereddy we have more resolution later).  It just bugs me to show .0 when we don't actually have that precision.

PR #660, opened 2022-02-16
MAINT: Update partial flag message

nawtrey:
* Change partial flag message to "Module data incomplete
due to runtime memory or record count limits"

* Correct test `test_module_table` to reflect
updated partial data flag message

* Fixes task "Update partial flag message" in gh-641


nawtrey:
Here is an example report: [example_report.tar.gz](https://github.com/darshan-hpc/darshan/files/8081601/example_report.tar.gz)

The updated flag from the above example: 
Screen Shot 2022-02-16 at 9 31 12 AM

tylerjereddy (approved):
LGTM, local spot check was "ok" as well. I checked off the TODO list item.

Issue #656, opened 2022-02-14
pydarshan can't load libdarshan-util-*.so from pypi on Cori&Perlmutter (NERSC/LBL)


bebosudo:
Installing pydarshan from pypi on [Cori or Perlmutter](https://docs.nersc.gov/systems/) causes pydarshan to fail at import time:

```console
[chiusole@cori10:cori]$ python3 -c 'import darshan'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/global/homes/c/chiusole/.local/lib/python3.6/site-packages/darshan/__init__.py", line 19, in <module>
    from darshan.report import DarshanReport
  File "/global/homes/c/chiusole/.local/lib/python3.6/site-packages/darshan/report.py", line 10, in <module>
    import darshan.backend.cffi_backend as backend
  File "/global/homes/c/chiusole/.local/lib/python3.6/site-packages/darshan/backend/cffi_backend.py", line 47, in <module>
    libdutil = find_utils(ffi, libdutil)
  File "/global/homes/c/chiusole/.local/lib/python3.6/site-packages/darshan/discover_darshan.py", line 216, in find_utils
    raise RuntimeError('Could not find libdarshan-util.so! Is darshan-util installed? Please ensure one of the the following: 1) export LD_LIBRARY_PATH=<path-to-libdarshan-util.so>, or 2) darshan-parser can found using the PATH variable, or 3) pkg-config can resolve pkg-config --path darshan-util, or 4) install a wheel that includes darshan-utils via pip.')
RuntimeError: Could not find libdarshan-util.so! Is darshan-util installed? Please ensure one of the the following: 1) export LD_LIBRARY_PATH=<path-to-libdarshan-util.so>, or 2) darshan-parser can found using the PATH variable, or 3) pkg-config can resolve pkg-config --path darshan-util, or 4) install a wheel that includes darshan-utils via pip.
```

This can be solved by manually providing the path to the libdarshan-util-*.so and libz-*.so* library in `LD_LIBRARY_PATH`:

```console
[chiusole@cori10:cori]$ LD_LIBRARY_PATH="$HOME/.local/lib/python3.6/site-packages/darshan.libs/:$LD_LIBRARY_PATH" python3 -c 'import darshan'; echo $?
0
```

The problem seems to be around this block:

[...]

PR #665, opened 2022-02-17
Update PyDarshan wheel building process to manylinux2014 (fix for #656)

jakobluettgau:
Fix build pipeline to use manylinux2014 to resolve pydarshan being unable to load libdarshan-util due to failing to find secondary libz dependency on dlopen (see  #656).


darshan-util/pydarshan/darshan/__init__.py
```diff
@@ -3,7 +3,7 @@
 PyDarshan also provides a suite of analysis utilities.
 """
 
-__version__ = '3.3.1.0'
+__version__ = '3.3.1.1'
```
tylerjereddy, line 6: I think, because you are changing the source proper in i.e., `discover_darshan.py`, it may indeed be the right choice to release a newer version. I guess you and Shane have a convention for what the different numbers/positions mean. If we weren't changing the source at all, and just rebuilding wheels, then only incrementing the build number on the wheels with `-1` might suffice, but probably not in this case.

One thing you might try, to save time, is upload a few test wheels to https://test.pypi.org/ first and ask the HPC center folks to point their `pip` installs at that location (`-i URL` flag I think?) before doing the actual release.

tylerjereddy (commented):
I cherry-picked the two commits in this PR into `pydarshan-devel` and the testsuite ran just fine as expected.

I suspect it is mostly you and Shane that will discuss this stuff related to the release process--nothing looks too unusual to me, and `manylinux2014` seems like a step in the right direction.

darshan-util/pydarshan/darshan/discover_darshan.py
```diff
@@ -201,7 +201,7 @@ def find_utils(ffi, libdutil):
         try:
             darshan_path = discover_darshan_pyinstaller()
             import glob
-            library_path = glob.glob(f'{darshan_path}/libdarshan-util*.so')[0]
+            library_path = glob.glob(f'{darshan_path}/libdarshan-util*.so*')[0]
```
shanedsnyder, line 204: Not immediately clear what these changes to the glob here are trying to do. Could you provide some more details just so I better understand what they are fixing?

darshan-util/pydarshan/darshan/__init__.py
shanedsnyder, line 6: I just scanned our release checklist to remind myself if anything was forgotten here, and the only thing probably worth updating as part of this PR is the pydarshan changelog? Maybe indicate what's changed in our wheel support and what what was fixed in Darshan log discovery?

I'll run tests and also double check with NERSC that this resolves issues there (probably using test release on PyPI that @tylerjereddy mentions).

darshan-util/pydarshan/darshan/discover_darshan.py
jakobluettgau: The names of the shared object files changed slightly, they are now postfixed with .0.0.0. Without the wildcard for the glob pattern an empty list would be returned and the lib-darshan util discovery fails for this variant.

darshan-util/pydarshan/darshan/__init__.py
jakobluettgau: NERSC confirmed that the new wheels work for them in https://github.com/darshan-hpc/darshan/issues/656#issuecomment-1045502017

PR #668, opened 2022-02-18
gracefully parse logs with unknown module ids

### LINKED ISSUE #621: segfault in darshan-parser if old darshan-utils encounters a new module
reported by @jakobluettgau

This can be reproduced by running darshan-parser from the most recent Darshan release on a log file that contains heatmap data.

In theory this should be fine; the parser is meant to ignore unknown module types.  It looks like the problem is in a for loop that accesses an array of module names.

carns:
- bug reported by Jakob Luettgau
- prevents memory corruption when attempting to parse logs that have compatible log format but contain unknown log ids

Fixes #621

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,7 +2,7 @@
 Darshan Release Change Log
 --------------------------
 
-Darshan-3.3.2
+Darshan-3.4
 =============
 * Add a new "heatmap" module that records histograms of I/O traffic over time
   on each process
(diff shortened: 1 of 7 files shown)


carns:
I meant to give an example of the output.  If you parse a log with an unknown module id (for example, one that was introduced in a later darshan release but did not otherwise perturb the log format), then you will see this in the module table:

```
# <UNKNOWN> module (id 14): 275 bytes (compressed), ver=1
```
... and this when the parser gets to the point that it would normally dump the records for that module:
```
# Warning: module id 14 is unknown. You may need a newer version of the Darshan utilities to parse it.
```

shanedsnyder (approved):

PR #671, opened 2022-02-20
TST: xfail HEATMAP tests

tylerjereddy:
* mark the two test cases requiring `HEATMAP` module
support as known failures (`xfail`) for now, so that
PRs in the main and logs repos can pass CI until the
support is added in

* because the two tests are not parametrized, the
`xfail`s are imperative, which means that only
those inputs that precede the `xfail`ed logs
will run for now, so I wouldn't want to leave
these in for long, just long enough to avoid
rushing the HEATMAP support addition/review

* an alternative is to simply `pass`/`continue`
for these test cases, which would allow all
of the other cases for the two tests to run
as usual, though it would leave out the
convenient `pytest` reminder that the tests
have known failures in them

### DIFF
--- a/darshan-util/pydarshan/darshan/tests/test_report.py
+++ b/darshan-util/pydarshan/darshan/tests/test_report.py
@@ -35,6 +35,8 @@ def test_jobid_type_all_logs_repo_files(log_repo_files):
     # this is primarily intended as a demonstration of looping
     # through all logs repo files in a test
     for log_filepath in log_repo_files:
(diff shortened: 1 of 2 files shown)


tylerjereddy:
I'll clean this up a bit more to skip less tests now b/c of Nik's comment in the cross-linked PR, but I note that my intentions here were pretty clear and we explicitly opted out of using `pass`/`continue` temporarily..

PR #673, opened 2022-02-21
normalize heatmap histograms

### LINKED ISSUE #672: normalize heatmap bin_width and nbins across ranks and heatmap types at run time
Right now each heatmap record produced by Darshan could have a different number of bins and/or bin width depending on when I/O occurred for that particular I/O method on a given rank.

I think the analysis APIs could normalize this for plotting purposes, but it looks like it could be done at runtime without much trouble and probably save a lot of confusion.  The number of bins and width will still be variable, but the values can match across all heatmap records.

As far as implementation, I think the solution is to implement the `mod_redux_func()` hook in darshan-heatmap.c and use it to perform a reduction operation and have all ranks agree on the stop time for heatmap measurements.  The existing `mod_output_func()` function already iterates through all records and prunes bins that extend beyond execution end time.  It just needs to be augmented to a) also call `collapse_heatmap()` to extend the bin width to include execution end time and b) use the agreed-upon collective end time if present.

If all heatmaps use the same end time for this calculation then they will all produce identical bin widths and nbins without further coordination.

This will not change the log format; we will still record width and bin count per heatmap record.

@shanedsnyder @jakobluettgau @nawtrey @tylerjereddy tagging you because this might change what needs to be done on the analysis side to parse the heatmaps.  We probably also want to regenerate the example logs so that they produce data as described here.

carns:
- all histograms emitted in a given job will now have the same bin width and number of bins

Fixes #672

### DIFF
--- a/darshan-runtime/lib/darshan-heatmap.c
+++ b/darshan-runtime/lib/darshan-heatmap.c
@@ -23,6 +23,12 @@
 #include "darshan.h"
 #include "darshan-heatmap.h"
 
+/* If set, this is the globally (across all ranks) agreed-upon timestamp to
+ * use as the end time for normalizing and pruning heatmap bins consistently.
+ * If not set, use locally derived end timestamp during output fn.
+ */
+double g_end_timestamp = 0;
+
 /* maximum number of bins per record */
 /* TODO: make this tunable at runtime */
 /* TODO: safety check that total record size plus trailing bins doesn't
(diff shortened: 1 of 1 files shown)


tylerjereddy:
It should be possible to evaluate/review this on the Python side as well--just use custom stuff like https://github.com/tylerjereddy/heatmap_diagonal, adjusted to produce variable bin counts on different ranks with current `main`, and look at the runtime datastructures produced before and after.

shanedsnyder:
I confirmed that new versions of E3SM-IO logs (in PR https://github.com/darshan-hpc/darshan-logs/pull/28) don't contain different values for `nbins` in the heatmap module:

```
darshan-parser /tmp/snyder_e3sm_io_id586491-96860_3-2-72310-13271552163192606762_1.darshan | grep NBINS | cut -f 5 | uniq
107
```
I hacked darshan-parser to display NBINS value to quickly confirm, you can see that every NBINS value, across STDIO, MPI-IO, and POSIX modules is the same (107).

Also worth noting that I no longer get errors in https://github.com/darshan-hpc/darshan/pull/615 related to extracting heatmap module data -- previous versions of the heatmap module could trigger some assertions there.

PR #675, opened 2022-03-01
TST: graceful HEATMAP xfails

tylerjereddy:
* only test cases that require runtime `HEATMAP`
module support are `xfail`ed in the test suite now,
instead of skipping the cases that also come after
them in iterable tests

* while the original design was meant to be temporary (gh-671),
and was made intentionally on the assumption that HEATMAP
support would be added shortly, this introduces parametrization
into the two tests that are involved and does away with the
old fixtures to avoid debates like the one in https://github.com/darshan-hpc/darshan-logs/pull/27

* on `pydarshan-devel`:
`300 passed, 2 xfailed in 38.83s`

* on this branch:
`318 passed, 6 xfailed in 96.56s (0:01:36)`

* there are 3 runtime heatmap logs, and two tests that
need to avoid them, so the xfail arithmetic looks correct there,
along with the restoration of a bunch more logs cases...

* as always, others should feel free to help streamline the testing/devops infrastructure

### DIFF
--- a/darshan-util/pydarshan/darshan/log_utils.py
+++ b/darshan-util/pydarshan/darshan/log_utils.py
@@ -32,6 +32,12 @@ except ImportError:
     has_log_repo = False
 
 
+def _provide_logs_repo_filepaths():
+    if has_log_repo:
+        return _produce_log_dict("darshan_logs").values()
+    return []
+
+
 @functools.lru_cache(maxsize=4)
 def _produce_log_dict(project):
     p = importlib_resources.files(project) # type: Any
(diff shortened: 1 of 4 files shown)


tylerjereddy:
ok, I'm going to have to make this handle absence of logs repo as well... which is kind of why we were avoiding this..

tylerjereddy:
Ok, CI is passing, should be ready for review

nawtrey (approved):
I double checked the changes, and verified the 2 altered tests are running and xfailing the correct logs. LGTM.

PR #684, opened 2022-03-08
WIP: unify darshan-merge and darshan-parser accumulator logic

carns:
The darshan-merge utility uses a logutils API function (see the `log_agg_records()` fn pointer) to combine records if needed.  This function has a high degree of overlap with functions implemented in darshan-parser (for the posix, mpi-io, and stdio record types) to combine records in `--total` mode.

This PR is intended to unify the two code paths by using the logutils API function in both cases, along the way generalizing it to automatically produce appropriate calculations depending on whether the records being aggregated refer to the same file (i.e., the "shared file" case) or not.  A later PR should then be able to further collapse these code paths so that the same underlying logic is used on runtime reduction operators as well.

Will likely need to disable variance calculation in darshan-merge for now, but that can be revisited later if we decide that we want that capability for reconstructing partial mmap log files.


carns:
This PR also fixes a few bugs in the darshan-parser aggregator logic; the log_agg_record functions used by darshan-merge were more accurate.

darshan-util/darshan-mpiio-logutils.c
```diff
@@ -518,6 +554,22 @@ static void darshan_log_agg_mpiio_files(void *rec, void *agg_rec, int init_flag)
                 }
                 break;
             case MPIIO_F_FASTEST_RANK_TIME:
+
+                if(!shared_file_flag || mpi_rec->base_rec.rank == -1)
```
shanedsnyder, line 558: At first glance, I'm not sure the second part of the if statement is needed here (`mpi_rec->base_rec.rank == -1`). If the `rank == -1` then there are 2 case:
- there are multiple files represent in the input set of logs, which would be handled by the first if condition (`!shared_file_flag`)
- there is a single file represented in the input set of logs, where ideally we could just preserve the FASTEST/SLOWEST counters rather than setting to `-1` here

darshan-util/darshan-mpiio-logutils.c
```diff
@@ -320,16 +320,45 @@ static void darshan_log_agg_mpiio_files(void *rec, void *agg_rec, int init_flag)
...
+
+    /* so far do all of the records reference the same file? */
+    if(agg_mpi_rec->base_rec.id == mpi_rec->base_rec.id)
+        shared_file_flag = 1;
+    else
+        agg_mpi_rec->base_rec.id = 0;
+
+    /* so far do all of the records reference the same rank? */
+    if(agg_mpi_rec->base_rec.rank != mpi_rec->base_rec.rank)
+        agg_mpi_rec->base_rec.rank = -1;
```
shanedsnyder, line 361: Probably worth pointing out that if the caller passes in records representing partial (i.e. not all ranks) access to a shared file, it's going to be assigned a `-1` rank, which has always been used to represent globally shared ranks in Darshan.

carns: Yeah, good call; that's a subtle issue.  I just added a big comment block to the agg_record fn ptr as a warning.

darshan-util/darshan-mpiio-logutils.c
```diff
@@ -518,42 +585,62 @@ static void darshan_log_agg_mpiio_files(void *rec, void *agg_rec, int init_flag)
...
+                    break;
                 }
-
-                if(mpi_time > agg_mpi_rec->fcounters[MPIIO_F_SLOWEST_RANK_TIME])
-                {
-                    agg_mpi_rec->counters[MPIIO_SLOWEST_RANK] = mpi_rec->base_rec.rank;
-                    agg_mpi_rec->counters[MPIIO_SLOWEST_RANK_BYTES] = mpi_bytes;
-                    agg_mpi_rec->fcounters[MPIIO_F_SLOWEST_RANK_TIME] = mpi_time;
+                if (init_flag ||
+                    mpi_slowest_time < agg_mpi_rec->fcounters[MPIIO_F_SLOWEST_RANK_TIME]) {
```
shanedsnyder, line 627: Looks like we need to flip the the `<` to `>` here (and in POSIX and STDIO codepaths). For slowest time, we want the highest number.

carns: Thanks @shanedsnyder , you are right!  I just pushed a fix.

Issue #663, opened 2022-02-16
general HDF5 module support


shanedsnyder:
We've made a ton of progress recently in getting v1 of the summary reports completed, but wanted to create an issue so we don't forget to generally include HDF5 support in our figures. I imagine it's not too difficult to extend the different figures to include HDF5 module data. I notice we do have some HDF5 data related to access sizes already, but we could probably include it in the op counts graphs, as well as in the I/O cost graph.

Thinking about it more, the HDF5 module probably requires a bit of special care since it's really 2 modules (H5F for file access, and H5D for dataset access). For I/O cost graphs and op count graphs, we will probably want to just combine the info from each module into a single "HDF5" field. Similarly, the per-module section of the report probably has an "HDF5" module section that characterizes both the H5F and H5D modules.

shanedsnyder:
Maybe we can just flesh out details at our next meeting and I can update the issue once it's more clear what the plan is.

nawtrey:
If we are going to combine module data that could take some care. We probably want it handled in some kind of aggregator that stores it in the `DarshanReport`, like we have for some of the other figures. 

For HDF5 logs, we currently only have `ior_hdf5_example.darshan` and `shane_macsio_id29959_5-22-32552-7035573431850780836_1590156158.darshan` that contain `H5F` and `H5D` modules, are these are sufficient cases for testing? If we plan to officially support HDF5 we should probably have representative logs for testing.

nawtrey:
Tyler had some good comments in today's meeting about collecting testing logs for this issue which I don't think my comment above captured. He suggested that we collect/create logs with known or easy-to-determine values (say we know it opened 5 files) for testing. Since we can't use the original darshan reports to confirm our outputs for HDF5-related things, this would give us a lot more confidence moving forward, especially if the data post-processing gets complicated.

tylerjereddy:
A Python-based approach like the one in https://github.com/darshan-hpc/darshan-logs/pull/22 and https://github.com/tylerjereddy/heatmap_diagonal could likely be used with `h5py` to do pretty much anything you want for producing custom `darshan` logs with "ground truth" HDF5 activity.

nawtrey:
I've been trying to develop some logs for testing here, and I've created logs for a couple cases that I would like some feedback on. I started with Tyler's approach and used `h5py` to write to 1 file/dataset per rank:

<details><summary>HDF5 diagonal code</summary>

```python
import os
import time
import argparse

import numpy as np
[...]

PR #685, opened 2022-03-09
WIP, ENH: HDF5 support for `plot_opcounts()` aggregators

nawtrey:
Description
------------
* Adds `H5F` and `H5D` specific sections to `plot_opcounts()`
aggregator functions `agg_ioops()` and `gather_count_data()`.
Current method does not combine `H5F` and `H5D` module
data together.

* Add `H5F` and `H5D` test cases to `test_plot_exp_common`
tests `test_xticks_and_labels` and `test_bar_heights`

* Contributes to issue #663

Considerations
----------------
I've played around with combining the `H5F` and `H5D` module data, and we can combine them fairly easily in `gather_count_data()` if we want to. The following diff shows one way of accomplishing this on this branch:

<details><summary>Diff</summary>

```diff
diff --git a/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py b/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
index 97ac093..501268c 100644
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
@@ -83,11 +83,12 @@ def gather_count_data(report, mod):
 
     elif mod == 'H5D':
         labels = ['Read', 'Write', 'Open', 'Flush']
+        h5f_data = report.summary['agg_ioops']["H5F"]
         counts = [
             mod_data['H5D_READS'],
             mod_data['H5D_WRITES'],
-            mod_data['H5D_OPENS'],
-            mod_data['H5D_FLUSHES'],
+            mod_data['H5D_OPENS'] + h5f_data['H5F_OPENS'],
+            mod_data['H5D_FLUSHES'] + h5f_data['H5F_FLUSHES'],
         ]
 
     return labels, counts
```

</details>

With the above changes, `plot_opcounts(mod="H5F")` would return only the file data (for cases where only `H5F` is present), and `plot_opcounts(mod="H5D")` would return the combined HDF5 data. So it wouldn't be terribly difficult to add it to the report this way, but it is probably confusing to input a single module and get 2 modules' data. This does allow us to stick to the original module names  (which are documented), but there might be an argument for allowing "HDF5" as an input for plotting functions that support it. 

I'm not sure what is the best way to do this, so if reviewers have a better idea I'd be happy to implement it. Maybe @jakobluettgau has a better vision for this since he wrote the original aggregator functions. We could also just merge this and deal with the module combining in a separate PR, I just thought I would mention it since it can be done here.

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/aggregators/agg_ioops.py
+++ b/darshan-util/pydarshan/darshan/experimental/aggregators/agg_ioops.py
@@ -24,7 +24,7 @@ def agg_ioops(self, mode='append'):
     ctx = {}
 
     # aggragate
-    mods = ['MPI-IO', 'POSIX', 'STDIO']
+    mods = ['MPI-IO', 'POSIX', 'STDIO', "H5F", "H5D"]
     for mod in mods:
 
         # check records for module are present
@@ -70,6 +70,24 @@ def agg_ioops(self, mode='append'):
             ctx[mod] = agg
             ctx[mod + '_indep_simple'] = agg_indep
             ctx[mod + '_coll_simple'] = agg_coll
+
+        elif mod == "H5F":
+            tmp = {
+                'Open':  agg[mod + '_OPENS'],
+                'Flush': agg[mod + '_FLUSHES'],
+            }
+            ctx[mod] = agg
+            ctx[mod + '_simple'] = tmp
+
+        elif mod == "H5D":
+            tmp = {
+                'Read':  agg[mod + '_READS'],
+                'Write': agg[mod + '_WRITES'],
+                'Open':  agg[mod + '_OPENS'],
+                'Flush': agg[mod + '_FLUSHES'],
+            }
+            ctx[mod] = agg
+            ctx[mod + '_simple'] = tmp
 
         else:
             # POSIX and STDIO share most counter names and are handled 
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
@@ -82,12 +82,17 @@ def gather_count_data(report, mod):
         ]
 
     elif mod == 'H5D':
-        labels = ['Read', 'Write', 'Open', 'Flush']
+        labels = [
+            'Data Read', 'Data Write', 'Data Open',
+            'Data Flush', 'File Open', 'File Flush',
+        ]
         counts = [
-            mod_data['H5D_READS'],
-            mod_data['H5D_WRITES'],
-            mod_data['H5D_OPENS'],
-            mod_data['H5D_FLUSHES'],
+            report.summary['agg_ioops']['H5D']['H5D_READS'],
+            report.summary['agg_ioops']['H5D']['H5D_WRITES'],
+            report.summary['agg_ioops']['H5D']['H5D_OPENS'],
+            report.summary['agg_ioops']['H5D']['H5D_FLUSHES'],
+            report.summary['agg_ioops']['H5F']['H5F_OPENS'],
+            report.summary['agg_ioops']['H5F']['H5F_FLUSHES'],
         ]
 
     return labels, counts
@@ -96,8 +101,15 @@ def plot_opcounts(report, mod, ax=None):
     """
     Generates a bar chart summary for operation counts.
 
-	Args:
-        report (DarshanReport): darshan report object to plot
+    Parameters
+    ----------
+
+    report (DarshanReport): darshan report object to plot
+
+    mod: the module to plot operation counts for (i.e. "POSIX",
+    "MPI-IO", "STDIO", "H5F", "H5D"). If "H5D" is input the returned
+    figure will contain both "H5F" and "H5D" module data.
+
     """
 
     if ax is None:
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
@@ -74,6 +74,22 @@ def gather_count_data(report, mod):
             mod_data['STDIO_FLUSHES']
         ]
 
+    elif mod == 'H5F':
+        labels = ['Open', 'Flush']
+        counts = [
+            mod_data['H5F_OPENS'],
+            mod_data['H5F_FLUSHES'],
+        ]
+
+    elif mod == 'H5D':
+        labels = ['Read', 'Write', 'Open', 'Flush']
+        counts = [
+            mod_data['H5D_READS'],
+            mod_data['H5D_WRITES'],
+            mod_data['H5D_OPENS'],
+            mod_data['H5D_FLUSHES'],
+        ]
+
     return labels, counts
 
 def plot_opcounts(report, mod, ax=None):
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_opcounts.py
@@ -83,8 +83,8 @@ def gather_count_data(report, mod):
 
     elif mod == 'H5D':
         labels = [
-            'Data Read', 'Data Write', 'Data Open',
-            'Data Flush', 'File Open', 'File Flush',
+            'H5D Read', 'H5D Write', 'H5D Open',
+            'H5D Flush', 'H5F Open', 'H5F Flush',
         ]
         counts = [
             report.summary['agg_ioops']['H5D']['H5D_READS'],
--- a/darshan-util/pydarshan/darshan/tests/test_plot_exp_common.py
+++ b/darshan-util/pydarshan/darshan/tests/test_plot_exp_common.py
@@ -110,6 +110,18 @@ darshan.enable_experimental()
             ['Ind. Read', 'Ind. Write', 'Ind. Open',
             'Col. Read', 'Col. Write', 'Col. Open', 'Sync'],
         ),
+        (
+            "ior_hdf5_example.darshan",
+            "H5F",
+            plot_opcounts,
+            ['Open', 'Flush'],
+        ),
+        (
+            "ior_hdf5_example.darshan",
+            "H5D",
+            plot_opcounts,
+            ['Read', 'Write', 'Open', 'Flush'],
+        ),
     ],
 )
 def test_xticks_and_labels(log_path, func, expected_xticklabels, mod):
@@ -250,7 +262,18 @@ def test_xticks_and_labels(log_path, func, expected_xticklabels, mod):
             plot_opcounts,
             [0, 7695, 0, 0, 64, 16, 0],
         ),
-
+        (
+            "ior_hdf5_example.darshan",
+            "H5F",
+            plot_opcounts,
+            [6, 0],
+        ),
+        (
+            "ior_hdf5_example.darshan",
+            "H5D",
+            plot_opcounts,
+            [12, 12, 6, 0],
+        ),
     ],
 )
 def test_bar_heights(filename, mod, fig_func, expected_heights):
--- a/darshan-util/pydarshan/darshan/tests/test_plot_exp_common.py
+++ b/darshan-util/pydarshan/darshan/tests/test_plot_exp_common.py
@@ -120,8 +120,8 @@ darshan.enable_experimental()
             "ior_hdf5_example.darshan",
             "H5D",
             plot_opcounts,
-            ['Data Read', 'Data Write', 'Data Open',
-            'Data Flush', 'File Open', 'File Flush'],
+            ['H5D Read', 'H5D Write', 'H5D Open',
+            'H5D Flush', 'H5F Open', 'H5F Flush'],
         ),
     ],
 )
--- a/darshan-util/pydarshan/darshan/tests/test_plot_exp_common.py
+++ b/darshan-util/pydarshan/darshan/tests/test_plot_exp_common.py
@@ -120,7 +120,8 @@ darshan.enable_experimental()
             "ior_hdf5_example.darshan",
             "H5D",
             plot_opcounts,
-            ['Read', 'Write', 'Open', 'Flush'],
+            ['Data Read', 'Data Write', 'Data Open',
+            'Data Flush', 'File Open', 'File Flush'],
         ),
     ],
 )
@@ -272,7 +273,7 @@ def test_xticks_and_labels(log_path, func, expected_xticklabels, mod):
             "ior_hdf5_example.darshan",
             "H5D",
             plot_opcounts,
-            [12, 12, 6, 0],
+            [12, 12, 6, 0, 6, 0],
         ),
     ],
 )


tylerjereddy (commented):
One thing I don't see clearly analyzed is whether `H5F_OPENS` and `H5D_OPENS` can mean different things in practice? For example, https://github.com/darshan-hpc/darshan/issues/663#issuecomment-1062419365 indicates:

> Unrelated to my last comment, my understanding of how we want to combine the H5F and H5D "Operation Counts" data is to sum the H5F and H5D totals.

> The H5F sum is always going to be greater than or equal to the H5D sum, but it would avoid the sort of "double counting" that we would get with a simple sum.

Indeed, I'm leaning toward your later ideas in that comment--naively summing both seems problematic to me for cases where it is the same file handle opening event, given that the matching issue describes a fused module of sorts.

Given the read-related issues in gh-663, an example analysis I might want to see is setting a subset of ranks to open files for writing, but not write anything, while other ranks do write something, and then use that to confirm a discrepancy between `H5F` and `H5D` open counters:

```python
from mpi4py import MPI
import numpy as np
from numpy.testing import assert_array_equal
import h5py

def main():
    comm = MPI.COMM_WORLD
    rank = comm.Get_rank()

    # round trip HDF5 IO test
    file_path = f"./test_rank_{rank}.hdf5"
    n_bytes = 10 * (rank + 1)

    bytes_to_write = np.ones(shape=n_bytes, dtype=np.int8)

    with h5py.File(file_path, "w") as f:
        if rank % 2 == 0:
            f.create_dataset("dataset", data=bytes_to_write)
        else:
            # odd ranks only open the file for writing, but
            # write nothing to it
            pass
        


if __name__ == "__main__":
    main()
```

In this case the result (below the fold) is as expected--`H5F` records 10 ranks opening files, regardless of whether they write something, while `H5D` only records even-numbered ranks that actually write data for opening events.

So, it seems like a shim layer/function might be in order--for the small number of fields between `H5F` and `H5D` that conceptually overlap, do a sort of intelligent join operation that avoids redundant counting.

There is perhaps even a pandas join or merge operation that does this kind of thing without much wrangling, since it is not an uncommon database type operation. Can you think of any scenarios where the HDF5 opens and flushes would not be properly accounted for by `H5F` columns alone?

If not, then a pandas merge/join/concatenate of some sort, with some renaming to `HDF5` in a few column names/places might suffice. If there are some tricky cases, then an intermediate function of some sort may need to some parsing.

<details>

```
{'POSIX': <darshan.report.DarshanRecordCollection object at 0x7fc0fcda9580>, 'H5F': <darshan.report.DarshanRecordCollection object at 0x7fc0fc8304c0>, 'H5D': <darshan.report.DarshanRecordCollection object at 0x7fc0fc830fd0>, 'DXT_POSIX': <darshan.report.DarshanRecordCollection object at 0x7fc0fcd80670>}
H5D dataframe counters:
    rank                    id  H5D_OPENS  H5D_READS  H5D_WRITES  H5D_FLUSHES  H5D_BYTES_READ  H5D_BYTES_WRITTEN  H5D_RW_SWITCHES  ...  H5D_CHUNK_SIZE_D3  H5D_CHUNK_SIZE_D4  H5D_CHUNK_SIZE_D5  H5D_USE_MPIIO_COLLECTIVE  H5D_USE_DEPRECATED  H5D_FASTEST_RANK  H5D_FASTEST_RANK_BYTES  H5D_SLOWEST_RANK  H5D_SLOWEST_RANK_BYTES
0     0  11686675357054819694          1          0           1            0               0                 10                0  ...                  0                  0                  0                         0                   0                 0                       0                 0                       0
1     2   8163471921814110286          1          0           1            0               0                 30                0  ...                  0                  0                  0                         0                   0                 0                       0                 0                       0
2     4   5895888996501598751          1          0           1            0               0                 50                0  ...                  0                  0                  0                         0                   0                 0                       0                 0                       0
3     6  11021700834876979282          1          0           1            0               0                 70                0  ...                  0                  0                  0                         0                   0                 0                       0                 0                       0
4     8   6635034257651523170          1          0           1            0               0                 90                0  ...                  0                  0                  0                         0                   0                 0                       0                 0                       0

[5 rows x 96 columns]
H5D dataframe fcounters:
    rank                    id  H5D_F_OPEN_START_TIMESTAMP  H5D_F_READ_START_TIMESTAMP  H5D_F_WRITE_START_TIMESTAMP  H5D_F_CLOSE_START_TIMESTAMP  ...  H5D_F_MAX_READ_TIME  H5D_F_MAX_WRITE_TIME  H5D_F_FASTEST_RANK_TIME  H5D_F_SLOWEST_RANK_TIME  H5D_F_VARIANCE_RANK_TIME  H5D_F_VARIANCE_RANK_BYTES
0     0  11686675357054819694                    0.787025                         0.0                     0.787257                          0.0  ...                  0.0              0.000046                      0.0                      0.0                       0.0                        0.0
1     2   8163471921814110286                    0.792443                         0.0                     0.792648                          0.0  ...                  0.0              0.000041                      0.0                      0.0                       0.0                        0.0
2     4   5895888996501598751                    0.785701                         0.0                     0.785906                          0.0  ...                  0.0              0.000041                      0.0                      0.0                       0.0                        0.0
3     6  11021700834876979282                    0.793039                         0.0                     0.793270                          0.0  ...                  0.0              0.000050                      0.0                      0.0                       0.0                        0.0
4     8   6635034257651523170                    0.784874                         0.0                     0.785060                          0.0  ...                  0.0              0.000050                      0.0                      0.0                       0.0                        0.0

[5 rows x 19 columns]
H5F dataframe counters:
    rank                    id  H5F_OPENS  H5F_FLUSHES  H5F_USE_MPIIO
0     0  12338924892616139460          1            0              0
1     1   2374806179835333680          1            0              0
2     2   4741602306122135968          1            0              0
3     3   6304996309454365857          1            0              0
4     4  14038849941392865518          1            0              0
5     5  17284205279612898412          1            0              0
6     6  11364139169532633768          1            0              0
7     7  12429378654238249285          1            0              0
8     8   1186195185660660511          1            0              0
9     9  10257107547145179444          1            0              0
H5F dataframe fcounters:
    rank                    id  H5F_F_OPEN_START_TIMESTAMP  H5F_F_CLOSE_START_TIMESTAMP  H5F_F_OPEN_END_TIMESTAMP  H5F_F_CLOSE_END_TIMESTAMP  H5F_F_META_TIME
0     0  12338924892616139460                    0.777810                          0.0                  0.786832                        0.0         0.009022
1     1   2374806179835333680                    0.777776                          0.0                  0.786602                        0.0         0.008826
2     2   4741602306122135968                    0.777799                          0.0                  0.792122                        0.0         0.014323
3     3   6304996309454365857                    0.777874                          0.0                  0.784825                        0.0         0.006951
4     4  14038849941392865518                    0.777780                          0.0                  0.785472                        0.0         0.007692
5     5  17284205279612898412                    0.777836                          0.0                  0.792042                        0.0         0.014206
6     6  11364139169532633768                    0.777750                          0.0                  0.792337                        0.0         0.014587
7     7  12429378654238249285                    0.777849                          0.0                  0.786175                        0.0         0.008325
8     8   1186195185660660511                    0.777625                          0.0                  0.784057                        0.0         0.006432
9     9  10257107547145179444                    0.777692                          0.0                  0.784916                        0.0         0.007224
```

</details>

nawtrey:
>Can you think of any scenarios where the HDF5 opens and flushes would not be properly accounted for by H5F columns alone?

I can't think of an instance where the `H5F` Open/Flush counters would not be a full account, but that's based on my limited understanding of these functions. 

To check if the flushes are working as I would expect, I created another log which calls `h5py.File.flush()` for half of the ranks. It appears to be picked up by `H5F` and not `H5D`, but it does seem to work as expected:

<details>

```
H5F DataFrame: 
   rank  H5F_OPENS  H5F_FLUSHES
0     0          1            1
1     1          1            0
2     2          1            1
3     3          1            0
4     4          1            1
5     5          1            0
6     6          1            1
7     7          1            0
8     8          1            1
9     9          1            0

H5D DataFrame: 
   rank  H5D_OPENS  H5D_FLUSHES  H5D_READS  H5D_WRITES  H5D_BYTES_READ  H5D_BYTES_WRITTEN
0     0          1            0          0           1               0                  1
1     1          1            0          0           1               0                  1
2     2          1            0          0           1               0                  1
3     3          1            0          0           1               0                  1
4     4          1            0          0           1               0                  1
5     5          1            0          0           1               0                  1
6     6          1            0          0           1               0                  1
7     7          1            0          0           1               0                  1
8     8          1            0          0           1               0                  1
9     9          1            0          0           1               0                  1
```

</details>

> So, it seems like a shim layer/function might be in order--for the small number of fields between H5F and H5D that conceptually overlap, do a sort of intelligent join operation that avoids redundant counting.

Yeah, there should be a simple way to do this with pandas. I think we can refactor some of the aggregation code which will make this a little more straightforward (see gh-689).

tylerjereddy:
Sounds like keeping most fields separate between `H5D` and `H5F` except for maybe combining "meta" times is what Shane has in mind, so my "review" comment is pretty far off base it seems, and the original intent of the PR seems to be close to what Shane wants at first glance (no need to relabel a bunch of columns or do database joins I don't think)

shanedsnyder:
> Sounds like keeping most fields separate between `H5D` and `H5F` except for maybe combining "meta" times is what Shane has in mind, so my "review" comment is pretty far off base it seems, and the original intent of the PR seems to be close to what Shane wants at first glance (no need to relabel a bunch of columns or do database joins I don't think)

Yep, sorry about the confusion guys. Just to make sure we're on the same page, I think the op counts graph would have the following:
- "file opens" (aggregated H5F_OPENS values)
- "dataset opens" (aggregated H5D_OPENS values)
- "dataset writes" (aggregated H5D_WRITES values)
- "dataset reads" (aggregated H5D_READS values)
- "file flushes" (aggregated H5F_FLUSHES values)
- "dataset flushes" (aggregated H5D_FLUSHES values)

(Not necessarily related to this PR, but as Tyler mentions, for the I/O cost plots that independently measure read/write/meta time components, you will need to combine the total META_TIMES from dataset and file modules. You shouldn't need to do anything like that for the op counts, though).

shanedsnyder:
> With the above changes, `plot_opcounts(mod="H5F")` would return only the file data (for cases where only `H5F` is present), and `plot_opcounts(mod="H5D")` would return the combined HDF5 data. So it wouldn't be terribly difficult to add it to the report this way, but it is probably confusing to input a single module and get 2 modules' data. This does allow us to stick to the original module names (which are documented), but there might be an argument for allowing "HDF5" as an input for plotting functions that support it.

I don't think I feel strongly enough about how you do this to handcuff how you implement, but passing in "HDF5" as the module name and letting the implementation get both H5F and H5D data seems reasonable. Or maybe it could take a list of modules as input without too much trouble? What you have already could work though, it just could be confusing as to how multiple modules are being combined when you ask for just one of them as you point out.

nawtrey:
Yeah this PR doesn't really deal with combining the `H5F`/`H5D` modules, although at this point I think I'm leaning towards combining them in `agg_ioops()`. But I think that can be handled in a separate PR.

nawtrey:
I pushed in some changes that allow `plot_opcounts()` to plot either `H5F` data or both `H5F` and `H5D` data, when `H5F` or `H5D` are input, respectively. This is still potentially misleading, like I mentioned before, but I didn't think it was worth the trouble to add a new module name "HDF5". Instead I added a note to the documentation for `plot_opcounts()` so this is more clear. 

For `ior_hdf5_example.darshan` with `H5F`:
ior_hdf5_example darshan_H5F

For `ior_hdf5_example.darshan` with `H5D`:
ior_hdf5_example darshan_H5D

I opted for using "data" over "dataset" to get the labels to clear, but if we want to use "dataset" we certainly can, I'll just have to rotate the labels.

jakobluettgau:
I think I would stick to the lingo used by the HDF5 API to avoid confusion or accidentally introducing ambiguity. H5D or just dataset resonate more with me at least.

nawtrey:
> I think I would stick to the lingo used by the HDF5 API to avoid confusion or accidentally introducing ambiguity. H5D or just dataset resonate more with me at least.

So something like "H5D Read", "H5D Write", "H5F Open", etc?

shanedsnyder:
> I think I would stick to the lingo used by the HDF5 API to avoid confusion or accidentally introducing ambiguity. H5D or just dataset resonate more with me at least.

Agreed, let's try to find a way to include "dataset" without jumbling up the x-axis labels. One alternative to rotating would be to just have the label span 2 rows? 
```
dataset      file
 open        open
```
Maybe that's a pain to pull off cleanly with matplotlib, though?

I'd vote for more generic "file" and "dataset" descriptors rather than "H5F" or "H5D" -- it may not be immediately clear to users what the differences are there just using the API prefixes, but they may know enough about HDF5 to understand what we mean when we say files or datasets.

nawtrey:
Okay I've updated the histogram bar labels to the method @jakobluettgau mentioned above.

For example, here is one for `ior_hdf5_example.darshan`:
ior_hdf5_example darshan_H5D

Here is a script that generates a bunch of example cases to look over:

<details>

```python
import darshan
from darshan.experimental.plots import plot_opcounts
darshan.enable_experimental()

def main(log_filename):
    log_filepath = get_log_path(log_filename)
    report = darshan.DarshanReport(log_filepath)
    for mod in ["H5F", "H5D"]:
        if mod in report.modules:
            fig = plot_opcounts(report, mod=mod)
            fig.savefig(f"{log_filename}_{mod}.png", dpi=300)

if __name__ == '__main__':

    test_logs = [
        "ior_hdf5_example.darshan",
        "shane_macsio_id29959_5-22-32552-7035573431850780836_1590156158.darshan",
        "hdf5_file_opens_only.darshan",
        "hdf5_diagonal_write_1_byte_dxt.darshan",
        "hdf5_diagonal_write_bytes_range_dxt.darshan",
        "hdf5_diagonal_write_half_flush_dxt.darshan",
        "hdf5_diagonal_write_half_ranks_dxt.darshan",
    ]

    for log_filename in test_logs:
        main(log_filename=log_filename)
```

</details>

shanedsnyder:
I think the figures look good.

Just to make sure we're on the same page, the intent in darshan-job-summary reports is to only pass in `H5D` so we get the unified op counts report, right?

nawtrey:
> Just to make sure we're on the same page, the intent in darshan-job-summary reports is to only pass in `H5D` so we get the unified op counts report, right?

Yes, when it gets added to the summary report it will show the unified version except for the (rare) case where only `H5F` is present. I just wanted to cover both bases.

tylerjereddy:
This seems "ok," so I'll merge. I will note that this vastly underutilizes the "ground truth" HDF5 logs available in the logs repo though.

For example, it would likely only take a few minutes to add simple tests for i.e., the case that writes 1 byte via HDF5 on 10 separate ranks (`hdf5_diagonal_write_1_byte_dxt.darshan`):

image

And when only alternate ranks have 1 byte written, all files get opened and half get written:

image

These are fair bit easier for me to review since I can't really say what `ior..` is doing as easily as those trivial programs, and if such tests ever were to fail, we could be pretty confident of a bug somewhere given the simplicity.

PR #691, opened 2022-03-12
TST:  narrow heatmap xfail

tylerjereddy:
* ~we no longer need to mark runtime `HEATMAP` tests
as known failures, since the binding support has been
merged now~

* narrow the `HEATMAP` xfail to only exclude `e3sm_io_heatmap_and_dxt.darshan`, which still uses too much memory

### DIFF
--- a/darshan-util/pydarshan/darshan/tests/test_report.py
+++ b/darshan-util/pydarshan/darshan/tests/test_report.py
@@ -38,8 +38,6 @@ def test_jobid_type_all_logs_repo_files(log_filepath):
     # log files in the darshan_logs package;
     # this is primarily intended as a demonstration of looping
     # through all logs repo files in a test
-    if "heatmap" in log_filepath:
-        pytest.xfail(reason="no runtime HEATMAP support")
     report = darshan.DarshanReport(log_filepath)
     assert isinstance(report.metadata['job']['jobid'], int)
 
--- a/darshan-util/pydarshan/darshan/tests/test_report.py
+++ b/darshan-util/pydarshan/darshan/tests/test_report.py
@@ -38,6 +38,8 @@ def test_jobid_type_all_logs_repo_files(log_filepath):
     # log files in the darshan_logs package;
     # this is primarily intended as a demonstration of looping
     # through all logs repo files in a test
+    if "e3sm_io_heatmap_and_dxt" in log_filepath:
+        pytest.xfail(reason="large memory requirements")
     report = darshan.DarshanReport(log_filepath)
     assert isinstance(report.metadata['job']['jobid'], int)
 
--- a/darshan-util/pydarshan/darshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/darshan/tests/test_summary.py
@@ -152,6 +152,8 @@ def test_main_all_logs_repo_files(tmpdir, log_filepath):
     # on the Darshan logs from the logs repo:
     # https://github.com/darshan-hpc/darshan-logs
 
+    if "e3sm_io_heatmap_and_dxt" in log_filepath:
+        pytest.xfail(reason="large memory requirements")
     argv = [log_filepath]
     with mock.patch("sys.argv", [""] + argv):
         with tmpdir.as_cwd():
--- a/darshan-util/pydarshan/darshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/darshan/tests/test_summary.py
@@ -152,8 +152,6 @@ def test_main_all_logs_repo_files(tmpdir, log_filepath):
     # on the Darshan logs from the logs repo:
     # https://github.com/darshan-hpc/darshan-logs
 
-    if "heatmap" in log_filepath:
-        pytest.xfail(reason="no runtime HEATMAP support")
     argv = [log_filepath]
     with mock.patch("sys.argv", [""] + argv):
         with tmpdir.as_cwd():


tylerjereddy:
Interesting, looks like there's a hard failure in one of the tests maybe, on Linux/Python 3.9:

```
/home/runner/work/_temp/67ef905d-ab26-4af3-a5d0-5d8bf0d04f00.sh: line 7:  6525 Killed                  pytest --pyargs darshan --cov-report xml --cov=$site_packages/darshan
tests/test_summary.py ...................
Error: Process completed with exit code 137.
```

I'll restart to see if it is reproducible, then move the debugging offline if it shows up again.

tylerjereddy:
MacOS CI also had an issue installing `asv` I think, maybe that's a fluke, otherwise we can pin `asv` to a lower version maybe.

As for the original testing error, exit code 137 can mean a memory/resource errors, which brings me back to my concerns about memory footprints and sparse vs. dense data structures. CI virtual machines may be substantially more memory constrained, so may need to keep an eye on this.

tylerjereddy:
Analyzing the increase in memory usage reveals the unsurprising suspect, Shane's large DXT + runtime heatmap log file (`e3sm_io_heatmap_and_dxt.darshan`).

Using [`pytest-monitor`](https://pytest-monitor.readthedocs.io/en/latest/operating.html) and on the respective branches:
- `pytest --db ./monitor_pydarshan_devel.db --pyargs darshan`
- `pytest --db ./monitor_no_xfail.db --pyargs darshan`

Then printing out the 5 highest memory usage tests per single-core `pytest` run, we hit almost 10 GB memory usage in this branch. That's well above the 7 GB limit for Linux GHA runners documented: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources . We're already at 7+ GB without this PR.

<details>
<summary> Python code used to produce dataframes below the fold </summary>

```python
import sqlite3
import pandas as pd

for db in ["monitor_pydarshan_devel.db", "monitor_no_xfail.db"]:
    cnx = sqlite3.connect(db)
    df = pd.read_sql_query("select MEM_USAGE, ITEM_PATH, ITEM, ITEM_VARIANT FROM TEST_METRICS", cnx).sort_values(by=["MEM_USAGE"], ascending=False).head(10).to_markdown(index=False)
    print("df:\n", df)
```

</details>

So, if nothing else, this should provide some metrics to start thinking about:
1) Are these memory footprints reasonable?
2) What can be done to improve them, and what are the tradeoffs with other kinds of performance metrics like execution speed?
3) What short- and long-term steps are needed to keep the test suite sustainable? It may be that the `xfail` may need to remain for that particular log filename for now. It is the second-largest logfile at 11 MB I think, but this could be deceptive since the compression of the `.darshan` format can conceal sparsity->density conversions with zeros.

`pydarshan-devel`:


 |   MEM_USAGE (MB) | ITEM_PATH                  | ITEM                                   | ITEM_VARIANT                                                                                                                                                                                                                                     |
|------------:|:---------------------------|:---------------------------------------|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
|     7509.45 | darshan.tests.test_summary | test_module_table                      | test_module_table[partial_data_dxt.darshan, expected_df6, 2]                                                                                                                                                                                     |
|     7474.47 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[sample, badost.darshan, 779]                                                                                                                                                                                                    |
|     7474.16 | darshan.tests.test_summary | test_get_full_command                  | test_get_full_command[sample, badost.darshan, /global/project/projectdirs/m888/glock/tokio, abc, results/bin.edison/ior , H , k , w , o ior, posix.out , s 64 , f /global/project/projectdirs/m888/glock/tokio, abc, results/inputs/posix1m2.in] |
|     7474.14 | darshan.tests.test_summary | test_generate_fig_unsupported_fig_type | test_generate_fig_unsupported_fig_type                                                                                                                                                                                                           |
|     7474.14 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[sample, dxt, simple.darshan, < 1]                                                                                                                                                                                               |
|     7474.14 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[sample, goodost.darshan, 4]                                                                                                                                                                                                     |
|     7474.14 | darshan.tests.test_summary | test_get_full_command                  | test_get_full_command[dxt.darshan, N/A]                                                                                                                                                                                                          |
|     7473.67 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[noposixopens.darshan, 1110]                                                                                                                                                                                                     |
|     7473.67 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[noposix.darshan, 39212]                                                                                                                                                                                                         |
|     7473.64 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[sample.darshan, 116]     

`treddy_no_xfail_heatmap`

 |   MEM_USAGE (MB) | ITEM_PATH                  | ITEM                                   | ITEM_VARIANT                                                                                                                                                            |
|------------:|:---------------------------|:---------------------------------------|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
|     9996.66 | darshan.tests.test_summary | test_main_all_logs_repo_files          | test_main_all_logs_repo_files[/Users/treddy/python_venvs/python_395/lib/python3.9/site, packages/darshan_logs/e3sm_io_heatmaps_and_dxt/e3sm_io_heatmap_and_dxt.darshan] |
|     6392.65 | darshan.tests.test_summary | test_module_table                      | test_module_table[partial_data_dxt.darshan, expected_df6, 2]                                                                                                            |
|     6121.86 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[sample, badost.darshan, 779]                                                                                                                           |
|     6121.71 | darshan.tests.test_summary | test_generate_fig_unsupported_fig_type | test_generate_fig_unsupported_fig_type                                                                                                                                  |
|     6121.71 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[sample, dxt, simple.darshan, < 1]                                                                                                                      |
|     6121.56 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[sample, goodost.darshan, 4]                                                                                                                            |
|     6115.53 | darshan.tests.test_summary | test_get_full_command                  | test_get_full_command[dxt.darshan, N/A]                                                                                                                                 |
|     6115.09 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[noposixopens.darshan, 1110]                                                                                                                            |
|     6115.09 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[noposix.darshan, 39212]                                                                                                                                |
|     6115.03 | darshan.tests.test_summary | test_get_runtime                       | test_get_runtime[sample.darshan, 116]                                                                                                                                   |

nawtrey:
Using [`memory_profiler`](https://pypi.org/project/memory-profiler/) I was able to get a better idea of the root cause of the memory issue, and (unsurprisingly) it is `get_heatmap_df()` where we see the majority of the memory usage. 

I used the following script (similar to `test_main_all_logs_repo_files`) on `pydarshan-devel` to get a profile of the memory over time:

<details>

```python
import os
from unittest import mock

import darshan
from darshan.cli import summary

def test_main_all_logs_repo_files(log_filepath):
    argv = [log_filepath]
    with mock.patch("sys.argv", [""] + argv):
        print("Running summary.main()...")
        summary.main()

if __name__ == "__main__":
    log_filepath = get_log_path("e3sm_io_heatmap_and_dxt.darshan")
    test_main_all_logs_repo_files(log_filepath=log_filepath)
```

</details>

Memory Profile: 
mprof_e3sm

It's a bit messy to look at, but it seems to illustrate a similar (large) memory footprint. `test_reindex` is a dummy function created to check if the reindexing is a major contributor to the overall memory footprint, and it doesn't appear to be the case here. A zoomed-in view below illustrates this point. What is interesting is the memory usage before & after `get_heatmap_df()` seem to be roughly the same for both heatmaps being generated.

test_reindex

This was pretty surprising to me, but if the reindexing isn't causing a large change, I was curious what part of `get_heatmap_df()` was. I used the line profiling capabilities of `memory_profiler` with just `get_heatmap_df()` flagged to get a better look:

<details>

```
Report generated successfully. 
Saving report at location: /Users/nawtrey/darshan_install/darshan/darshan-util/pydarshan/e3sm_io_heatmap_and_dxt_report.html
Filename: /Users/nawtrey/darshan_install/darshan/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py

Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
   276  529.727 MiB -3566.320 MiB           2   @profile
   277                                         def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFrame:
   278                                             """
   279                                             Builds an array similar to a 2D-histogram, where the y data is the unique
   280                                             ranks and the x data is time. Each bin is populated with the data sum
   281                                             and/or proportionate data sum for all IO events read/written during the
   282                                             time spanned by the bin.
   283                                         
   284                                             Parameters
   285                                             ----------
   286                                         
   287                                             agg_df: a ``pd.DataFrame`` containing the aggregated data determined
   288                                             by the input modules and operations.
   289                                         
   290                                             xbins: the number of x-axis bins to create.
   291                                         
   292                                             nprocs: the number of MPI ranks/processes used at runtime.
   293                                         
   294                                             Returns
   295                                             -------
   296                                         
   297                                             hmap_df: dataframe with time intervals for columns and rank
   298                                             index (0, 1, etc.) for rows, where each element contains the data
   299                                             read/written by the corresponding rank in the given time interval.
   300                                         
   301                                             Examples
   302                                             --------
   303                                             The first column/bin for the `hmap_df` generated from
   304                                             "examples/example-logs/ior_hdf5_example.darshan":
   305                                         
   306                                                        (0.0, 0.09552296002705891]
   307                                                 rank
   308                                                 0                   8.951484e+05
   309                                                 1                   3.746313e+05
   310                                                 2                   6.350999e+05
   311                                                 3                   1.048576e+06
   312                                         
   313                                             """
   314                                             # generate the bin edges by generating an array of length n_bins+1, then
   315                                             # taking pairs of data points as the min/max bin value
   316  529.727 MiB    0.523 MiB           2       max_time = agg_df["end_time"].max()
   317  529.727 MiB    0.000 MiB           2       bin_edge_data = np.linspace(0.0, max_time, xbins + 1)
   318                                             # create dummy variables for start/end time data, where dataframe columns
   319                                             # are the x-axis bin ranges
   320  620.715 MiB  136.805 MiB           2       cats_start = pd.get_dummies(pd.cut(agg_df["start_time"], bin_edge_data, precision=16))
   321  739.164 MiB   60.090 MiB           2       cats_end = pd.get_dummies(pd.cut(agg_df["end_time"], bin_edge_data, precision=16))
   322                                             # get series for the elapsed times for each dxt segment
   323  739.164 MiB -176.809 MiB           2       elapsed_times_dxt_segments = agg_df["end_time"] - agg_df["start_time"]
   324                                             # calculate the time interval spanned by each bin
   325  739.164 MiB -176.809 MiB           2       bin_size = bin_edge_data[1] - bin_edge_data[0]
   326                                             # get the ratio of the bin time interval over the dxt segment elapsed times
   327  739.176 MiB -176.797 MiB           2       true_fraction_of_dxt_segment_in_a_bin = bin_size / elapsed_times_dxt_segments
   328                                             # create a version of the above series where any ratio above 1 is set to 1
   329  743.914 MiB -171.234 MiB           2       fraction_of_dxt_segment_in_a_bin = true_fraction_of_dxt_segment_in_a_bin.copy()
   330  743.914 MiB -180.711 MiB           2       fraction_of_dxt_segment_in_a_bin.mask(fraction_of_dxt_segment_in_a_bin > 1, 1, inplace=True)
   331                                             # create a dataframe of binned dxt start events where occupied bins contain
   332                                             # the start time and unoccupied bins are populated with NaNs
   333 1928.223 MiB 1214.926 MiB           2       start_times_in_bins = cats_start.mul(agg_df["start_time"], axis=0).replace(0, np.nan)
   334                                             # create a dataframe of binned dxt start events where occupied bins contain
   335                                             # the proportional time spent in the bin
   336 2875.801 MiB  -37.051 MiB           2       start_fraction_bin_occupancy = (bin_edge_data[1:] - start_times_in_bins) / bin_size
   337                                             # create a dataframe of binned dxt end events where occupied bins contain
   338                                             # the end time and unoccupied bins are populated with NaNs
   339 3823.367 MiB -815.578 MiB           2       end_times_in_bins = cats_end.mul(agg_df["end_time"], axis=0).replace(0, np.nan)
   340                                             # create a dataframe of binned dxt end events where occupied bins contain
   341                                             # the proportional time spent in the bin
   342 4770.934 MiB -1594.082 MiB           2       end_fraction_bin_occupancy = (end_times_in_bins - bin_edge_data[:-1]) / bin_size
   343                                             # combine the start/end fractional bin occupancy dataframes by multiplying
   344                                             # them together. Fill any missing values with (1)
   345 5804.309 MiB -2253.551 MiB           2       combo = start_fraction_bin_occupancy.mul(end_fraction_bin_occupancy, fill_value=1, axis=0)
   346 5886.766 MiB -4237.855 MiB           2       combo.mask(combo > 1, 1, inplace=True)
   347                                             # add the start/end dummy variable dataframes
   348                                             # and replace any zeros with NaN's
   349 5887.020 MiB -4402.516 MiB           2       cats = cats_start.add(cats_end, fill_value=0)
   350 6829.996 MiB -3290.516 MiB           2       cats.replace(0, np.nan, inplace=True)
   351                                             # for each row (IO event) fill in any empty (NaN) bins
   352                                             # between filled bins because those are time spans b/w start
   353                                             # and stop events
   354 5276.996 MiB -6813.656 MiB           2       cats.interpolate(method="linear", limit_area="inside", axis=1, inplace=True)
   355                                             # each time bin containing an event has a 1 in it, otherwise NaN
   356                                             # store mask for restoring fully occupied bins
   357 5395.445 MiB -3568.074 MiB           2       mask_occ = cats == 2
   358                                             # set the fraction of segment per occupied bin
   359 5390.770 MiB -3809.648 MiB           2       cats = cats.mul(fraction_of_dxt_segment_in_a_bin, axis=0)
   360                                             # start/end in adjacent bins require special treatment
   361 5602.375 MiB -3551.797 MiB           2       adjacent_mask = (cats.count(axis=1) == 2) & (cats.sum(axis=1) == 2)
   362                                             # multiply them by the true fractions of DXT segments, which
   363                                             # may actually be > 1 (i.e., the non-1-capped fractions)
   364 5607.203 MiB -3774.867 MiB           2       cats[adjacent_mask] = cats[adjacent_mask].mul(true_fraction_of_dxt_segment_in_a_bin[adjacent_mask], axis=0)
   365                                             # adjust the fractions of the starts/ends for partial occupancy
   366                                             # of bins
   367 4407.176 MiB -4922.219 MiB           2       cats = cats.mul(combo, fill_value=1, axis=0)
   368                                             # start and end are in bin, so restore it
   369 4569.332 MiB -2360.008 MiB           2       cats = cats.mask(mask_occ, 1)
   370                                             # each full or fractional bin event is now multiplied by
   371                                             # the bytes data
   372 4574.094 MiB -2679.559 MiB           2       cats = cats.mul(agg_df["length"], axis=0)
   373 4574.102 MiB -2689.074 MiB           2       cats.index = agg_df["rank"]
   374 4590.922 MiB -2814.688 MiB           2       hmap_df = cats.groupby("rank").sum()
   375 4590.926 MiB -2848.324 MiB           2       hmap_df = hmap_df.reindex(index=range(nprocs), fill_value=0.0)
   376 4590.930 MiB -2848.328 MiB           2       return hmap_df
```

</details>

From the `memory_profiler` documentation (emphasis mine):
>The **first column** represents the line number of the code that has been profiled, the **second column** (Mem usage) the memory usage of the Python interpreter after that line has been executed. The **third column** (Increment) represents the difference in memory of the current line with respect to the last one. The **last column** (Line Contents) prints the code that has been profiled.

Several rows of the `Increment` column don't make sense to me, but looking at the memory usage column, it does seem like the footprint grows steadily throughout the function. I don't think there is any 1 culrpit since we are storing several different equally-shaped dataframes, but this does seem like the primary issue considering we see a ~4GiB change from start to finish.

Here is an archive containing the `.dat` file used to generate the figures above (using `mprof plot`): [mprof_runs.tar.gz](https://github.com/darshan-hpc/darshan/files/8247230/mprof_runs.tar.gz)

As for solutions, I think it depends on how much memory usage we can tolerate. Obviously if we have a hard limit at 7GB for tests, then we may have to avoid generating the DXT heatmaps for larger cases like this one. But if this is an intolerable amount for users, then we will have to find a way to get more efficiency out of this. There might be some low hanging fruit in terms of improving memory usage in `get_heatmap_df()`, maybe some of the sparse representations could be leveraged, but we might want to get an idea of what is a reasonable footprint before time is spent there.

nawtrey:
> There's probably a way to use a smaller memory buffer to accumulate/bin/combine/process DXT records by rank as they get read in, as long as the time bins/bounds are established early on. 

At the moment the range is based on the max value in the aggregate dataframe, but this will be changed to the "end time" stored in the metadata eventually (gh-575), which should be easy enough to access. Pandas also has its own [interval function](https://pandas.pydata.org/docs/reference/api/pandas.interval_range.html#pandas.interval_range) that we might use if we want to get away from using the dummy variables, although I'm not entirely sure how we could accomplish that yet.  

I played around with  `get_heatmap_df()` a bit and was able to reduce the peak memory by ~25% by allowing some of the dataframes to be garbage collected sooner, and rearranging the execution order of the code:

25_percent_reduction

The changes come with a 20-30% increase in run time for many of the DXT heatmap benchmark cases. This is likely because I reconstruct some of the intermediate dataframes twice instead of keeping them throughout. *Maybe* that's an "okay" short-term solution for now, but I'm not totally sold on it. The DXT heatmap is already the slowest figure to generate, so slowing it down further doesn't excite me

tylerjereddy:
I think a carefully considered solution will be both faster and more memory efficient yeah (perhaps something in cffi/Cython/C, although I wouldn't rule out Python just yet if the idea is solid enough).

Short term, I'll probably just revise this to `xfail` more narrowly on the big file only.

tylerjereddy:
Ok, I revised the PR to narrow the `xfail` instead of eliminate it--this branch runs 4 extra tests vs. `pydarshan-devel`.

carns:
Presumably the memory pressure will be lower when directly plotting heatmap module data (without DXT)?

In the long run, for logs with DXT data, we might have to decide if the DXT data should be used automatically (because of it's higher fidelity) or if it should be an optional argument, at least when the module data is above some threshold.  We could put a message in the warning box along the lines of "DXT data is present in the log but was not used to generate this summary.  Rerun the report with --dxt (this may be more memory and compute intensive) to ..."

nawtrey:
> Presumably the memory pressure will be lower when directly plotting heatmap module data (without DXT)?

Yeah the runtime heatmap data shouldn't be a problem at all. The final DXT dataframe is only `800.1 KB`, as Tyler pointed out, so the runtime heatmap should be the same.

> In the long run, for logs with DXT data, we might have to decide if the DXT data should be used automatically (because of it's higher fidelity) or if it should be an optional argument, at least when the module data is above some threshold. We could put a message in the warning box along the lines of "DXT data is present in the log but was not used to generate this summary. Rerun the report with --dxt (this may be more memory and compute intensive) to ..."

For this log we have the following record counts:
```
DXT_POSIX: 624
DXT_MPIIO: 1536
```
`DXT_POSIX` is actually the problematic heatmap here, but I don't think we would know that until we collected all of the DXT segments (using `get_aggregate_data()`):
```
DXT_POSIX agg_df.shape: (620996, 4)
DXT_MPIIO agg_df.shape: (110796, 4)
```

We could set a threshold based on the shape of this dataframe I think, or possibly further upstream, but I'm not aware of a way to detect the number of DXT segments without running through a lot of the DXT heatmap code. I think it would be doable though, just not as simple as checking the number of DXT records.

nawtrey:
> Ok, I revised the PR to narrow the `xfail` instead of eliminate it--this branch runs 4 extra tests vs. `pydarshan-devel`.

I opened gh-692 to address the remaining issues here. I checked this branch locally (`329 passed, 2 xfailed in 136.33s`) and the CI is passing, so LGTM.

carns:
I was curious about this e3sm example log; here is what it reports for the module sizes:
```
# log file regions
# -------------------------------------------------------
# header: 360 bytes (uncompressed)
# job data: 562 bytes (compressed)
# record table: 258 bytes (compressed)
# POSIX module: 20609 bytes (compressed), ver=4
# MPI-IO module: 510 bytes (compressed), ver=3
# PNETCDF module: 114 bytes (compressed), ver=2
# LUSTRE module: 9878 bytes (compressed), ver=1
# STDIO module: 86 bytes (compressed), ver=2
# DXT_POSIX module: 8864076 bytes (compressed), ver=1
# DXT_MPIIO module: 1820298 bytes (compressed), ver=2
# APMPI module: 139307 bytes (compressed), ver=1
# HEATMAP module: 315930 bytes (compressed), ver=1
```
So the compressed trace version of the DXT POSIX and MPIIO modules are about 8.5 and 1.5 MiB respectively.  If we wanted a rough heuristic for when to disable automatic processing of DXT data, maybe just disable if either module is above 2 MiB?

That's not perfect (maybe some near 2 MiB will take a long time, and maybe some beyond that point are fine).  That would mainly be a safety check to try to keep the analysis from bogging down for an end user unless they explicitly ask for it.

Issue #652, opened 2022-02-09
integrated runtime heatmap data into summary heatmap figure


carns:
When DXT is not present, use the DXT heatmap module data to recreate figure.

- [x] new logs repo example that includes heatmap module data (from an actual run)
- [x] new logs repo example that has both heatmap and dxt data for sanity check
- [x] bindings for heatmap module
- [x] summary code to use heatmap data

carns:
We can knock out the first two checkboxes on this with the same run.  @shanedsnyder is looking into running an E3SM benchmark example on Cori for this purpose.

shanedsnyder:
See https://github.com/darshan-hpc/darshan-logs/pull/22 and https://github.com/darshan-hpc/darshan-logs/pull/23 for example logs for this work. The former is a smaller scale example that could be used as a starting point that is easy to visually inspect, the latter is a pretty large-scale stress test once we're confident with the first.

nawtrey:
With gh-615 merged I checked off the bindings.

shanedsnyder:
Closed via #715

PR #696, opened 2022-03-17
MAINT: Use runtime for DXT heatmap x-axis scaling

nawtrey:
### Description

* Update `plot_heatmap()` to use the runtime for scaling x-axis

* Add `tmax` parameter to `get_heatmap_df` to allow for setting
the heatmap maximum time. Default behavior is to use the final
DXT segment end time.

* Add logic to `plot_heatmap` to prevent truncation of data
when using the job runtime for setting plotting boundaries

* Change `set_x_axis_ticks_and_labels()` and
`get_x_axis_tick_labels()` to use runtime for
x-axis labels

* Update `set_x_axis_ticks_and_labels()` parameters and expected
xticklabels in `test_set_x_axis_ticks_and_labels()`

* Contributes to x-axis rescaling portion of issue https://github.com/darshan-hpc/darshan/issues/575

### Comments
This gets us closer to resolving gh-652 because it removes the dependence on `get_aggregate_data()`, at least when it comes to the x-axis boundary setting. This should allow for an alternate code path that can generate the runtime heatmap without creating an aggregate dataframe.

There are a lot of changes to the values in some of the regression tests, but the heatmap dataframe sums remained the same, which gave me some confidence. We can probably clean up `test_runtime_dxt_heatmap_similarity()` further since the runtime/DXT heatmaps should align much closer now. I took a quick look and the resulting dataframes for that test are now nearly identical, with the exception of the column names having slightly different values (seems to be floating point related).

For the benchmarks, running `asv run --show-stderr` locally, the only errors I see are the (expected) `KeyError`s, so those look good.

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -273,7 +273,7 @@ def get_aggregate_data(
     return agg_df
 
 
-def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFrame:
+def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int, tmax: float = None) -> pd.DataFrame:
     """
     Builds an array similar to a 2D-histogram, where the y data is the unique
     ranks and the x data is time. Each bin is populated with the data sum
@@ -289,6 +289,8 @@ def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFram
     xbins: the number of x-axis bins to create.
 
     nprocs: the number of MPI ranks/processes used at runtime.
+
+    tmax: the maximum time for the heatmap interval bins.
 
     Returns
     -------
@@ -310,10 +312,12 @@ def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFram
         3                   1.048576e+06
 
     """
+    if not tmax:
+        # use final DXT segment end time
+        tmax = float(agg_df["end_time"].max())
     # generate the bin edges by generating an array of length n_bins+1, then
     # taking pairs of data points as the min/max bin value
-    max_time = agg_df["end_time"].max()
-    bin_edge_data = np.linspace(0.0, max_time, xbins + 1)
+    bin_edge_data = np.linspace(0.0, tmax, xbins + 1)
     # create dummy variables for start/end time data, where dataframe columns
     # are the x-axis bin ranges
     cats_start = pd.get_dummies(
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -273,7 +273,7 @@ def get_aggregate_data(
     return agg_df
 
 
-def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int, tmax: float = None) -> pd.DataFrame:
+def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFrame:
     """
     Builds an array similar to a 2D-histogram, where the y data is the unique
     ranks and the x data is time. Each bin is populated with the data sum
@@ -289,8 +289,6 @@ def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int, tmax: float = 
     xbins: the number of x-axis bins to create.
 
     nprocs: the number of MPI ranks/processes used at runtime.
-
-    tmax: the maximum time for the heatmap interval bins.
 
     Returns
     -------
@@ -312,12 +310,10 @@ def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int, tmax: float = 
         3                   1.048576e+06
 
     """
-    if not tmax:
-        # use final DXT segment end time
-        tmax = float(agg_df["end_time"].max())
     # generate the bin edges by generating an array of length n_bins+1, then
     # taking pairs of data points as the min/max bin value
-    bin_edge_data = np.linspace(0.0, tmax, xbins + 1)
+    max_time = agg_df["end_time"].max()
+    bin_edge_data = np.linspace(0.0, max_time, xbins + 1)
     # create dummy variables for start/end time data, where dataframe columns
     # are the x-axis bin ranges
     cats_start = pd.get_dummies(
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
@@ -63,7 +63,7 @@ def get_x_axis_ticks(ax: Any, n_xlabels: int = 4) -> npt.NDArray[np.float64]:
 
 
 def get_x_axis_tick_labels(
-    agg_df: pd.DataFrame, n_xlabels: int = 4
+    max_time: float, n_xlabels: int = 4
 ) -> Union[npt.NDArray[np.float64], npt.NDArray[np.intc]]:
     """
     Creates the x-axis tick mark labels.
@@ -71,8 +71,7 @@ def get_x_axis_tick_labels(
     Parameters
     ----------
 
-    agg_df: a ``pd.DataFrame`` containing the aggregated data determined
-    by the input modules and operations.
+    max_time: the maximum time to plot.
 
     n_xlabels: the number of x-axis tick marks to create. Default is 4.
 
@@ -82,7 +81,6 @@ def get_x_axis_tick_labels(
     x_ticklabels: array of x-axis tick mark labels of length ``n_xlabels``.
 
     """
-    max_time = agg_df["end_time"].values.max()
     # for the x tick labels, start at 0 and end with
     # the max time (converted to an integer)
     if max_time <= 1:
@@ -187,7 +185,7 @@ def get_y_axis_tick_labels(ax: Any, n_ylabels: int = 6) -> npt.NDArray[np.intc]:
 
 
 def set_x_axis_ticks_and_labels(
-    jointgrid: Any, agg_df: pd.DataFrame, n_xlabels: int = 4
+    jointgrid: Any, tmax: float, n_xlabels: int = 4
 ):
     """
     Sets the x-axis tick mark locations and labels.
@@ -197,15 +195,14 @@ def set_x_axis_ticks_and_labels(
 
     jointgrid: a ``sns.axisgrid.JointGrid`` object.
 
-    agg_df: a ``pd.DataFrame`` containing the aggregated data determined
-    by the input modules and operations.
+    tmax: the maximum time to plot.
 
     n_xlabels: the number of x-axis tick marks to create. Default is 4.
 
     """
     # retrieve the x-axis tick mark locations and labels
     xticks = get_x_axis_ticks(ax=jointgrid.ax_joint, n_xlabels=n_xlabels)
-    xticklabels = get_x_axis_tick_labels(agg_df=agg_df, n_xlabels=n_xlabels)
+    xticklabels = get_x_axis_tick_labels(max_time=tmax, n_xlabels=n_xlabels)
     # set the x-axis ticks and labels
     jointgrid.ax_joint.set_xticks(xticks)
     jointgrid.ax_joint.set_xticklabels(xticklabels, minor=False)
@@ -396,8 +393,24 @@ def plot_heatmap(
         align="edge",
     )
 
-    # set the x and y tick locations and labels
-    set_x_axis_ticks_and_labels(jointgrid=jgrid, agg_df=agg_df, n_xlabels=4)
+    # retrieve the final DXT segment end time
+    tmax_dxt = float(agg_df["end_time"].max())
+    # calculate the elapsed runtime
+    runtime = report.metadata["job"]["end_time"] - report.metadata["job"]["start_time"]
+    # ensure a minimum runtime value of 1 second
+    runtime = max(runtime, 1)
+    # if the data max time exceeds the runtime, buffer by 1 second
+    # until our timer precision improves to prevent truncation
+    if tmax_dxt > runtime:
+        runtime += 1
+    # scale the x-axis to span the calculated run time
+    scale_factor = runtime / tmax_dxt
+    jgrid.ax_joint.set_xlim(0.0, np.ceil(xbins * scale_factor))
+    # rescale x-ticks so x-ticks are adjusted properly downstream
+    jgrid.ax_joint.set_xticks(scale_factor * np.asarray(jgrid.ax_joint.get_xticks()))
+
+    # set the x and y tick locations and labels using the runtime
+    set_x_axis_ticks_and_labels(jointgrid=jgrid, tmax=runtime, n_xlabels=4)
     set_y_axis_ticks_and_labels(jointgrid=jgrid, n_ylabels=6)
 
     # cleanup the marginal bar graph ticks and tick labels
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
@@ -28,38 +28,24 @@ import darshan
 from darshan.experimental.plots import heatmap_handling
 
 
-def get_x_axis_ticks(ax: Any, n_xlabels: int = 4) -> npt.NDArray[np.float64]:
+def get_x_axis_ticks(bin_max: float, n_xlabels: int = 4) -> npt.NDArray[np.float64]:
     """
     Creates the x-axis tick mark locations.
 
     Parameters
     ----------
 
-    ax: a ``matplotlib`` axis object.
+    bin_max: the maximum number of bins.
 
     n_xlabels: the number of x-axis tick marks to create. Default is 4.
 
     Returns
     -------
 
-    xticks: array of x-axis tick mark locations of length ``n_xlabels``.
+    Array of x-axis tick mark locations of length ``n_xlabels``.
 
     """
-    # get the original x-axis tick locations
-    initial_xticks = ax.get_xticks()
-    if len(initial_xticks) < n_xlabels:
-        # if there are too few initial x-axis tick locations, generate
-        # a new array with 0 as the minimum and the same maximum as
-        # the original x-axis ticks
-        xticks = np.linspace(0, np.max(initial_xticks), n_xlabels)
-    else:
-        # use the original tick marks to make new arrays that
-        # contain a subset of the original ticks/labels
-        tick_idx = np.round(np.linspace(0, initial_xticks.size - 1, n_xlabels)).astype(
-            int
-        )
-        xticks = initial_xticks[tick_idx]
-    return xticks
+    return np.linspace(0, bin_max, n_xlabels)
 
 
 def get_x_axis_tick_labels(
@@ -185,7 +171,7 @@ def get_y_axis_tick_labels(ax: Any, n_ylabels: int = 6) -> npt.NDArray[np.intc]:
 
 
 def set_x_axis_ticks_and_labels(
-    jointgrid: Any, tmax: float, n_xlabels: int = 4
+    jointgrid: Any, tmax: float, bin_max: float, n_xlabels: int = 4
 ):
     """
     Sets the x-axis tick mark locations and labels.
@@ -197,11 +183,13 @@ def set_x_axis_ticks_and_labels(
 
     tmax: the maximum time to plot.
 
+    bin_max: the maximum number of bins.
+
     n_xlabels: the number of x-axis tick marks to create. Default is 4.
 
     """
     # retrieve the x-axis tick mark locations and labels
-    xticks = get_x_axis_ticks(ax=jointgrid.ax_joint, n_xlabels=n_xlabels)
+    xticks = get_x_axis_ticks(bin_max=bin_max, n_xlabels=n_xlabels)
     xticklabels = get_x_axis_tick_labels(max_time=tmax, n_xlabels=n_xlabels)
     # set the x-axis ticks and labels
     jointgrid.ax_joint.set_xticks(xticks)
@@ -404,13 +392,10 @@ def plot_heatmap(
     if tmax_dxt > runtime:
         runtime += 1
     # scale the x-axis to span the calculated run time
-    scale_factor = runtime / tmax_dxt
-    jgrid.ax_joint.set_xlim(0.0, np.ceil(xbins * scale_factor))
-    # rescale x-ticks so x-ticks are adjusted properly downstream
-    jgrid.ax_joint.set_xticks(scale_factor * np.asarray(jgrid.ax_joint.get_xticks()))
-
+    xbin_max = xbins * (runtime / tmax_dxt)
+    jgrid.ax_joint.set_xlim(0.0, xbin_max)
     # set the x and y tick locations and labels using the runtime
-    set_x_axis_ticks_and_labels(jointgrid=jgrid, tmax=runtime, n_xlabels=4)
+    set_x_axis_ticks_and_labels(jointgrid=jgrid, tmax=runtime, bin_max=xbin_max, n_xlabels=4)
     set_y_axis_ticks_and_labels(jointgrid=jgrid, n_ylabels=6)
 
     # cleanup the marginal bar graph ticks and tick labels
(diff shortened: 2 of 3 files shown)


nawtrey:
I think this is still closer to what we want, but there are still some issues with setting the heatmap maximum time value based on the integer runtime. For example, for `runtime_and_dxt_heatmaps_diagonal_write_only.darshan` I had to change the value to 3.4s manually (instead of the runtime of 3s) to get the full dataframe plotted. Otherwise we miss 2-3 of the ranks we are looking for. 

Here's what this branch generates in the summary report for the aforementioned log:
runtime_and_dxt_heatmaps_diagonal_write_only

As you can see, the x-axis is set based off of the runtime, but as a result we have sliced out the data for the highest 2-3 ranks. I'm guessing this is because the runtime doesn't get rounded up to 4 seconds unless it hits 3.5s, which we barely miss for this case, but it's something we should keep in mind I think. Especially when our testing logs tend to have shorter runtimes.

darshan-util/pydarshan/darshan/tests/test_report.py
```diff
@@ -441,7 +441,7 @@ def test_runtime_dxt_heatmap_similarity():
 
     # DXT heatmap:
     agg_df = heatmap_handling.get_aggregate_data(report=report, ops=["write"])
-    dxt_heatmap_df = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=32, nprocs=32)
+    dxt_heatmap_df = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=34, nprocs=32, tmax=3.4)
```
tylerjereddy, line 444: Why are we feeding in a magic number like `3.4` in here? It should always be possible to conveniently calculate values like this programmatically--in this case, if the default time precision isn't good enough, just round it up to prevent truncating the data unnecessarily, and apply the same policy globally (pad a 1 second resolution timer by + 1 second). This isn't just some weird test case--in theory, there could be an infinite number of logs that would have truncated data if we don't round up.

Honestly, I'm not a huge fan of discarding the time resolution DXT gives us instead of devising an approach to use the best resolution we have available with the heatmaps that a given report contains, i.e., the max time across all the different heatmap dataframes for example. Maybe it could later be relaxed when the runtime resolution improves.

tylerjereddy (commented):
Not sure I understand why we would intentionally truncate perfectly valid data like the example plotted, apart from a superficial concern about whitespace. I think I still have a pretty strong preference for a carefully-crafted approach that takes the max time from all heatmap dataframes selected for a given summary report. Why aren't we doing that instead? I can see how it would require more refactoring I suppose. Nonetheless, intentionally truncating data is a bit of a strange design instead of just adding a buffer until time resolution increases?

We can easily restore the truncated data here, though my preference is still to not throw away our highest time resolution data.

```diff
diff --git a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
index dc47c6d1..957641b0 100644
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
@@ -336,7 +336,9 @@ def plot_heatmap(
     # calculate the elapsed runtime
     runtime = report.metadata["job"]["end_time"] - report.metadata["job"]["start_time"]
     # ensure a minimum runtime value of 1 second
-    runtime = max(runtime, 1)
+    # for DXT heatmap we buffer by at least 1 second until
+    # our timer precision improves to prevent truncation
+    runtime = max(runtime, 1) + 1
     nprocs = report.metadata["job"]["nprocs"]
     # get the heatmap data array
     hmap_df = heatmap_handling.get_heatmap_df(agg_df=agg_df, xbins=xbins, nprocs=nprocs, tmax=runtime)

```

image

tylerjereddy:
My vision would probably be to have the time resolution calculated differently if we have:

- runtime heatmap only
- a single DXT heatmap
- multiple DXT heatmaps (logic to compare across and take greatest runtime)
- DXT and runtime heatmaps together (different logic to compare across)

If DXT tracing can run for shorter than the full tracing time, some logic for that too, since that "whitespace" would be "real inactive time" I suppose.

darshan-util/pydarshan/benchmarks/benchmarks/dxt_heatmap.py
```diff
@@ -69,8 +69,8 @@ def setup(self, unique_ranks, bin_count, density):
     def time_get_heatmap_df(self, unique_ranks, bin_count, density):
         # benchmark get_heatmap_df() handling of variable
         # numbers of unique ranks/bins
-        heatmap_handling.get_heatmap_df(self.agg_df, xbins=bin_count, nprocs=unique_ranks)
+        heatmap_handling.get_heatmap_df(self.agg_df, xbins=bin_count, nprocs=unique_ranks, tmax=1)
 
 
     def peakmem_get_heatmap_df(self, unique_ranks, bin_count, density):
-        heatmap_handling.get_heatmap_df(self.agg_df, xbins=bin_count, nprocs=unique_ranks)
+        heatmap_handling.get_heatmap_df(self.agg_df, xbins=bin_count, nprocs=unique_ranks, tmax=1)
```
tylerjereddy, line 76: same comment on the "magic number" for time here--this feels a bit fragile IMO, instead of being extracted programmatically from a report for example.

darshan-util/pydarshan/darshan/tests/test_report.py
nawtrey: Yeah I see your point. I'll update this to use the report metadata when I pad `tmax`.

darshan-util/pydarshan/benchmarks/benchmarks/dxt_heatmap.py
nawtrey: Alright, I'll can change this to use something like `agg_df["end_time"].max()`.

tylerjereddy: Hmm, maybe, I'm kind of thinking this should just happen inside `get_hetmap_df()` or something. The design my warrant some careful consideration.

If `get_heatmap_df()` is going to be used for generating runtime maps as well, it could just use all the timing information available to find an appropriate upper timing bound internally. It depends on where the decision about which heatmaps to plot is going to be made I suspect.

tylerjereddy: If the decision is only relevant in the plotting code, which it may be (?), then a function to inspect the contents of `agg_df` and `runtime_hmap_df` (when available) or whatever might make sense for setting `max_t`.

Then the plotting code retrieves the DXT/runtime datastructures with their "common denomintaor" `max_t`? I guess then I wonder what use we'd have for padding, since the data would always be bound to the observed upper limit. Can runtime heatmap final bin ever finish before the DXT max end time?

Maybe the Argonne folks have a different vision though.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -273,7 +273,7 @@ def get_aggregate_data(
     return agg_df
 
 
-def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFrame:
+def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int, tmax: float) -> pd.DataFrame:
```
tylerjereddy, line 276: Another design thought for consideration, again depending on what the Argonne folks really want to happen: what if `tmax` defaulted to `None`, and when it was `None`, we just return the usual properly-trimmed DXT/runtime heatmap without modification?

A bit like a sensible default that encapsulates the data--it may also have the bonus of reducing the number of testing changes apart from maybe adding a check to make sure the `tmax` is respected.

So, the plotting code could call with a custom `tmax` when it needs to make the heatmaps flush on end time, but `get_heatmap_df` still does something quite sensible even outside that scenario.

nawtrey: Something like this makes sense to me. Would we want to do something similar for the minimum time? We could improve the resolution that way as well.

darshan-util/pydarshan/benchmarks/benchmarks/dxt_heatmap.py
nawtrey: Yeah I think I would like to handle the heatmap retrieval (either DXT or runtime) in `get_heatmap_df()`. There is a challenge there because the heatmaps are stored under generic "POSIX" and "MPIIO" module names, but I think if we pass along the report object to `get_heatmap_df()` we can do something like the following:

```python
# retrieve DXT POSIX heatmap
hmap_df = get_heatmap_df(report, mod="DXT_POSIX", ops=["read", "write"])
# retrieve runtime POSIX heatmap
hmap_df = get_heatmap_df(report, mod="POSIX", ops=["read", "write"])
```

Passing the report object would remove the need for `nprocs` or `tmax` as formal parameters, and we can handle cases based on whether the `HEATMAP` module is present or not. We would have to move the call to `get_aggregate_data()` inside of `get_heatmap_df()` as well, but I think I like that better anyways and this makes some of the `tmax` cases [here](https://github.com/darshan-hpc/darshan/pull/696#issuecomment-1071583214) pretty easy I think.

nawtrey: After thinking about this more, I think it might be nice to have the following functions defined:
1. `get_dxt_heatmap_df(report, mod, ops, xbins, tmax=None)`: basically the current `get_heatmap_df()` function renamed, with the call to `get_aggregate_data()` moved inside the function body (it is only needed to get the DXT heatmap)
2. `get_runtime_heatmap_df(report, mod, ops, tmax=None)`: new function for handling the runtime heatmap retrieval, should be pretty simple. Maybe so simple it is unnecessary
3. `get_heatmap_tmax(report, mod)`: determines what the maximum x-axis value should be for a heatmap depending on the module input
4. `get_heatmap_df(report, mod, ops, xbins=None)`: function for handling all of the above function calls. This would be the function you would call to actually retrieve the dataframe for a given module.

With a default `tmax=None` for both DXT/runtime heatmap retrieval functions they would use their "normal" methods for calculating `tmax`, but we would have the option to override it when we wanted to (i.e. for multiple heatmap cases). For plotting in `plot_heatmap()` we can retrieve `tmax` from the column names/intervals in the returned dataframe from `get_heatmap_df()`.

This could be accomplished in a series of smaller PR's:
1.  move `get_aggregate_data()` call into `get_heatmap_df()` body, fix x-ticks functions to receive `tmax` as input (to remove dependence on `agg_df()`)
2. rename `get_heatmap_df()` to `get_dxt_heatmap_df()`
3. add `get_runtime_heatmap_df()` (maybe with a unit test)
4. add `get_heatmap_tmax()` and `get_heatmap_df()`, and generalize `plot_heatmap()` so it can plot both heatmap types. This is where we would decide how to handle setting `tmax` for the various cases and actually fix gh-575

Does this seem sensible? I would probably create a new PR and close this one since most of the changes here are likely unnecessary.  But I could handle 1., 3., and parts of 4. at the same time because they wouldn't overlap I don't think.

tylerjereddy: Hard to say, sounds like a fair bit of refactoring.

I thought `plot_heatmap()` would basically just need to retrieve some form of raw data for the runtime and DXT heatmaps to find a common `tmax`, then retrieve dataframes adjusted to that `tmax`.

You could check with the team this afternoon to make sure you're absolutely clear on the summary report requirements for the different heatmap permutation scenarios.

tylerjereddy:
I believe the suggestion from the last meeting was to try leaving the dataframes largely unchanged (confined to their natural time bounds) and to plot them with `seaborn`  using a `tmax` that is runtime + 1 second, and just living with any extra whitespace that produces. Assuming `seaborn` tolerates that reasonably well.

nawtrey:
> I believe the suggestion from the last meeting was to try leaving the dataframes largely unchanged (confined to their natural time bounds) and to plot them with `seaborn` using a `tmax` that is runtime + 1 second, and just living with any extra whitespace that produces. Assuming `seaborn` tolerates that reasonably well.

Thanks for adding the comment here, I meant to do the same. I worked on it a bit and this is definitely possible, I just have to work out some kinks with the x-axis labels. The upside is the DXT heatmap dataframes should remain unchanged, so the testing changes should be a fair bit simpler.

nawtrey:
Alright, I think this is much closer to what we discussed in our meeting last week. There is now an optional `tmax` parameter for `get_heatmap_df()` that allows for more dynamic adjustment of the heatmap (for future features, maybe) but is not leveraged in `plot_heatmap()`. The default behavior of `plot_heatmap()` is to plot the heatmap figure with x limits in the interval `[0, runtime]`, where the runtime is always greater than 1 and always greater than the final DXT segment value (to prevent truncation). The DXT dataframe itself is unchanged though.

nawtrey:
The CI is failing due to https://github.com/darshan-hpc/darshan-logs/issues/31, so when Tyler's PR get's merged we can reroll the CI here.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
```diff
@@ -396,8 +393,24 @@ def plot_heatmap(
...
-    # set the x and y tick locations and labels
-    set_x_axis_ticks_and_labels(jointgrid=jgrid, agg_df=agg_df, n_xlabels=4)
+    # retrieve the final DXT segment end time
+    tmax_dxt = float(agg_df["end_time"].max())
+    # calculate the elapsed runtime
+    runtime = report.metadata["job"]["end_time"] - report.metadata["job"]["start_time"]
+    # ensure a minimum runtime value of 1 second
+    runtime = max(runtime, 1)
+    # if the data max time exceeds the runtime, buffer by 1 second
+    # until our timer precision improves to prevent truncation
```
tylerjereddy, line 403: I thought we agreed on runtime + 1 second, ignoring DXT data and just passing through to the padded plots. I don't really care too much I don't think if it looks 'ok' when I do a more thorough review, though in cases where DXT heatmap analysis gets automatically suppressed due to size in the future it may not be sustainable.

When we only have to do deal with higher-resolution logs the issue will be less crucial perhaps, though older log files will likely be reality for some time.

darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
```diff
@@ -115,21 +93,19 @@ def test_set_x_axis_ticks_and_labels(
...
     else:
         filepath = get_log_path(filepath)
         # for all other data sets just load the data from the log file
+        # calculate the elapsed runtime
         report = darshan.DarshanReport(filepath)
-        agg_df = heatmap_handling.get_aggregate_data(
-            report=report, mod="DXT_POSIX", ops=["read", "write"]
-        )
+        runtime = report.metadata["job"]["end_time"] - report.metadata["job"]["start_time"]
+        runtime = max(runtime, 1)
```
tylerjereddy, line 104: How does this help when `runtime` was rounded down from `1.3` seconds? (i.e., is the approach used in the test the sustainable approach?). It feels like it is just kind of a fragment of the true logic, and we're actually using `tmax` for this test, but not in general? It seems confusing.

darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
```diff
@@ -115,21 +93,19 @@ def test_set_x_axis_ticks_and_labels(
     if filepath is None:
         # don't have any data sets with a max time between 1 and 10, so
         # create a synthetic one here
-        data = [[4, 1.03378843, 1.03387713, 0], [4000, 1.04216653, 1.04231459, 0]]
-        cols = ["length", "start_time", "end_time", "rank"]
-        agg_df = pd.DataFrame(data=data, columns=cols)
+        runtime = 1
```
tylerjereddy, line 96: hmm, the comment above doesn't really make sense anymore, and the basis for deleting the previous test case is?

darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
tylerjereddy: Ah, I guess the `tmax` is just for setting the ticks, though my comment about thoroughly handling the desired runtime compared to what happens in source remains here, I think.

darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
```diff
@@ -273,7 +273,7 @@ def get_aggregate_data(
     return agg_df
 
 
-def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int) -> pd.DataFrame:
+def get_heatmap_df(agg_df: pd.DataFrame, xbins: int, nprocs: int, tmax: float = None) -> pd.DataFrame:
```
tylerjereddy, line 276: I'm not sure there's much value in changing `get_heatmap_df` in this PR--there are no new tests for the new functionality and we aren't using it in this PR?

We probably also want the new argument typed as `Optional[float]` to avoid the potential future issue with:

`mypy --no-implicit-optional darshan/experimental/plots/heatmap_handling.py `

`darshan/experimental/plots/heatmap_handling.py:276: error: Incompatible default for argument "tmax" (default has type "None", argument has type "float")  [assignment]`

Though I'd probably just leave `get_heatmap_df()` alone in this PR at this point.

darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
tylerjereddy: Yeah, the old `tmax_dxt` was `1.04231459`, so by the new logic the closest thing to that old case would seem to be something like this:

```diff
@@ -93,8 +93,13 @@ def test_set_x_axis_ticks_and_labels(
     if filepath is None:
         # don't have any data sets with a max time between 1 and 10, so
         # create a synthetic one here
+        data = [[4, 1.03378843, 1.03387713, 0], [4000, 1.04216653, 1.04231459, 0]]
+        cols = ["length", "start_time", "end_time", "rank"]
+        agg_df = pd.DataFrame(data=data, columns=cols)
+        tmax_dxt = float(agg_df["end_time"].max())
         runtime = 1
-
+        if tmax_dxt > runtime:
+            runtime += 1
```

Not sure how you make that pass the test, but then this is the new logic/approach, as far as I can tell. I guess you'd need all the scaling stuff as well.

I guess my overall point is just that this testing change is a bit weird on its own, since it seems to merely allow the test to pass instead of representing the old case, which indeed appears to be a useful scenario to test (though I'm not sure things are factored correctly to test it here, or if much is gained by pasting the source of the DXT plotting code here as well).

tylerjereddy: It is probably the case that the axis scaling logic is drifting a bit from a set of unified unit-testable function(s) because of the addition of all of this stuff outside of the actual function:

```python
    # scale the x-axis to span the calculated run time
    scale_factor = runtime / tmax_dxt
    jgrid.ax_joint.set_xlim(0.0, np.ceil(xbins * scale_factor))
    # rescale x-ticks so x-ticks are adjusted properly downstream
    jgrid.ax_joint.set_xticks(scale_factor * np.asarray(jgrid.ax_joint.get_xticks()))
```

I didn't check if things get cleaner to test if that gets kind of packed into `set_x_axis_ticks_and_labels()`, though it is possible with a func signature change I suspect.

darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
nawtrey: I guess I figured since we have the option to only pad when necessary it made more sense, but I can switch it back. Either is fine with me

tylerjereddy: Yeah, probably not worth changing at the moment, experience/usage will probably guide future changes anyway.

I think my only substantive design thought at the moment is the encapsulation of some of the new logic before `set_x_axis_ticks_and_labels`. For example, there's a step before the function is called described as:

> rescale x-ticks so x-ticks are adjusted properly downstream

which might logically fit into a function like `set_x_axis_ticks_and_labels()`. But not sure if that is just me trying to get things scoped for convenient testing purposes, and if cramming the rest of that logic in the function would go too far.

darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
nawtrey: Yeah that's a good point. I'll add in the full logic.

darshan-util/pydarshan/darshan/tests/test_plot_dxt_heatmap.py
nawtrey: Iterating on this more, it seems moving the below code into `set_x_axis_ticks_and_labels()` makes the test a lot more difficult to manage. 

```python
    # rescale x-ticks so x-ticks are adjusted properly downstream
    jgrid.ax_joint.set_xticks(scale_factor * np.asarray(jgrid.ax_joint.get_xticks()))
```

Instead I think I could pass in a new formal parameter `bin_max` (`xbins * scale_factor`) and change `get_x_axis_ticks()` to use that value instead of pulling it from the axis object. The benefit to this is we no longer need the above line of code, and I think it makes the code simpler overall. The pasted diff below shows what this would look like in comparison. 

<details><summary>diff</summary>

```diff 
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py
@@ -28,7 +28,7 @@ import darshan
 from darshan.experimental.plots import heatmap_handling
 
 
-def get_x_axis_ticks(ax: Any, n_xlabels: int = 4) -> npt.NDArray[np.float64]:
+def get_x_axis_ticks(bin_max: float, n_xlabels: int = 4) -> npt.NDArray[np.float64]:
     """
     Creates the x-axis tick mark locations.
 
@@ -42,24 +42,10 @@ def get_x_axis_ticks(ax: Any, n_xlabels: int = 4) -> npt.NDArray[np.float64]:
 
     """
-    # get the original x-axis tick locations
-    initial_xticks = ax.get_xticks()
-    if len(initial_xticks) < n_xlabels:
-        # if there are too few initial x-axis tick locations, generate
-        # a new array with 0 as the minimum and the same maximum as
-        # the original x-axis ticks
-        xticks = np.linspace(0, np.max(initial_xticks), n_xlabels)
-    else:
-        # use the original tick marks to make new arrays that
-        # contain a subset of the original ticks/labels
-        tick_idx = np.round(np.linspace(0, initial_xticks.size - 1, n_xlabels)).astype(
-            int
-        )
-        xticks = initial_xticks[tick_idx]
-    return xticks
+    return np.linspace(0, bin_max, n_xlabels)
 
 
 def get_x_axis_tick_labels(
@@ -185,7 +171,7 @@ def get_y_axis_tick_labels(ax: Any, n_ylabels: int = 6) -> npt.NDArray[np.intc]:
 
 
 def set_x_axis_ticks_and_labels(
-    jointgrid: Any, tmax: float, n_xlabels: int = 4
+    jointgrid: Any, tmax: float, bin_max: float, n_xlabels: int = 4
 ):
     """
     Sets the x-axis tick mark locations and labels.
@@ -201,7 +187,7 @@ def set_x_axis_ticks_and_labels(
 
     """
     # retrieve the x-axis tick mark locations and labels
-    xticks = get_x_axis_ticks(ax=jointgrid.ax_joint, n_xlabels=n_xlabels)
+    xticks = get_x_axis_ticks(bin_max=bin_max, n_xlabels=n_xlabels)
     xticklabels = get_x_axis_tick_labels(max_time=tmax, n_xlabels=n_xlabels)
     # set the x-axis ticks and labels
     jointgrid.ax_joint.set_xticks(xticks)
@@ -404,13 +390,10 @@ def plot_heatmap(
     if tmax_dxt > runtime:
         runtime += 1
     # scale the x-axis to span the calculated run time
-    scale_factor = runtime / tmax_dxt
-    jgrid.ax_joint.set_xlim(0.0, np.ceil(xbins * scale_factor))
-    # rescale x-ticks so x-ticks are adjusted properly downstream
-    jgrid.ax_joint.set_xticks(scale_factor * np.asarray(jgrid.ax_joint.get_xticks()))
-
+    xbin_max = xbins * (runtime / tmax_dxt)
+    jgrid.ax_joint.set_xlim(0.0, xbin_max)
     # set the x and y tick locations and labels using the runtime
-    set_x_axis_ticks_and_labels(jointgrid=jgrid, tmax=runtime, n_xlabels=4)
+    set_x_axis_ticks_and_labels(jointgrid=jgrid, tmax=runtime, bin_max=xbin_max, n_xlabels=4)
     set_y_axis_ticks_and_labels(jointgrid=jgrid, n_ylabels=6)
````

</details>

If this seems "okay" I'll go ahead and fix the tests that this breaks, and push it in.

tylerjereddy:
Alright, I checked a few examples locally on this branch, let's give this a try, thanks.

I do have some ideas for testing the final report in ways that are sensitive to this, but they are out of scope for this PR.

Issue #697, opened 2022-03-18
ENH: multiple operation aggregation for `Heatmap.to_df()` method


nawtrey:
The current method for retrieving the runtime IO heatmap with both "read" and "write" data is to run something like the following:

```python
rd_df = report.heatmaps["POSIX"].to_df(op="read")
wr_df = report.heatmaps["POSIX"].to_df(op="write")
rd_wr_df = rd_df + wr_df
```

I think it would be nice if we could input multiple operations and retrieve the summed dataframe directly from the `to_df()` method instead of handling it downstream. So the above dataframe could be generated with just:

```python
rd_wr_df = report.heatmaps["POSIX"].to_df(ops=["read", "write"])
```

This is how it is handled in `plot_heatmap()` and `get_aggregate_data()`, so I think it would be good to keep them consistent.

tylerjereddy:
Fixed by gh-702

PR #702, opened 2022-03-31
ENH: Runtime heatmap multiple operation support

nawtrey:
* Add multiple operation (i.e. read and write) data aggregation
in `Heatmap.to_df()` method. This more closely reflects the way
other heatmap-related functions operate which should allow for 
a more seamless integration of the runtime heatmap into 
`plot_heatmap()`.

* Correct call signatures for heatmap `to_df()` method

* Add test `test_heatmap_operations` for checking different
operation inputs and comparing their results

* Fix issue #697

### DIFF
--- a/darshan-util/pydarshan/darshan/datatypes/heatmap.py
+++ b/darshan-util/pydarshan/darshan/datatypes/heatmap.py
@@ -1,3 +1,6 @@
+import functools
+from typing import Sequence
+
 import numpy as np
 import pandas as pd
 
@@ -53,7 +56,7 @@ class Heatmap():
             print()
             for op in ['read', 'write']:
                 print(op)
-                plt.pcolor(self.to_df(op=op))
+                plt.pcolor(self.to_df(ops=[op]))
                 plt.show()
 
     def add_record(self, rec: dict):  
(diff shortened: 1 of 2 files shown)


darshan-util/pydarshan/darshan/datatypes/heatmap.py
```diff
@@ -113,9 +117,15 @@ def to_df(self, op: str, interval_index: bool = True):
...
+        df_list = []
+        for op in ops:
+            data = self._data[op]
+            df = pd.DataFrame.from_dict(data, orient='index', columns=columns)
+            df.index.name = "rank"
+            df_list.append(df)
+
+        # sum the dataframes
+        hmap_df = functools.reduce(lambda x, y: x.add(y), df_list)
+        return hmap_df
```
tylerjereddy, line 130: This `functools` business seems like a rather complicated way to write `sum(df_list)`. This diff passes the tests and is  more readable I'd say:

```diff
index 580cf17d..79cfc3af 100644
--- a/darshan-util/pydarshan/darshan/datatypes/heatmap.py
+++ b/darshan-util/pydarshan/darshan/datatypes/heatmap.py
@@ -1,4 +1,3 @@
-import functools
 from typing import Sequence
 
 import numpy as np
@@ -126,6 +125,5 @@ class Heatmap():
             df_list.append(df)
 
         # sum the dataframes
-        hmap_df = functools.reduce(lambda x, y: x.add(y), df_list)
-        return hmap_df
+        return sum(df_list)

```

tylerjereddy (commented):
One of the CI-reported gaps in test coverage is worth adding a test for (below); the other case is a plotting functionality that seems less relevant for now so that coverage can be omitted I think. I'll try to push in a test for the relevant scenario.

image

The general idea seems "ok" to me, I don't think we gain too much by doing this but doesn't hurt for the consistency I suppose. I'll probably just address my own comments here since Nik will be busy.

nawtrey:
> I'll probably just address my own comments here since Nik will be busy.

I can push in a test and the simplification here today.

nawtrey:
Alright, I believe the reviewer comments have been addressed here. Locally that line is now covered:

```
Name                                                                                  Stmts   Miss  Cover   Missing
-------------------------------------------------------------------------------------------------------------------
darshan/datatypes/heatmap.py                                                             64     10    84%   44, 54-59, 78, 83, 118
```

tylerjereddy:
LGTM now, merging, thanks

PR #704, opened 2022-04-06
ENH: Add HDF5 operation counts figures to summary report

nawtrey:
* Add `H5D` or `H5F` to operation count figure modules.
Only 1 module is added since `H5D` figure encompasses
both HDF5 modules' data.

* Correct expected figure count in `test_main_without_args`
for `ior_hdf5_example.darshan` case since `H5D` section
now contains an operation counts figure.

* Contribute to issue #663

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -398,6 +398,15 @@ class ReportData:
         ################################
         ## Per-Module Statistics
         ################################
+
+        # for the operation counts, since the `H5D` variant contains
+        # both modules' data, we either want `H5F` or `H5D`, not both
+        opcounts_mods = ["POSIX", "MPI-IO", "STDIO"]
+        if "H5D" in self.report.modules:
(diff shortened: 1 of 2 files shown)


nawtrey:
This just adds an operation counts figure if one or both of the HDF5 modules are present, I'll open a separate PR to change the section titles to always read `Per-Module Statistics: HDF5` for any `H5F` or `H5D` figures. It may cause some contention for cases where the specific HDF5 API is not listed, but I'll discuss that there I think.

tylerjereddy (commented):
Seems ok, so I'll put it in.

I do think the design should be more symmetric long-term though. For `H5F` only I think we should just add the empty parts for `H5D`--i.e., have a single set of named API categories that are either filled or not for `HDF5` combined, rather than having the "lonely `H5F`" design.

`hdf5_file_opens_only.darshan` 
image

vs. `hdf5_diagonal_write_half_ranks_dxt.darshan`

image

Seems less confusing for comparing reports where something may have gone wrong with HDF5 API (which we know can happen with `h5py`, etc.).

PR #705, opened 2022-04-07
PyDarshan: Mitigate build errors from dependencies and speedup wheel building

jakobluettgau:
This pull request changes the wheel building procedure by not installing pydarshan dependencies to the manylinux docker container with the following effects:

* This aviods running into build error because of dependencies changing their build procedure (an alternative could be to pin to a working set of dependencies)
* The build is significantly faster as various dependencies would attempt to build from source
* But, we can not run tests directly after building (but we did not do this so far anyways)

I also made two small changes to the Makefile:
* `make wheels` now automatically issues add-modules
* the add-modules target performs `git submodule init `and `git submoudle update`

### DIFF
--- a/darshan-util/pydarshan/devel/build-wheels.sh
+++ b/darshan-util/pydarshan/devel/build-wheels.sh
@@ -27,16 +27,26 @@ make distclean
 cd /
 
 
-# Do not build for end-of-life python versions
-rm -f /opt/python/cp27-cp27m
-rm -f /opt/python/cp27-cp27mu
-rm -f /opt/python/cp35-cp35m
+# Uncomment any of the following lines to exclude python variant from build process
+#rm -f /opt/python/cp36-cp36m
+#rm -f /opt/python/cp37-cp37m
+#rm -f /opt/python/cp38-cp38
+#rm -f /opt/python/cp39-cp39
+#rm -f /opt/python/cp310-cp310
+#
+#rm -f /opt/python/pp37-pypy37_pp73
+#rm -f /opt/python/pp38-pypy38_pp73
+#rm -f /opt/python/pp39-pypy39_pp73
+
+
 ls /opt/python
 
 
 # Compile wheels
 for PYBIN in /opt/python/*/bin; do
-    "${PYBIN}/pip" install -r /io/requirements_dev.txt
+    # JL: we do not really need any dependencies to build the wheel,
+    #     but requirements install needs to be renabled when testing automatically
+    #"${PYBIN}/pip" install -r /io/requirements_wheels.txt	                    
     "${PYBIN}/pip" wheel /io/ --build-option "--with-extension" --no-deps -w /io/wheelhouse/${PLAT}
 done
 
--- a/darshan-util/pydarshan/Makefile
+++ b/darshan-util/pydarshan/Makefile
@@ -72,11 +72,13 @@ docs-show:
 
 # add AutoPef Python packages...
 add-modules:
+	git submodule init
+	git submodule update
 	for file in `find ../../modules/ -name "*-backend.py"` ; do \
 		cp $$file darshan/backend/`basename $$file | cut -d '-' -f 1`.py ; \
 	done
 
-wheels:
+wheels: add-modules
 	./devel/build-all-platforms.sh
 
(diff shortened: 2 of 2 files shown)


tylerjereddy (commented):
Sounds good to me if it helps the wheel build process for the near-term release. Obviously, we should try to test the wheels somehow as a general policy--I'm hoping this all becomes much less burdensome for us if we switch to [`cibuildwheel`](https://cibuildwheel.readthedocs.io/en/stable/) in the future, but that's out of scope for now.

darshan-util/pydarshan/devel/build-wheels.sh
```diff
@@ -27,16 +27,26 @@ make distclean
...
+
 ls /opt/python
 
 
 # Compile wheels
 for PYBIN in /opt/python/*/bin; do
-    "${PYBIN}/pip" install -r /io/requirements_dev.txt
+    # JL: we do not really need any dependencies to build the wheel,
+    #     but requirements install needs to be renabled when testing automatically
+    #"${PYBIN}/pip" install -r /io/requirements_wheels.txt
```
shanedsnyder, line 49: Should this just be `requirements.txt`? Or is there a `requirements_wheels.txt` that needs to be committed? The former seems to work fine for me.

shanedsnyder:
Thanks! This looks to be working for me.

Before merging, what do you think about the following modifications I made to force pytest to run for each python version we build wheels for?

```
diff --git a/darshan-util/pydarshan/devel/build-wheels.sh b/darshan-util/pydarshan/devel/build-wheels.sh
index 02220e7..d1a48ec 100755
--- a/darshan-util/pydarshan/devel/build-wheels.sh
+++ b/darshan-util/pydarshan/devel/build-wheels.sh
@@ -44,9 +44,6 @@ ls /opt/python

 # Compile wheels
 for PYBIN in /opt/python/*/bin; do
-    # JL: we do not really need any dependencies to build the wheel,
-    #     but requirements install needs to be renabled when testing automatically
-    #"${PYBIN}/pip" install -r /io/requirements_wheels.txt                     
     "${PYBIN}/pip" wheel /io/ --build-option "--with-extension" --no-deps -w /io/wheelhouse/${PLAT}
 done

@@ -55,8 +52,13 @@ for whl in /io/wheelhouse/${PLAT}/*.whl; do
     repair_wheel "$whl"
 done

-## Install packages and test
-#for PYBIN in /opt/python/*/bin/; do
-#    "${PYBIN}/pip" install darshan --no-index -f /io/wheelhouse/${PLAT}
-#    (cd "$HOME"; "${PYBIN}/nosetests" darshan)
-#done
+# Install packages and test
+cd /io/
+for PYBIN in /opt/python/*/bin/; do
+    # JL: we do not really need any dependencies to build the wheel,
+    #     but requirements install needs to be renabled when testing automatically
+    "${PYBIN}/pip" install -r /io/requirements.txt
+    "${PYBIN}/pip" install pytest
+    "${PYBIN}/pip" install darshan --no-index -f /io/wheelhouse/${PLAT}
+    "${PYBIN}/pytest" --pyargs darshan ./io/
+done
```

I basically shifted the pip install of requirements down into a separate for loop that also installs the wheels and runs pytest against them. I opted to not comment this out for now -- I assume this wheel building process is done mostly for release purposes, in which case I'm fine with it taking a lot longer to complete for the assurance that things don't break on any python versions. We can comment it out in our local repos if we have need to build wheels with a quicker turnaround. What do you think? If it looks okay, would you mind committing and then I can merge this all in and move forward with the 3.3.1.1 release.

shanedsnyder:
Not sure why I don't hit the original build errors related to the rust compiler anymore with these updated scripts with the patch applied from above -- it looks to be doing roughly the same thing it was originally doing. If for some reason you are still running into those errors, then maybe it would be best to comment out that for loop that does the pip install of the wheels and runs pytest against them? Whatever you think is best.

shanedsnyder (approved):

shanedsnyder:
I'm going to go ahead and merge the changes in this PR since they appear to be building us functional wheels again.

We can defer the addition of `pytest` tests for each wheel to a separate PR.

Issue #706, opened 2022-04-07
MAINT, TST: HDF5 report adjustments


tylerjereddy:
Two suggestions from previous comments:

- have the operation counts for the combined `HDF5` category always include both `H5D` and `H5F` "columns" to make it easier to compare cases where only `H5F` is present, even if that's rare; showing empty/inactive `H5D` categories also seems more "explicit" to me, rather than expecting the user to understand that `H5D` was inactive -- see: https://github.com/darshan-hpc/darshan/pull/704#pullrequestreview-934388709
- actually leverage some of the "ground truth" HDF5 test files added -- see: https://github.com/darshan-hpc/darshan/pull/685#issuecomment-1083398514

nawtrey:
As far as the asymmetry between `H5F` and `H5F+H5D` cases, I would suggest a slightly different approach. For the `H5F` only case, I think we should change the column names to match (e.g. `H5F Open`, `H5F Flush`) to avoid confusion, but I _don't_ think we should include the `H5D` counters if it was never used. I think it helps distinguish between counters that happen to be zero and counters that are zero because they were never a possibility. 

I guess if a user was _really_ interested in why their `H5D` counters were all zero, they could check to see if the `H5D` module is present at the top of the report.. so I guess I could go either way on this. I think either solution is better than the current situation anyways.

nawtrey:
I'll try to address the second bullet point today.

tylerjereddy:
>  I think it helps distinguish between counters that happen to be zero and counters that are zero because they were never a possibility.

For me it is a bit strange to combine `H5D` and `H5F` under one `HDF5` heading, then only show one of them because the other has zero recorded activity. If I'm not familiar with the HDF5 API structure, I may want to, at a glance, see that a file was opened but never written to, and to clearly see that darshan runtime is capable of tracking the `HDF5` writing operations.

nawtrey:
The team seems to favor Tyler's original suggestion, so I'll try to get that done today as well.

tylerjereddy:
Fixed by gh-714

PR #707, opened 2022-04-07
MAINT: Change `H5D` and `H5F` figure section titles to HDF5

nawtrey:
* Change the per-module report section titles to
`Per-Module Statistics: HDF5` for `H5F` and `H5D` figures

* Fix issue #663

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -408,6 +408,12 @@ class ReportData:
             opcounts_mods.append("H5F")
 
         for mod in self.report.modules:
+
+            if "H5" in mod:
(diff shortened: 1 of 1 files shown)


nawtrey:
I think with this change we can close issue gh-663, but I think it makes the need for gh-706 more apparent, specifically for the `H5F` case. 

Here's an example with the new title:
Screen Shot 2022-04-07 at 11 12 05 AM

I think I'll try to address that sooner than later so it is more clear what the `Open` and `Flush` columns represent for that case.

carns:
This PR looks good to me.

tylerjereddy (approved):
Ok, I checked a few logs locally on this branch, patch diff coverage is 100 %, and Phil seems happy too, so merging.

PR #711, opened 2022-04-07
MAINT: Include `H5D` columns in `H5F`-only operation counts figure

nawtrey:
Description
--------------

* Change operation counts figure to always list columns for
both `H5F` and `H5D` modules. When `H5D` module data is not
present the values are set to zero.

* Contribute to issue #706

Example
----------
Here is the updated operation counts figure for `hdf5_file_opens_only.darshan`:
image


carns:
This looks good.  I like the label names we've arrived at too.

tylerjereddy (approved):
I checked the rebased version of this feature branch locally, patch diff is 100 %, and Phil seems happy, so merging thanks.

Issue #717, opened 2022-04-20
BUG: ReportFigure ignores fig_width for large figs


tylerjereddy:
`ReportFigure` has a `fig_width` argument that is documented as:

> fig_width : the width of the figure in pixels. 

Let's try adding in the "data access by category" figure to the report and setting its width to a single pixel:

```diff
diff --git a/darshan-util/pydarshan/darshan/cli/summary.py b/darshan-util/pydarshan/darshan/cli/summary.py
index 7fbc8fad..6b64069d 100644
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -22,6 +22,7 @@ from darshan.experimental.plots import (
     plot_common_access_table,
     plot_access_histogram,
     plot_opcounts,
+    data_access_by_filesystem,
 )
 
 darshan.enable_experimental()
@@ -458,6 +459,20 @@ class ReportData:
                 )
                 self.figures.append(opcount_fig)
 
+        #########################
+        # Data Access by Category
+        if not {"POSIX", "STDIO"}.isdisjoint(set(self.report.modules)):
+            data_access_by_cat_fig = ReportFigure(
+                section_title="Data Access by Category",
+                fig_title="",
+                fig_func=data_access_by_filesystem.plot_with_report,
+                fig_args=dict(report=self.report, num_cats=8),
+                fig_description="",
+                fig_width=1,
+            )
+            self.figures.append(data_access_by_cat_fig)
+
+
```

The result is still way bigger than any other figure/image in the html file generated with `python -m darshan summary /Users/treddy/github_projects/darshan-logs/darshan_logs/e3sm_io_heatmaps_and_dxt/e3sm_io_heatmap_only.darshan`:

![image](https://user-images.githubusercontent.com/7903078/164276632-a0202322-83db-4495-a686-519024000e6b.png)

nawtrey:
It appears this only happens when `fig_title=""`. If you populate the field `fig_width` should work again. I think it's because we are using the figure title as the alternate text in the `<img>` tag, so when it isn't present we get something like this:

```
<img src=data:ima...= alt= width=350>
```

which must be preventing the width from being set properly. Strictly speaking we don't _need_ the alternate text, so simply removing that should prevent this from happening. Alternatively we could have a default figure title. 

[...]

PR #718, opened 2022-04-20
WIP, ENH: summary with data access cats

### LINKED ISSUE #466: job summary subsection: data access by category
We already have a good start on this figure in #397, so the task here would be to just incorporate that figure into the job summary template.

I think we will need some logic to determine how many file systems to summarize (everyone accessed? top N in terms of bytes moved?), and also need to make sure we have some summary information on standard streams (STDIN/STDOUT/STDERR) captured by STDIO module.

One technical question (not strictly related to this subsection, but to tables in general) is whether we need (or want) to implement the table in HTML code or whether we can use whatever it is that @tylerjereddy is already using to generate the tables, for instance.

tylerjereddy:
Fixes #466 (eventually)

* add the "Data Access by Category" figure to the Python
summary report, and adjust tests to reflect the additional
figure, and support for both `POSIX` and `STDIO` with this fig

* some issues to think about
  - ~the new figure is ridicuously large compared to the previous
    figures in the report and the option to specify the width
    of the figure in the `ReportFigure` has no effect--see example in gh-717~
  - if we have a sample log that lacks both `POSIX` and `STDIO`,
    then `test_posix_absent` might be suitably replaced instead
    of deleted (may help if test coverage isn't 100 % anymore)
  - the code changes here contain comments about an eventual API redesign,
    the issue there should hopefully be clear from the diff
  - test suite ran in 3:34 on 6 cores locally--didn't quantify the
    slowdown relative to `pydarshan-devel` with new analysis/figs
  - there's some discussion from Shane in the matching issue about:
    - how to pick the top `N` categories vs. the magic number I've used here
    - something about ditching the plot entirely and using html table instead?

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -22,6 +22,7 @@ from darshan.experimental.plots import (
     plot_common_access_table,
     plot_access_histogram,
     plot_opcounts,
+    data_access_by_filesystem,
 )
 
 darshan.enable_experimental()
@@ -459,6 +460,20 @@ class ReportData:
                     fig_width=350,
                 )
                 self.figures.append(opcount_fig)
+
+        #########################
+        # Data Access by Category
+        if not {"POSIX", "STDIO"}.isdisjoint(set(self.report.modules)):
+            data_access_by_cat_fig = ReportFigure(
+                section_title="Data Access by Category",
+                fig_title="",
+                fig_func=data_access_by_filesystem.plot_with_report,
+                fig_args=dict(report=self.report, num_cats=8),
+                fig_description="",
+                fig_width=1,
+            )
+            self.figures.append(data_access_by_cat_fig)
+
 
 
     def build_sections(self):
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -469,8 +469,11 @@ class ReportData:
                 fig_title="",
                 fig_func=data_access_by_filesystem.plot_with_report,
                 fig_args=dict(report=self.report, num_cats=8),
-                fig_description="",
-                fig_width=1,
+                fig_description="Summary of data access volume "
+                                "categorized by storage "
+                                "target (e.g., file system "
+                                "mount point) and sorted by volume.",
+                fig_width=500,
             )
             self.figures.append(data_access_by_cat_fig)
 
--- a/darshan-util/pydarshan/darshan/experimental/plots/data_access_by_filesystem.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/data_access_by_filesystem.py
@@ -646,17 +646,25 @@ def plot_with_report(report: darshan.DarshanReport,
     file_id_dict = report.data["name_records"]
     filesystem_roots = identify_filesystems(file_id_dict=file_id_dict,
                                             verbose=verbose)
+    # NOTE: this is a bit ugly, STDIO and POSIX are both combined
+    # automatically later in the control flow, so an API redesign
+    # may be in order eventually
+    default_mod = "POSIX"
+    if "POSIX" not in report.modules:
+        # STDIO can also be used for this figure/analysis
+        default_mod = "STDIO"
+
     file_rd_series, file_wr_series = unique_fs_rw_counter(report=report,
                                                           filesystem_roots=filesystem_roots,
                                                           file_id_dict=file_id_dict,
                                                           processing_func=process_unique_files,
-                                                          mod='POSIX',
+                                                          mod=default_mod,
                                                           verbose=verbose)
     bytes_rd_series, bytes_wr_series = unique_fs_rw_counter(report=report,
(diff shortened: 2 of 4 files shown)


darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -458,6 +459,20 @@ def register_figures(self):
...
 
+        #########################
+        # Data Access by Category
+        if not {"POSIX", "STDIO"}.isdisjoint(set(self.report.modules)):
+            data_access_by_cat_fig = ReportFigure(
+                section_title="Data Access by Category",
+                fig_title="",
+                fig_func=data_access_by_filesystem.plot_with_report,
+                fig_args=dict(report=self.report, num_cats=8),
+                fig_description="",
```
tylerjereddy, line 470: Feel free to suggest content for these fields, and a placeholder when the figure cannot be produced if you want that.

More pressing at the moment is the inability to resize the figure through `ReportFigure` `fig_width` though I think.

shanedsnyder: > there's some discussion from Shane in the matching issue about:
> 
>     * how to pick the top `N` categories vs. the magic number I've used here

I think as long as the current code chooses the top N categories (in terms of total read/write volume) for displaying, we're good. 8 is a reasonable enough value for me -- if more than 8 mount points are found in the log, though, we just want to make sure the 8 most important are displayed. From a quick scan of the code I think this is the case, so good to go.

> 
>     * something about ditching the plot entirely and using html table instead?

Can probably just ignore this. My initial concern was how exactly we could embed tables into the HTML report (mostly just wanting to avoid generating tables with our own HTML code if possible) -- looks like all the tables are generated as mpl figures, which is fine by me.

shanedsnyder: For title: "Data access by category"

For description: "Provides access details for the top data access categories, in terms of cumulative read/write volume. Categories are shown in descending order of total I/O volumes. Possible data access categories currently include file system mount points and standard streams accessed by the application."

@carns might have some better ideas on how to tweak the language. "categories" is kind of a vague term that's hard to work with. But, we need a term vague enough to encapsulate mount points, standard streams, and even things like DAOS object storage pools -- anything a user may read/write from/to basically.

nawtrey:
As I mentioned in gh-717, adding a figure title and updating the figure size corrects the figure width issue for me (diff below):

Screen Shot 2022-04-20 at 2 31 36 PM


<details><summary>The diff</summary>

```diff
diff --git a/darshan-util/pydarshan/darshan/cli/summary.py b/darshan-util/pydarshan/darshan/cli/summary.py
index 6b64069d..ea901469 100644
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -464,11 +464,11 @@ class ReportData:
         if not {"POSIX", "STDIO"}.isdisjoint(set(self.report.modules)):
             data_access_by_cat_fig = ReportFigure(
                 section_title="Data Access by Category",
-                fig_title="",
+                fig_title="Data Access by Filesystem",
                 fig_func=data_access_by_filesystem.plot_with_report,
                 fig_args=dict(report=self.report, num_cats=8),
                 fig_description="",
-                fig_width=1,
+                fig_width=400,
             )
             self.figures.append(data_access_by_cat_fig)
```

</details>

tylerjereddy:
I think the section and figure titles are a bit redundant here, so it may actually be a case where the alt text/title should be excluded for now, so I don't think I want to go that route just yet.

darshan-util/pydarshan/darshan/cli/summary.py
carns: I'm good with that title.  Maybe simplify the description/caption, though?

"Summary of data access volume categorized by storage target (e.g., file system mount point) and sorted by volume."

We can expand on that later if we add other category types, but it's probably best to keep it simple for now.  The "Standard streams" category arguably isn't really a storage target, but I think people will get it.

tylerjereddy:
I made the following adjustments:

- rebased the PR to pull in Nik's bug fix for width adjustment
- adjusted the width up from 1 px to 500 px
- added the caption suggested by Phil

I'll leave it up to reviewers what still needs to be addressed here vs. what is already in other issues vs. what needs a new issue.

Note that whitespace/cropping/legibility may be another thing to consider since the figure is large and then being pixel-resized, e.g. with `e3sm_io_heatmap_only.darshan`:

image

tylerjereddy:
The extra analysis/plot adds about 30 seconds to the full testsuite for me locally, FWIW.

shanedsnyder:
Maybe we can punt the issues related to resizing the figure to another issue and move forward with this for the time being. I think things look mostly good, but would like to clear those things up at some point if we can.

Not sure what leads to the whitespace issue, but maybe there's a way to right-justify all the labels and somehow wrap or truncate them if they get too long (and start to creep into the plotting area) -- it's possible those strings could be pretty long, generally speaking.

shanedsnyder (approved):

shanedsnyder:
Outstanding issue deferred to #728, merging this in. Thanks!

PR #719, opened 2022-04-21
WIP, ENH: Update summary report styling

nawtrey:
Description
-------------
* Use stylesheet from the following location:
https://unpkg.com/style.css@1.0.0/style.css

* Misc adjustments to stylesheet:
  - Add margins to body so titles and figures
  do not touch window edge
  - Remove `figure` border
  - Add `td:empty` entry to prevent empty columns in
  "Darshan Log Information" table from being displayed
  - Change horizontal rule height to 1px, make color darker
  - Remove table tr `border-top`
  - Change table pattern to `2n+1` so tables
  start with darker color

* Correct expected string in `test_stylesheet`

* Add CSS class `section_header`, add to section
title tags to more clearly delineate sections

* Remove "Gill Sans" font family from body

Comments
------------
There has been a fair amount of discussion about adding a proper CSS style sheet to the report. This was the one that got the most positive feedback so I've basically just copy/pasted it into the `style.css` file. The details on the style sheet can be found on their [GitHub page](https://github.com/ungoldman/style.css). It as an ISC license (similar to MIT), so it should be okay to use this modified version of it. The updated style sheet is about 7 KB now. 

I'm opening this as a draft; if we want to keep the basic CSS we have now I can simply close this PR, but if we decide to keep the changes feedback on the current styling is also welcome. 

Examples
------------
Here is a set of example logs that I think is representative of the logs from the darshan-logs repo: [example_reports.tar.gz](https://github.com/darshan-hpc/darshan/files/8533221/example_reports.tar.gz)

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/base.html
+++ b/darshan-util/pydarshan/darshan/cli/base.html
@@ -8,7 +8,7 @@
         ${report_data.stylesheet}
       </style>
     </head>
-    <body style="font-family: Gill Sans MT">
+    <body>
 
       <!-- Add the header  -->
       <div class="fixed-header">
@@ -31,7 +31,7 @@
       <!-- Add sections and their figures  -->
       % for sect_title in report_data.sections:
           <section>
-          <h2>${sect_title}</h2>
+          <h2 class="section_header">${sect_title}</h2>
 
           % if len(report_data.sections[sect_title]) > 1:
             <!-- if more than 1 figure in section, use grid wrapper class -->
--- a/darshan-util/pydarshan/darshan/cli/style.css
+++ b/darshan-util/pydarshan/darshan/cli/style.css
@@ -24,6 +24,11 @@ th, td {
 table {
   border-collapse: collapse;
 }
+.section_header {
+  display: block;
+  background-color: #ddd;
+  padding: 5px;
+}
 .fixed-header, .fixed-footer{
     width: 100%;
     display: block;
--- a/darshan-util/pydarshan/darshan/cli/style.css
+++ b/darshan-util/pydarshan/darshan/cli/style.css
@@ -16,7 +16,9 @@ html {
 }
 
 body {
-  margin: 0;
+  margin-top: 0px;
+  margin-left: 30px;
+  margin-right: 30px;
 }
 
 article,
@@ -270,6 +272,10 @@ th {
   padding: 0;
 }
 
+td:empty {
+  display: none;
+}
+
 table {
   border-collapse: collapse;
   border-spacing: 0;
@@ -292,10 +298,9 @@ table th {
 
 table tr {
   background-color: #fff;
-  border-top: 1px solid #ccc;
 }
 
-table tr:nth-child(2n) {
+table tr:nth-child(2n+1) {
   background-color: #f8f8f8;
 }
 
@@ -303,10 +308,10 @@ hr {
   box-sizing: content-box;
   overflow: visible;
   background: transparent;
-  height: 4px;
+  height: 1px;
   padding: 0;
   margin: 1em 0;
-  background-color: #e7e7e7;
+  background-color: #777;
   border: 0 none;
 }
 
@@ -337,7 +342,6 @@ figure {
 
 figure img {
   background: white;
-  border: 1px solid #c7c7c7;
   padding: 0.25em;
 }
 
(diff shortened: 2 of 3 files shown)


nawtrey:
CI failed while building `darshan-util`:
```
Run mkdir darshan_install
Submodule 'modules/autoperf' (https://github.com/argonne-lcf/autoperf.git) registered for path 'modules/autoperf'
Cloning into '/Users/runner/work/darshan/darshan/modules/autoperf'...
fatal: unable to access 'https://github.com/argonne-lcf/autoperf.git/': The requested URL returned error: 429
fatal: clone of 'https://github.com/argonne-lcf/autoperf.git' into submodule path '/Users/runner/work/darshan/darshan/modules/autoperf' failed
Failed to clone 'modules/autoperf'. Retry scheduled
Cloning into '/Users/runner/work/darshan/darshan/modules/autoperf'...
fatal: unable to access 'https://github.com/argonne-lcf/autoperf.git/': The requested URL returned error: 429
fatal: clone of 'https://github.com/argonne-lcf/autoperf.git' into submodule path '/Users/runner/work/darshan/darshan/modules/autoperf' failed
Failed to clone 'modules/autoperf' a second time, aborting
Error: Process completed with exit code 1.
```

I think I'll wait for feedback before re-triggering the CI.

shanedsnyder:
I really like this a lot, thanks for digging around for it!

shanedsnyder:
I added this to the pydarshan-v1-milestone now so we can prioritize merging ahead of our pre-release, unless others have objections.

tylerjereddy (commented):
I have some comments here:

* ultimately, if Shane likes it I think we should aim to put it in--from a few sample logs locally it is a bit more readable in my opinion as well
* I still think there may be some UX/design improvement on the whitespace/alignment handling on large monitors (for example, below the fold)

<details>

image

</details>

* I think my main concern is how this has been "engineered" though--I'd strongly encourage the team to follow a philosophy more like this for decent-sized volumes of vendored code:
  - instead of squishing the vendored CSS sheet AND our custom shims into a single commit, separate this into two commits: 1) the exact vendored upstream style sheet; 2) a separate well-written commit that contains the patch for our custom shims
  - why? because if the base sheet is ever updated, it becomes much easier to re-apply our customization patch, because we can isolate the diff to a single commit without doing a bunch of manual archeology
  - is this enough to make it easy? Probably still not enough--even better would be to follow what a lot of projects do with vendored "docs styles" and so on, and treat this as a submodule that points to a separate repo in `darshan-hpc/stylesheet` or something like that:
    - why? well, even if Nik uses two separate commits here, if this gets merged to `pydarshan-devel`, then squash merged to `main` in the "big merge," the advantage of my initial suggestion is lost in the history again
    - the other point of the submodule is to have a very small and clean commit history that cycles through:
      - vendor upstream sheet
      - apply our patch
      - vendor new upstream sheet
      - apply our patch
    - also, if there is a "UX team" at some point--they can kind of iterate with their own clean history/repo that won't get interwoven with the main project

shanedsnyder:
> * I think my main concern is how this has been "engineered" though--I'd strongly encourage the team to follow a philosophy more like this for decent-sized volumes of vendored code:

This makes a lot of sense to me. I think at the very least we should adopt the first approach (using separate commits to delineate what our changes are). Up to you guys if you want to go forward with the second approach -- for my 2c, we could probably just adopt the first one and revisit whether we want a separate repo if we find ourselves having to modify things again in the future?

nawtrey:
Sorry for the late response -- Yeah those are all good points. I didn't think about updating the sheet later but I can split this into 2 commits to make the changes more clear. If we wait and merge it into `main` instead, that should keep the commit history clear

nawtrey:
I've split the changes into 3 commits, rebased on `main`, and changed this PR to target `main`.

Summary of the 3 commits: 
1. Misc changes (adding the section header CSS class, removing "Gill Sans" from body)
2. Applying the new style sheet directly, fully replacing the style sheet
3. Applying our "patch" which includes our custom-defined CSS classes for handling the header, footer, grid formatting, and the new section header, as well as the small changes to the stylesheet which fix some issues with the tables.

I also looked into the figure gaps (noted in Tyler's [comment](https://github.com/darshan-hpc/darshan/pull/719#pullrequestreview-951259259)). The large gap between figures is a result of the `figure` and `img` tag scaling (or lack thereof). I tried adjusting these to 100% (for all figures) and that fixed the large gap, at least when I zoom out in my browser window (I don't have an ultra wide monitor on hand). Unfortunately this results in figures that are much larger than others when comparing sections with a single figure vs. multiple figures. We may be able to set something like the `max-width` parameter in the CSS as well, which might fix that issue.. I'm sure there is a simple solution that allows for similarly sized figures without an abundance of whitespace, but I think that can be handled separately from this PR? I guess it depends on how quickly the team wants this merged. 

As a side note, handling the figure scaling in the CSS would probably remove the need for the `ReportFigure(fig_width)` parameter entirely, which I know we've discussed already.

Lastly, I'm a little concerned with our ability to scale figures _and_ scale the figure text size since the font sizes are determined in the plotting code, independently from each other and the report text size. For the time being, as long as appropriate figure + text size pairs are chosen, the figures should be okay I think. But the "ideal" report would have common font sizing throughout, at least at a glance.

tylerjereddy (approved):
LGTM now, I checked a bunch of things locally include a few sample summary reports and that the commits are sensibly organized.

Once CI passes I plan to merge this in, with commits preserved for obvious reasons.

Remaining design/layout concerns can be deferred I think yeah.

tylerjereddy:
thanks!

PR #722, opened 2022-04-26
BUG: cats only allow POSIX/STDIO

### LINKED ISSUE #678: BUG: autoperf handling by "data access by category" infrastructure
Splitting off from gh-397. With code similar to this some of the autoperf records/data structure details produce "categories" that we probably don't want (see image below). I think @shanedsnyder has some idea for handling this.

```python
import darshan
from darshan.experimental.plots import data_access_by_filesystem

log_path = get_log_path("mpi-io-test.darshan")
fig = data_access_by_filesystem.plot_with_report(report=report,
                                                     num_cats=None)
fig.savefig("mpi-io-test.png", dpi=300)
```

mpi-io-test darshan

tylerjereddy:
Fixes #678

* only allow `POSIX` and `STDIO` records
to get passed through the control flow
that produces the "Data Access by Category"
plots

* add one new test case and adjust old cases
to reflect the more selective plotting of
categories

Below the fold I'll place the updated examples for the problematic cases
cited in the matching issue. It is trivial to add new test cases so if reviewers
have others they know pretty well and want to see tested that should be easy
enough to do.

<details>

<summary>
For mpi-io-test that had spurious autoperf entries and inactive std streams
</summary>

mpi-io-test

</details>

<details>
<summary>
For the Tensorflow monitoring log that had spurious heatmap entries and inactive std streams
</summary>

tf_fail

</details>

I have to admit that because this makes the number of categories much smaller in general, the "width" of the bars is a bit excessive sometimes, but that is perhaps to be dealt with separately.

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -25,8 +25,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade mypy pyflakes asv codecov lxml
-          python -m pip install pytest-cov "pytest<7.1.2"
+          python -m pip install --upgrade pytest mypy pyflakes asv pytest-cov codecov lxml
           # matplotlib is pinned because of
           # gh-479
           python -m pip install 'matplotlib<3.5'
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -25,8 +25,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install --upgrade mypy pyflakes asv codecov lxml
-          python -m pip install pytest-cov "pytest==7.0.1"
+          python -m pip install --upgrade pytest mypy pyflakes asv pytest-cov codecov lxml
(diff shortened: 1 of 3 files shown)


tylerjereddy:
Pinning `pytest` to `7.1.1` had no effect, I'll try rolling it back a bit farther but since I can't reproduce locally anyway, not sure. I understand the complaint of the error message, but we violate that suggestion with good reason and it has mostly worked so far.

tylerjereddy:
Ok, I've tried pinning to `pytest` `7.0.1` as well. That didn't work, so I'll try to circle back tonight--the useful `act` tool for iterating on GitHub actions locally is currently blocked at LANL because of our SSL interception protocols :/ . If someone else wants to see if they can reproduce with `act` on this branch feel free to check.

I also tried `git diff --no-index --word-diff` on this failure log vs. a recent success log and didn't see anything super-obvious, though it can be easy to miss something subtle.

tylerjereddy:
ok, CI issues are resolved now

shanedsnyder (approved):

shanedsnyder:
Outstanding issue deferred to #728, merging this in. Thanks!

PR #725, opened 2022-04-27
MAINT: pyflakes for act

### LINKED ISSUE #721: MAINT: act tool currently fails
Use of the [act tool](https://github.com/nektos/act) has been growing in the community to facilitate local iteration on CI testing for GitHub actions.

It almost succeeds with our current `pydarshan-devel` branch--the main test suite does pass for all Python versions it seems, but there is a failure later on: 

```
| =========== 412 passed, 2 xfailed, 346 warnings in 403.52s (0:06:43) ===========
[Python Testing/test_pydarshan-9 ]   ✅  Success - Test with pytest
[Python Testing/test_pydarshan-9 ] ⭐  Run pyflakes check
[Python Testing/test_pydarshan-9 ]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /var/run/act/workflow/10] user= workdir=
[Python Testing/test_pydarshan-9 ]   ❌  Failure - pyflakes check
[Python Testing/test_pydarshan-9 ] exit with `FAILURE`: 1
Error: Job 'test_pydarshan' failed
```

May need to double check the robustness of the shell-style exit code handling in the `pyflakes check` section of CI control flow:

`pyflakes darshan/backend | (grep -E -i 'assigned to but never used' || exit 0 && exit 123)`

tylerjereddy:
Fixes #721

* this allows the `pyflakes` check to pass when using
the [`act`](https://github.com/nektos/act) tool

* not quite sure why the tool deviates slightly from
GitHub actions proper, but I did check that if I add
a string that is present in the linter log it does
fail with `act`

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -81,7 +81,11 @@ jobs:
       - name: pyflakes check
         run: |
           cd darshan-util/pydarshan
-          pyflakes darshan/backend | (grep -E -i 'assigned to but never used' || exit 0 && exit 123)
+          set +e
+          pyflakes darshan/backend 2>&1 | tee lint_pyflakes_ci.txt
+          set -e
+          function check_pyflakes() { if grep -c -E -i "assigned to but never used" lint_pyflakes_ci.txt; then return -1; fi }
+          check_pyflakes
       - name: asv check
         run: |
           export LD_LIBRARY_PATH=$PWD/darshan_install/lib


tylerjereddy:
probably, though not super high-prio; should confirm we still need it for merging; could also consider using a conditional and keeping the CI the way it is and only using a different path if running locally

tylerjereddy:
Ideally we'd just find a symmetric solution that works in both cases

tylerjereddy:
this is targeted at the wrong branch, but I also don't like my own solution here, so closing

PR #726, opened 2022-04-28
increase MAX_MODS to 64

### LINKED ISSUE #667: boost the maximum number of modules (DARSHAN_MAX_MODS) beyond 16
There is a hard coded limit on the number of modules iin darshan-log-format.h:

```
#define DARSHAN_MAX_MODS 16
```

This is not a limit on the number of modules that can be _activated_ at run time; it is a limit on how many can be defined in the code base.  Each module in Darshan gets a distinct integer identifier that does not change over time.

origin/main already has 15 modules, and we anticipate multiple new ones shortly that will push us over the limit.

Bumping this value will probably require incrementing the overall Darshan log format because it alters the length of static arrays in the header.

### LINKED ISSUE #712: audit code to ensure safe usage of DARSHAN_MAX_MODS
We should inspect darshan-runtime and darshan-util usage of `DARSHAN_MAX_MODS` to ensure safety. Specifically, arrays like `darshan_module_names` only have `DARSHAN_KNOWN_MODULE_COUNT` entries, but some code sets up an iterator that goes up to `DARSHAN_MAX_MODS`, which could lead to accessing invalid pointers. One example would be the logic for printing info when users set `DARSHAN_INTERNAL_TIMING` env var: https://github.com/darshan-hpc/darshan/blob/41467a22a86dc95ac389cc6cfa21ad6908cb6137/darshan-runtime/lib/darshan-core.c#L767

shanedsnyder:
This PR bumps Darshan's MAX_MODS value from 16 to 64. This is a small change to the header for the most part.

There are additional commits to address a couple of other issues:

- switched almost all references to `DARSHAN_MAX_MODS` to `DARSHAN_KNOWN_MODULE_COUNT` for safety purposes (fixes #712)
- small `darshan-dxt-parser` fix to align behavior with `darshan-parser` -- make sure to include info in the module summary section of parser output for modules that report partial data

This PR is marked WIP for a few reasons:

1. There is no backwards-compatibility support currently for previous log versions. I am still working on that.
2. No Python bindings support for this change yet.
3. I don't want to rush this in ahead of the upcoming pre-release, we can make sure we get this right and get it merged afterwards so it's part of the following release (pre-release or 3.4.0)

Fixes #667

### DIFF
--- a/darshan-util/darshan-logutils.c
+++ b/darshan-util/darshan-logutils.c
@@ -1145,7 +1145,7 @@ static int darshan_log_get_header(darshan_fd fd)
         header.partial_flag = (uint64_t)header_3_00.partial_flag;
         memcpy(&header.name_map, &header_3_00.name_map,
             (1 + DARSHAN_MAX_MODS_3_00) * sizeof(header_3_00.name_map));
-        mempcpy(&header.mod_ver, &header_3_00.mod_ver,
+        memcpy(&header.mod_ver, &header_3_00.mod_ver,
             (DARSHAN_MAX_MODS_3_00) * sizeof(header_3_00.mod_ver[0]));
 
         fd->job_map.off = sizeof(header_3_00);
--- a/darshan-util/darshan-logutils.c
+++ b/darshan-util/darshan-logutils.c
@@ -1025,7 +1025,8 @@ static int darshan_log_get_header(darshan_fd fd)
     }
     else if((strcmp(fd->version, "3.10") == 0) ||
             (strcmp(fd->version, "3.20") == 0) ||
-            (strcmp(fd->version, "3.21") == 0))
+            (strcmp(fd->version, "3.21") == 0) ||
+            (strcmp(fd->version, "3.40") == 0))
     {
         fd->state->get_namerecs = darshan_log_get_namerecs;
     }
--- a/darshan-util/darshan-logutils.c
+++ b/darshan-util/darshan-logutils.c
@@ -80,7 +80,7 @@ struct darshan_fd_int_state
 
 /* each module's implementation of the darshan logutil functions */
 #define X(a, b, c, d) d,
-struct darshan_mod_logutil_funcs *mod_logutils[DARSHAN_MAX_MODS] =
+struct darshan_mod_logutil_funcs *mod_logutils[DARSHAN_KNOWN_MODULE_COUNT] =
 {
     DARSHAN_MODULE_IDS
 };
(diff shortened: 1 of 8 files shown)


shanedsnyder:
I resolved the WIP status of this PR and think it's ready for review/merge.

Backwards compatibility code has been added to handle the up-conversion of logs with old header versions.

AFAICT, no Python bindings changes are required -- this change only touches the Darshan log header, which is not exposed to PyDarshan. Limited testing on my end didn't trip up anything and CI is passing, so looks good.

To test these changes out, I wrote a script that tries to parse a bunch of Darshan log files using 2 different parser versions: 3.4.0 version (which uses our old header definition that is present in all existing 3.x logs) and a version based on this branch with a header that supports 64 modules. I then use diff to compare the output from darshan-parser in each case. In my testing using the darshan-logs repo and also our suite of Darshan release logs, the only reported difference in any case was the increased header size -- all counters and log data are identical. So, on the testing front things look good. Script is copied below if useful:

<details>
<summary>log script</summary>
<pre>
#!/bin/bash
if [ "$#" -ne 3 ] ; then
    echo "Usage: log-test <log_dir> <old_darshan_prefix> <3_41_darshan_prefix>"
    exit -1
fi <br>
log_dir=$1
prefix_old=$2
prefix_3_41=$3 <br>
for log in `find $log_dir -name "*.darshan"`; do
    log_base=`basename $log`
    $prefix_old/bin/darshan-parser --show-incomplete $log > $log_base.old.out
    $prefix_3_41/bin/darshan-parser --show-incomplete $log > $log_base.3_41.out
    diff $log_base.old.out $log_base.3_41.out
done
</pre>
</details>

darshan-util/darshan-logutils.c
```diff
@@ -991,6 +992,51 @@ static int darshan_log_get_namerecs(void *name_rec_buf, int buf_len,
     return(buf_processed);
 }
 
+
```
carns, line 995: What is the expected format of the log version string?

If it's `<decimal>.<decimal>` we could probably simplify this a lot with `sscanf`?

Something like:

```
ret = sscanf(ver_str, "%d.%d", maj_num, min_num);
if(ret < 2) {
   /* error, it's not formatted right */
}
/* otherwise we are good */
```

carns (changes requested):
Looks good, other than tidying up the string parsing.

Thanks for fixing all of those mod count paths!  I didn't realize we had so many spots like that left, but the changes all make sense.

darshan-util/darshan-logutils.c
shanedsnyder: Good call, I don't know how I ended up with anything other than this implementation :)

See [9ae46b2](https://github.com/darshan-hpc/darshan/pull/726/commits/9ae46b2fbf89b3f1aa7b13cb3f2d3e91a285c229) for satisfying diff of changes.

carns (approved):
great, thanks for the parsing cleanup

PR #727, opened 2022-05-02
run pytest on each built wheel

shanedsnyder:
Originally discussed in #705, but these changes add some `pytest` testing that is exercised for each wheel/python version that we currently support. This slows down wheel building considerably, but that's probably fine as we only do this as part of a release process where we want to double check we aren't hitting weird errors for specific versions.

This is far from sophisticated, but is useful to sanity check there are no errors until we get a more comprehensive wheel building and CI setup using `cibuildwheels`.


shanedsnyder:
@jakobluettgau -- I corrected an error in the patch I originally mentioned in your previous PR, and it is accounted for here. That probably caused the errors you mentioned to me last week, if you wouldn't mind trying again when you have a chance.

PR #731, opened 2022-05-04
massive merge of ongoing PyDarshan development in pydarshan-devel

shanedsnyder:
This branch encapsulates the following:

- merge of current `main` into current `pydarshan-devel`
- removed a couple of stale files missed in a previous merge
- really small whitespace fixes I noticed during merge

It's not really practical to scan this entire diff, but I did ensure that things look sensible comparing this branch to both current `main` (changes exclusively correspond to work in `pydarshan` directory as well as `.git`) and to current `pydarshan-devel` (changes mostly correspond to recent dev activity on our `main` branch).

I'm going to hold off on merging this until I've done more testing but will put this here for others to sanity check in the mean time.


.github/workflows/main_ci.yml
```diff
@@ -0,0 +1,94 @@
+name: Python Testing
+
+on:
+  push:
+    branches:
+      - pydarshan-devel
+  pull_request:
+    branches:
+      - pydarshan-devel
```
tylerjereddy, line 9: I can do it if you want, but do you mind if we change this branch name to `main` here and above.

If I see the CI run for this PR that would be useful in terms of the number of tests that run vs. current `pydarshan-devel` locally, etc.

darshan-util/darshan-logutils.h
```diff
@@ -71,6 +71,7 @@ struct darshan_mod_info
     int  len;
     int  ver;
     int  idx;
+    int partial_flag;
```
tylerjereddy, line 74: double checking: is this expected for merging back into `main`? (along with other changes in `darshan-util`)

tylerjereddy (commented):
I made two small comments to start, overall the Python changes I skimmed over seemed reasonable.

darshan-util/darshan-logutils.h
carns: looks ok to me, assuming the python bindings match this change too (I assume they do).

carns (approved):
It's not tractable to review all of this at this stage, but I think we've done a good job in the incremental merges to the devel branch.  Good to go from my perspective if it is passing CI.

darshan-util/darshan-logutils.h
shanedsnyder: Yeah, this was the infrastructure needed to pass the module "partial_flag" up to PyDarshan so that we could notify about modules that were missing data. It was only implemented in pydarshan-devel, so this is expected in the merge.

.github/workflows/main_ci.yml
shanedsnyder: I just pushed a commit for this.

It may already be the case, but I'm assuming there's a way to restrict these tests so they only run on changes to darshan-util/pydarshan? That should help restrict running tests when nothing really has changed that would impact pydarshan.

shanedsnyder: Yeah, probably any changes to `darshan-util`, as well as `include` which has the log format headers.

tylerjereddy:
Comparing CI here with latest `pydarshan-devel` on one of our local nodes shows identical test suite statistics, which would be almost impossible to achieve without an extremely close match to current `pydarshan-devel`, so +1 on merge.

| Condition | With Logs Repo | No Logs Repo |
| ------------- | ------------- | ------------ |
| `pydarshan-devel` skylake-gold LANL node Python 3.9 | `412 passed, 2 xfailed` | `318 passed, 60 skipped`
| Linux CI here Python 3.9 (with)/3.8 (without) | `412 passed, 2 xfailed` | `318 passed, 60 skipped`
| MacOS CI here Python 3.9 (with)/3.8 (without) | `412 passed, 2 xfailed`  | `318 passed, 60 skipped`

Issue #730, opened 2022-05-04
BUG: potential heatmap errors for test case with buggy log data


shanedsnyder:
See details of Python errors in https://github.com/darshan-hpc/darshan-logs/pull/35.

They seem to imply a shape mismatch in heatmap module data, though I did not see anything obviously wrong in terms of inconsistent heatmap shapes across processes when scanning the log records (should check programmatically to confirm). It's also worth noting that this example log shows general corruption or overflow of heatmap counters, so it does seem reasonable that heatmap parameters could be inconsistent somewhere, leading to the error. We should figure out whether the issue is related to corruption of module data at runtime, or if there is a subtle bug in PyDarshan code.

tylerjereddy:
@shanedsnyder 

On the `pydarshan` side, this is what I think is happening with that tensorflow log:

There are 18 MPI ranks, but 6 of them are completely inactive, so we get a shape mismatch between `nprocs` (*all* 18 ranks according to the darshan bindings) used for y-axis, and the data present in the `STDIO` heatmap dataframe (*only* 12 ranks).

This doesn't cause a crash for the original DXT control flow, because that goes through `get_heatmap_df()`, which has a pandas `reindex` operation that adds inactive ranks back in as all-zero rows.

So, suggested path forward:

1. I should make a much smaller/simpler reproducing log file with a i.e., a few active ranks and one or two inactive ranks, with `HEATMAP` monitoring turned on.
2. Write a reproducing test with that smaller log file.
3. Add a reindex operation to the `HEATMAP` control flow somewhere to fix the bug (basically, include inactive ranks as all-zero rows in the data structure).

tylerjereddy:
Looks like the third rank (rank 2) writes about 42 GB of data before the DXT POSIX stops recording it, while HEATMAP allows it to go right to the end, writing about 200 GB of data. Curiously though, the overflow in HEATMAP is on rank 1, not the rank writing tons of data, which we previously observed. I do see a divide by zero happening in the float processing for DXT POSIX somewhere, and there is an integer vs. float thing going on here too. I'll probably just focus on the matter of the overflow though, I'll see if I can get some `printf` traction there or something.


```
[...]

PR #735, opened 2022-05-07
BUG: pydarshan HEATMAP inactive handle

tylerjereddy:
* deal with 1/2 of the issue described in gh-730

* allow `plot_heatmap()` to gracefully handle
the scenario where a log file has some inactive
ranks and the new runtime `HEATMAP` tracking

* this branch is meant to accompany the new logs
repo PR: https://github.com/darshan-hpc/darshan-logs/pull/37

* minor testing changes to accomodate that new
log file; the test suite will fail with that new
log file without the patch provided here

If you're curious, this is what the new HEATMAP looks like for that log file we couldn't previously process, with the "darker" ranks printing one extra character because the `printf` includes i.e., `rank 10` instead of `rank 9` (so the heatmap scale is super sensitive here).

image


shanedsnyder:
Looks good, thanks!

PR #737, opened 2022-05-11
BUG: fix runtime HEATMAP overflow

tylerjereddy:
This deals with the C/runtime portion of gh-730 and https://github.com/darshan-hpc/darshan-logs/pull/35.

* `heatmap_update()` could sometimes overflow in the early
stages of monitoring, so fix that issue; I see no (unit) tests for this function
so I guess we're depending on code review/static analysis/my own
iterations

* if you are curious what the debug prints showed at the start of the horovod/tensorflow monitoring before the fix, see below the fold

<details>

```
2022-05-10 16:46:37.801963: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1525] Created device /job:localhost/replica:0/task:0/device:GPU:0 with 30943 MB memory:  -> device: 1, name: Tesla V100-PCIE-32GB, pci bus id: 0000:d8:00.0, compute capability: 7.0
2022-05-10 16:46:37.805229: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1525] Created device /job:localhost/replica:0/task:0/device:GPU:0 with 30943 MB memory:  -> device: 0, name: Tesla V100-PCIE-32GB, pci bus id: 0000:86:00.0, compute capability: 7.0
size: 0
seconds_in_bin: 0.000000
start_time: 4.508365
end_time: 4.508365
second term: inf
intermediate_bytes: -9223372036854775808
negative intermediate_bytes ------------------------------
size: 0
seconds_in_bin: 0.000000
start_time: 4.530574
end_time: 4.530574
second term: inf
intermediate_bytes: -9223372036854775808
negative intermediate_bytes ------------------------------
size: 0
seconds_in_bin: 0.000000
start_time: 4.588113
end_time: 4.588113
second term: inf
intermediate_bytes: -9223372036854775808
negative intermediate_bytes ------------------------------
size: 0
seconds_in_bin: 0.000000
start_time: 4.923882
end_time: 4.923882
second term: inf
intermediate_bytes: -9223372036854775808
negative intermediate_bytes ------------------------------
size: 0
seconds_in_bin: 0.000000
start_time: 5.071924
end_time: 5.071924
second term: inf
intermediate_bytes: -9223372036854775808
negative intermediate_bytes ------------------------------
```

</details>

I can easily reproduce the overflow locally as well:

```c
#include <stdio.h>
#include <math.h>

int main() {
    int64_t intermediate_bytes;
    intermediate_bytes = round(0. * (0. / 0.));
    printf("intermediate_bytes: %lld\n", intermediate_bytes);
}
```

`intermediate_bytes: -9223372036854775808`

The exact value that can sneak into the log sometimes.

Sample processing of a logfile that we couldn't do before (similar to the cross-linked logs repo case) looks much more sensible than without the runtime fix at https://github.com/darshan-hpc/darshan/issues/730#issuecomment-1122637152:

image

### DIFF
--- a/darshan-runtime/lib/darshan-heatmap.c
+++ b/darshan-runtime/lib/darshan-heatmap.c
@@ -373,6 +373,7 @@ void heatmap_update(darshan_record_id heatmap_id, int rw_flag,
     struct heatmap_record_ref *rec_ref;
     int bin_index = 0;
     double top_boundary, bottom_boundary, seconds_in_bin;
+    int64_t intermediate_bytes;
 
     HEATMAP_PRE_RECORD_VOID();
 
@@ -419,13 +420,25 @@ void heatmap_update(darshan_record_id heatmap_id, int rw_flag,
             return;
         }
 
+        intermediate_bytes = round(size * (seconds_in_bin/(end_time-start_time)));
+
+        /* NOTE: intermediate_bytes consistently produced the int64_t value
+         * -9223372036854775808 in the early stages of monitoring
+         *  for the case of round(0 * (0 / 0)), so just set to 0 when negative;
+         *  this appears to happen primarily in the very early stages
+         *  of monitoring when seconds_in_bin is 0.0; see gh-730
+         *  and https://github.com/darshan-hpc/darshan-logs/pull/35
+         */
+        if(intermediate_bytes < 0)
+            intermediate_bytes = 0;
+
         /* proportionally assign bytes to this bin */
         if(rw_flag == HEATMAP_WRITE)
             rec_ref->heatmap_rec->write_bins[bin_index] +=
-                round(size * (seconds_in_bin/(end_time-start_time)));
+                intermediate_bytes;
         else
             rec_ref->heatmap_rec->read_bins[bin_index] +=
-                round(size * (seconds_in_bin/(end_time-start_time)));
+                intermediate_bytes;
     }
 
     HEATMAP_POST_RECORD();


darshan-runtime/lib/darshan-heatmap.c
```diff
@@ -419,13 +420,25 @@ void heatmap_update(darshan_record_id heatmap_id, int rw_flag,
...
+
+        /* NOTE: intermediate_bytes consistently produced the int64_t value
+         * -9223372036854775808 in the early stages of monitoring
+         *  for the case of round(0 * (0 / 0)), so just set to 0 when negative;
+         *  this appears to happen primarily in the very early stages
+         *  of monitoring when seconds_in_bin is 0.0; see gh-730
+         *  and https://github.com/darshan-hpc/darshan-logs/pull/35
+         */
+        if(intermediate_bytes < 0)
+            intermediate_bytes = 0;
```
tylerjereddy, line 433: it also occurs to me that short-circuiting to `0` when `size` is `0` might save a bit of electricity

carns:
Thanks for the catch @tylerjereddy !

We probably need a safety check even higher up instead, though, for end_time <= start_time.  Otherwise some other combination of arguments might cause a div-by-zero exception?

shanedsnyder:
> Thanks for the catch @tylerjereddy !
> 
> We probably need a safety check even higher up instead, though, for end_time <= start_time. Otherwise some other combination of arguments might cause a div-by-zero exception?

Maybe 2 changes ultimately, based on comments from you guys:
- bail out of the function completely for `size==0`
- only apply size to the very first bin in cases where `end_time<=start_time` (maybe it's easier to do this ahead of the loop? I didn't look that closely)

tylerjereddy:
> FWIW I'm guessing that what happens to trigger this is probably user code issuing read() or fread() calls until EOF. The last zero byte read might return too quickly (maybe all in userspace in some cases?) for the timer to increment.

@carns Thanks for the suggestion--so far I've not been able to reproduce with a simple C-based regression test that does something like this in gh-756. I'm open to suggestions if that reproducer doesn't quite match what you had in mind.

If nothing else, it should demonstrate one approach I had in mind for providing simple C-based reproducers to the functional test suite for team members that prefer that to Python.

carns:
Thanks @tylerjereddy .  I haven't followed the testing infrastructure discussion, but I'll grab that .c example to play with manually on my laptop to see if I can repro the problem.

shanedsnyder:
@tylerjereddy here's a patch for this I came up with along the lines of my last comment:

```
diff --git a/darshan-runtime/lib/darshan-heatmap.c b/darshan-runtime/lib/darshan-heatmap.c
index 7b64e4f..b36ae75 100644
--- a/darshan-runtime/lib/darshan-heatmap.c
+++ b/darshan-runtime/lib/darshan-heatmap.c
@@ -375,6 +375,9 @@ void heatmap_update(darshan_record_id heatmap_id, int rw_flag,
     double top_boundary, bottom_boundary, seconds_in_bin;
     int64_t intermediate_bytes;

+    /* if size is zero, we have no work to do here */
+    if(size == 0) return;
+
     HEATMAP_PRE_RECORD_VOID();

     rec_ref = darshan_lookup_record_ref(heatmap_runtime->rec_id_hash, &heatmap_id, sizeof(darshan_record_id));
@@ -420,17 +423,10 @@ void heatmap_update(darshan_record_id heatmap_id, int rw_flag,
             return;
         }

-        intermediate_bytes = round(size * (seconds_in_bin/(end_time-start_time)));
-
-        /* NOTE: intermediate_bytes consistently produced the int64_t value
-         * -9223372036854775808 in the early stages of monitoring
-         *  for the case of round(0 * (0 / 0)), so just set to 0 when negative;
-         *  this appears to happen primarily in the very early stages
-         *  of monitoring when seconds_in_bin is 0.0; see gh-730
-         *  and https://github.com/darshan-hpc/darshan-logs/pull/35
-         */
-        if(intermediate_bytes < 0)
-            intermediate_bytes = 0;
+        if(end_time > start_time)
+            intermediate_bytes = round(size * (seconds_in_bin/(end_time-start_time)));
+        else
+            intermediate_bytes = size;

         /* proportionally assign bytes to this bin */
         if(rw_flag == HEATMAP_WRITE)
```

It contains a short-circuit for the case `size == 0`, and gracefully avoids the divide by zero so we don't have to correct it afterwards.

carns:
@shanedsnyder 's fix looks good to me; should cover all the bases.

darshan-runtime/lib/darshan-heatmap.c
```diff
@@ -419,13 +423,18 @@ void heatmap_update(darshan_record_id heatmap_id, int rw_flag,
             return;
         }
 
+        if(end_time > start_time)
+            intermediate_bytes = round(size * (seconds_in_bin/(end_time-start_time)));
+        else
+            intermediate_bytes = size;
+
```
tylerjereddy, line 430: @shanedsnyder @carns I pushed in Shane's patch/adjustments and confirmed they allow the regression test over in gh-756 (which now actually works, albeit in a complex way) to pass.

tylerjereddy: you may want to edit the history to co-author the patch though

shanedsnyder (approved):

shanedsnyder:
Thanks guys, going to merge this.

PR #739, opened 2022-05-11
WIP: DAOS DFS module support

shanedsnyder:
DAOS DFS module support is mostly complete in this PR. It is based on #726 (should be merged first), since this module is the first to push us past our previous max of 16.

Captured counters are similar in nature to our existing POSIX module. Record names are generated using: `pool_uuid:cont_uuid:/file/path`

Known issues:
- `dfs_obj_global2local()` is not supported, currently, as we have no DAOS APIs to convert args to this function to object name to generate a Darshan record. I am investigating whether there's another way for all DAOS processes to agree on record ID in cases where this call is used (it is used in IOR, which is where I've done all my testing thus far). One possibility is to base Darshan record IDs off of underlying DAOS object ID rather than it's name. 
- Non-blocking read/write calls are not properly timed by Darshan currently -- we just time how long it takes for the operation to be submitted, not when it eventually completes. We should use a similar approach to what we do in POSIX AIO routines.
- Updated documentation for building darshan-runtime with DAOS support
- Updated documentation on DAOS record formats in darshan-util docs


tylerjereddy:
If the CI remains green here it is probably ready for some initial review/criticism. Beyond the code itself, you could also i.e., just try generating the summary report with the new DFS-enabled log file and see if there is anything that doesn't make sense.

darshan-util/pydarshan/darshan/tests/test_report.py
```diff
@@ -78,6 +78,84 @@ def test_load_records():
...
+                               "DFS_CHUNK_SIZE", "DFS_USE_DTX", "DFS_FASTEST_RANK",
+                               # performance-related metrics probably can't be
+                               # expected to match exactly, even if the technologies
+                               # are similar in perf
+                               "DFS_F_READ_TIME", "DFS_F_WRITE_TIME",
+                               "DFS_F_META_TIME", "DFS_F_MAX_READ_TIME",
+                               "DFS_F_MAX_READ_TIME", "DFS_F_MAX_WRITE_TIME",
+                               "DFS_F_FASTEST_RANK_TIME", "DFS_F_SLOWEST_RANK_TIME"]:
+                continue
+            elif "timestamp" in column_name.lower():
```
shanedsnyder, line 127: You could probably just filter out any counter/fcounter column name with `TIME` in it to simplify things here -- you're right that we can't expect any of those to ever be exactly equivalent.

darshan-util/pydarshan/darshan/tests/test_report.py
```diff
@@ -78,6 +78,84 @@ def test_load_records():
...
+            # for some columns we can't reasonably expect a match
+            # or we need to handle the data differently between POSIX
+            # and DAOS DFS
+            if column_name in ["id", "DFS_LOOKUPS", "DFS_NB_READS", "DFS_NB_WRITES",
+                               "DFS_GET_SIZES", "DFS_PUNCHES", "DFS_MOVES",
+                               "DFS_EXCHANGES",
+                               # NOTE: POSIX_STATS exists and is different
+                               # from DFS_STATS--is this unusual??
+                               "DFS_STATS",
+                               "DFS_CHUNK_SIZE", "DFS_USE_DTX", "DFS_FASTEST_RANK",
```
shanedsnyder, line 118: May want to go ahead and add `DFS_SLOWEST_RANK` here, too. I'm guessing they are the same for these 2 logs, but aren't guaranteed to be. We'll probably have 1 more iteration of replacing these IOR logs (to get the right 3.4.1 release version in them), so might save us having to fix then.

darshan-util/pydarshan/darshan/tests/test_report.py
```diff
@@ -78,6 +78,84 @@ def test_load_records():
...
+    for counter_type in ["counters", "fcounters"]:
+        for column_name in dfs_data_dict[counter_type].columns:
+            # for some columns we can't reasonably expect a match
+            # or we need to handle the data differently between POSIX
+            # and DAOS DFS
+            if column_name in ["id", "DFS_LOOKUPS", "DFS_NB_READS", "DFS_NB_WRITES",
+                               "DFS_GET_SIZES", "DFS_PUNCHES", "DFS_MOVES",
+                               "DFS_EXCHANGES",
+                               # NOTE: POSIX_STATS exists and is different
+                               # from DFS_STATS--is this unusual??
```
shanedsnyder, line 116: I'd have to have a closer look at the differences between IOR's POSIX and DFS backends, but I wouldn't worry too much about the stat call discrepancy. POSIX `stat()` might be used more liberally in the POSIX backend, and DFS has other methods of getting stat-like info (i.e., `dfs_get_size()`), so hard to draw many conclusions, I think.

shanedsnyder:
This all looks good to me, thanks @tylerjereddy ! Looks like the code to add DFS support for the summary tool was pretty trivial, which is what I was hoping for.

I added a review for some small things, but I don't really have any major suggestions. Beyond those small things, I'm not sure any additional PyDarshan stuff is needed for this PR. I still have some work to do on some final runtime library stuff before we can merge, though.

shanedsnyder:
Oh, and as I alluded to in one of my review comments, we'll probably have to re-generate these logs with the appropriate log format versions and runtime library versions and move them to to the log repo at some point. Sorry, there's probably an easier way that could be handled going forward (maybe bumping our library versions right after releases instead of right before them?)...

tylerjereddy:
I think I addressed your first round of revisions--maybe you can deal with the merge conflicts in the C library build system and ping me when I need to adjust this PR again when you change something/regenerate the log files, etc.

shanedsnyder:
Thanks @tylerjereddy, all looks great to me. I resolved the review comments I had, I think we're good to go on PyDarshan stuff here.

PR #740, opened 2022-05-13
CI: focus pypdarshan ci on include/darshan-util paths

### LINKED ISSUE #732: CI: only run pydarshan tests for subset of changes
Based on the comments here: https://github.com/darshan-hpc/darshan/pull/731#discussion_r865992718

we probably only want to run the `pydarshan` CI/Python test suite for:

- any changes to `darshan-util`
-  changes to `include` which has the log format headers

I suggest we adjust that carefully after the big merge in gh-731. As for how we do it, the internet seems pretty helpful here i.e., https://stackoverflow.com/a/63822945/2942522 has:

```
on:
  push:
    paths:
      - folder1/**
```

shanedsnyder:
PyDarshan CI should only run on pushes and PRs that affect `darshan-util` and/or `include` paths.

Fixes #732

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -4,9 +4,15 @@ on:
   push:
     branches:
       - main
+    paths:
+      - darshan-util/**
+      - include/**
   pull_request:
     branches:
       - main
+    paths:
+      - darshan-util/**
+      - include/**
 
 jobs:
   test_pydarshan:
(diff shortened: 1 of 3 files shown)


shanedsnyder:
Can only test the `pull_request` portion of this obviously, but I verified that pydarshan CI only ran for commits that modified something in `darshan-util` or `include`, so this looks to be working as expected.

tylerjereddy (approved):
LGTM, I don't know if there are plans to use the CI for some of the other C code checking, but makes sense at the moment.

PR #741, opened 2022-05-16
MAINT, CI: cibuildwheel support

shanedsnyder:
This PR includes updates for supporting `cibuildwheel`:

- `cibuildwheel` configuration in `pyproject.toml` which works for my local Linux builds over Docker
- first cut at GitHub workflow for automatically building wheels on updates to code that affects PyDarshan
- change `setup.py` argument `--with-extension` to env variable `PYDARSHAN_BUILD_EXT` 

The last change was needed because I couldn't find any other way to get user arguments to `setup.py` from `cibuildwheel`. There seems to be lack of general support in `cibuildwheel` for user-specified options to `setup.py`, as well as issues with `pip` supporting the `--build-option` flag we had been using -- see discussion at https://github.com/pypa/cibuildwheel/issues/268. The issue with `--build-option` is what I think was leading to recurring issues with our existing wheel building setup, with that option being silently ignored and the extension ultimately not being built.

WIP tag while I verify the GitHub actions are working as my local test have.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -13,6 +13,7 @@ on:
     paths:
       - darshan-util/**
       - include/*
+  workflow_dispatch:
 
 jobs:
   get_commit_message:
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -1,0 +1,36 @@
+name: Build Wheels
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - darshan-util/**
+      - include/**
+  pull_request:
+    branches:
+      - main
+    paths:
+      - darshan-util/**
+      - include/*
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.5.0
+        # to supply options, put them in 'env', like:
+        # env:
+        #   CIBW_SOME_OPTION: value
+
+      - uses: actions/upload-artifact@v2
+        with:
+          path: ./wheelhouse/*.whl
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -15,6 +15,26 @@ on:
       - include/*
 
 jobs:
+  get_commit_message:
+    name: Get commit message
+    runs-on: ubuntu-latest
+    if: github.repository == 'darshan-hpc/darshan'
+    outputs:
+      message: ${{ steps.commit_message.outputs.message }}
+    steps:
+      - name: Checkout darshan
+        uses: actions/checkout@v3
+        # Gets the correct commit message for pull request
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Get commit message
+        id: commit_message
+        run: |
+          set -xe
+          COMMIT_MSG=$(git log --no-merges -1 --oneline)
+          echo "::set-output name=message::$COMMIT_MSG"
+          echo github.ref ${{ github.ref }}
+
   build_wheels:
     name: Build wheels on ${{ matrix.os }}
     needs: get_commit_message
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -17,6 +17,7 @@ on:
 jobs:
   build_wheels:
     name: Build wheels on ${{ matrix.os }}
+    needs: get_commit_message
     if: >-
       contains(needs.get_commit_message.outputs.message, '[wheel build]') ||
       github.event_name == 'workflow_dispatch'
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -27,9 +27,8 @@ jobs:
 
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.5.0
-        # to supply options, put them in 'env', like:
-        # env:
-        #   CIBW_SOME_OPTION: value
+        with:
+          package-dir: ./darshan-util/pydarshan
 
       - uses: actions/upload-artifact@v2
         with:
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -17,6 +17,9 @@ on:
 jobs:
   build_wheels:
     name: Build wheels on ${{ matrix.os }}
+    if: >-
+      contains(needs.get_commit_message.outputs.message, '[wheel build]') ||
+      github.event_name == 'workflow_dispatch'
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, macos-latest]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,35 @@
...
+      - main
+    paths:
+      - darshan-util/**
+      - include/**
+  pull_request:
+    branches:
+      - main
+    paths:
+      - darshan-util/**
+      - include/*
```
tylerjereddy, line 15: I just wanted to make a suggestion here based on something I saw in another project:

> Numpy currently uses `cibuildwheel <https://https://cibuildwheel.readthedocs.io/en/stable/>`_
in order to build wheels through continuous integration services. To save resources, the
cibuildwheel wheel builders are not run by default on every single PR or commit to main.

> If you would like to test that your pull request do not break the wheel builders,
you may either append ``[wheel build]`` to the end of the commit message of the commit...

and then the CI config has something like this:

> .github/workflows/wheels.yml:      contains(needs.get_commit_message.outputs.message, '[wheel build]')

maybe you'll want to defer that to another issue, but wanted to mention that

darshan-util/pydarshan/pyproject.toml
```diff
@@ -3,3 +3,26 @@ requires = [
...
+skip = [
+    "*musllinux*",
+    "*i686*"
+]
+test-requires = [
+    "pytest",
+    "lxml",
+    "matplotlib<3.5",
+    "importlib_resources;python_version<'3.9'"
+]
```
tylerjereddy, line 27: It may be worth a comment or issue or something like that to remind us to make the logs repo available for testing here eventually, so that wheels get a thorough testing treatment before release.

We can't just add the project name here, as you likely realized per https://github.com/darshan-hpc/darshan-logs/issues/15

.github/workflows/build_wheels.yml
shanedsnyder: Yeah, I like that idea and planned on asking for feedback for that specifically, because the wheel building process takes a _long_ time (especially for PyPy builds for whatever reason). Building them on demand (maybe on big changes and at release time) makes a of sense. I'll try to address that in this PR once I'm sure things are running properly otherwise. Thanks for the pointer.

darshan-util/pydarshan/pyproject.toml
shanedsnyder: Good idea. You could probably make `cibuildwheel` just clone the repo and install it from source, but we should just make a PyPI package probably. I'll make an issue to track that.

shanedsnyder:
This is kind of working, but still some things left to resolve:

- PyPy 3.9 builds for macosx are failing due to not finding openblas/lapack scipy dependencies (despite brew installing them)
- PyPy 3.9 builds for manylinux were failing at some point, causing Python to abort (segfault?) in some obscure callstack
- restrict wheel builds to on-demand along lines mentioned by Tyler in previous comment
- open issue to remind us to integrate darshan-logs package once it's available on PyPI

There's also something that's slightly confusing to me, in that the testing phase is taking a really long time (30+ minutes on PyPy builds), presumably because it looks like most (all?) of the dependencies are being built from source (which is also causing issue 1 above). Does that make sense? Not sure why they can't just be installed from binary packages on the index just as I would do if running manual tests locally.

darshan-util/pydarshan/pyproject.toml
```diff
@@ -3,3 +3,40 @@ requires = [
...
+test-command = "pytest {package}"
+
+[tool.cibuildwheel.linux]
```
jakobluettgau, line 22: From the example: https://cibuildwheel.readthedocs.io/en/stable/options/#configuration-file
```
[tool.cibuildwheel.linux]
repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel}"
```

And more info here:
https://cibuildwheel.readthedocs.io/en/stable/options/#repair-wheel-command

darshan-util/pydarshan/pyproject.toml
```diff
@@ -3,3 +3,40 @@ requires = [
...
+    "yum install -y blas-devel lapack-devel",
+    "yum install -y libxslt-devel libxml2-devel",
+    "yum install -y libjpeg libjpeg-devel",
+    "git submodule update --init",
+    "./prepare.sh",
+    "./configure --disable-darshan-runtime --enable-apxc-mod --enable-apmpi-mod",
+    "make install"
+]
+
+[tool.cibuildwheel.macos]
```
jakobluettgau, line 33: From the example: https://cibuildwheel.readthedocs.io/en/stable/options/#configuration-file
```
[tool.cibuildwheel.macos]
repair-wheel-command = [
  "delocate-listdeps {wheel}",
  "delocate-wheel --require-archs {delocate_archs} -w {dest_dir} {wheel}",
]
```

And more info here:
https://cibuildwheel.readthedocs.io/en/stable/options/#repair-wheel-command

jakobluettgau:
One other thing that occurred to me is that, while the submodules initialized and the darshan-utils shared object is build with them enabled, the python files from those are not copied over yet into the pydarshan/darahsn/backend directory, right?

`cd ./darshan-util/pydarshan && make add-modules`

darshan-util/pydarshan/pyproject.toml
shanedsnyder: Thanks for catching that, I'll have a look at this and below and try to add them to the pyprojec.toml

tylerjereddy:
> There's also something that's slightly confusing to me, in that the testing phase is taking a really long time (30+ minutes on PyPy builds), presumably because it looks like most (all?) of the dependencies are being built from source

I'm not sure why we'd even want to offer PyPy wheels? This is done at the base of the ecosystem only in exceptional circumstances where there is a strong evidence of a performance benefit. Have we really done detailed studies justifying the usage of the JIT-compiled version (PyPy) of pydarshan vs. conventional CPython wheels? We're not testing with PyPy in our CI normally, and we don't do benchmarks with PyPy at the moment either.

At the moment, it looks like wheel builds are confined to PyPy (`build = "pp39*"`) in this PR, but I'd suggest just avoiding PyPy altogether and using the conventional interpreted CPython implementation of Python (`cp*` type builds--an exclusion list like this for example: `skip = "pp* *-manylinux_i686 *_ppc64le *_s390x *-musllinux*"`; I did something like that for my early testing over in https://github.com/scipy/scipy/pull/15925).

As to the original question--yes, it makes sense those builds take forever and have problems because many packages like SciPy won't provide wheels--we found that PyPy support was just too great of a burden (it is an entirely different implementation of Python itself), even with a PyPy dev in our community.

shanedsnyder:
Okay, I think I've addressed all comments here and am content with where things are at. My most recent commit updates our PyDarshan release checklist to make clear how we could take wheels generated from actions and upload them to PyPI.

I dropped the WIP tag, but I'll leave it to you guys whether there's more to address here. If you do approve the PR and merge it, please do make sure to squash into a single commit -- a lot of these commits were just trying to get CI to run and things like that, so don't want all the noise in the git history.

tylerjereddy:
When I run with this branch locally: `cibuildwheel --platform linux`

I get an error:

```
...
Installed:
  libjpeg-turbo.x86_64 0:1.2.90-8.el7 libjpeg-turbo-devel.x86_64 0:1.2.90-8.el7

Complete!

                                                             ✕ 32.95s
Error: Command ['sh', '-c', 'yum install -y blas-devel lapack-devel && yum install -y libxslt-devel libxml2-devel && yum install -y libjpeg libjpeg-devel && git submodule update --init && ./prepare.sh && ./configure --disable-darshan-runtime --enable-apxc-mod --enable-apmpi-mod && make install'] failed with code 128. 
```

tylerjereddy:
A bit more info here, `git submodule update --init` is the command that causes that particular issue. A bit of a guess, but maybe there is a bit of magic happening with the `package-dir` variable in CI vs. what happens when I run from within `pydarshan` folder locally? Different effective working directory/context or something?

shanedsnyder:
Oh, maybe it's this: you need to run cibuildwheel from the top-level directory in the repo so it can properly build darshan-util. I've been running it like:

`cibuildwheel --platform linux darshan-util/pydarshan`

tylerjereddy:
Ok, that works, it may be worth documenting somewhere.

```
5 wheels produced in 16 minutes:
  darshan-3.4.0.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl   856 kB
  darshan-3.4.0.0-cp36-cp36m-manylinux_2_17_x86_64.manylinux2014_x86_64.whl    856 kB
  darshan-3.4.0.0-cp37-cp37m-manylinux_2_17_x86_64.manylinux2014_x86_64.whl    856 kB
  darshan-3.4.0.0-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl     856 kB
  darshan-3.4.0.0-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl     855 kB

```

tylerjereddy:
Ok, I think this can go in now. It is quite similar to some of the upstream configs, like the one used by NumPy, and is a big step forward for our wheel/release process.

gh-744 covers my main concern for a follow-up; one other question is whether we really want `path` directives on the wheel build process? It could be that the release manager will just want to build wheels without changing files at a particular path. I think the manual trigger may work in that case though, not sure.

tylerjereddy (approved):

tylerjereddy:
squash merged in

PR #747, opened 2022-05-20
MAINT: deal with pd deprecation

tylerjereddy:
* treat all `FutureWarning` as errors in the `pytest`
CI checks, and fixup the large number of `FutureWarning`s
coming from usage of `df.append()` by using the more modern
`pd.concat()` instead

* the full testsuite drops from 819 to 48 warnings for me
locally, as a result of this change


shanedsnyder:
LGTM :+1:

Issue #690, opened 2022-03-09
ENH: support for "mangled"/cythonized h5py bindings


tylerjereddy:
`h5py` uses Cython to generate C code, which then calls HDF5 C functions. Although I'm not 100 % sure this is the problem described for HDF5 read-invisibility in gh-663, a sample wrapped symbol that I see in a local build is `__real___pyx_f_4h5py_4defs_H5Dread` instead of `H5Dread`.

I started hacking on trying to capture those `H5D` cythonized read events (see below the fold), but that is getting into the runtime/C binding weeds.

<details>

 ```diff
diff --git a/darshan-runtime/lib/darshan-hdf5.c b/darshan-runtime/lib/darshan-hdf5.c
index 50d9694..dd18a4b 100644
--- a/darshan-runtime/lib/darshan-hdf5.c
+++ b/darshan-runtime/lib/darshan-hdf5.c
@@ -40,7 +40,7 @@ DARSHAN_FORWARD_DECL(H5Dcreate1, hid_t, (hid_t loc_id, const char *name, hid_t t
 DARSHAN_FORWARD_DECL(H5Dcreate2, hid_t, (hid_t loc_id, const char *name, hid_t dtype_id, hid_t space_id, hid_t lcpl_id, hid_t dcpl_id, hid_t dapl_id));
 DARSHAN_FORWARD_DECL(H5Dopen1, hid_t, (hid_t loc_id, const char *name));
 DARSHAN_FORWARD_DECL(H5Dopen2, hid_t, (hid_t loc_id, const char *name, hid_t dapl_id));
-DARSHAN_FORWARD_DECL(H5Dread, herr_t, (hid_t dataset_id, hid_t mem_type_id, hid_t mem_space_id, hid_t file_space_id, hid_t xfer_plist_id, void * buf));
+DARSHAN_FORWARD_DECL(__pyx_f_4h5py_4defs_H5Dread, herr_t, (hid_t dataset_id, hid_t mem_type_id, hid_t mem_space_id, hid_t file_space_id, hid_t xfer_plist_id, void * buf));
 DARSHAN_FORWARD_DECL(H5Dwrite, herr_t, (hid_t dataset_id, hid_t mem_type_id, hid_t mem_space_id, hid_t file_space_id, hid_t xfer_plist_id, const void * buf));
 #ifdef DARSHAN_HDF5_VERS_1_10_PLUS
[...]

PR #750, opened 2022-05-31
WIP, ENH: allow H5D_READS without opens

tylerjereddy:
Very much a work in progress...

* allow the darshan runtime to increment `H5D_READS` without
assuming that the conventional `H5D` API was used to open
a file, as long as the return code from `H5Dread()` is non-negative.
(after all, how else could this return `0` if the file handle/dataset info
isn't properly available through some means that our runtime
cannot track?)

* this is based on a debugging/analysis of the `h5py` control
flow, which does not appear to call any of the conventional
`H5D` open/creation API functions (`H5F` is still accessed
through conventional means); instead, it appears that control
flow proceeds though the proxy function `dset_rw`, and darshan
then captures a call to `H5Dread()`

* considerably more work will be needed here I think, because
various other f(counters) are assumed to be valid but likely
are not in this case (like the timing for `H5D` events that
didn't happen);

* I also thought about trying to capture i.e., `dset_rw()`,
but it seems to me that there might be a sustainability issue
here--i.e., what if Rust and other languages were to have
"native" bindings that used yet another route, etc.? I guess
the question is whether we'd be happy to increment an open
event that is not using a conventional `H5D` API--my current intuition
is that this would not convey what is going on very accurately
(i.e., it is more accurate to say that darshan captured the read event
but didn't see the `H5D` open event through the conventional API)

* using this reproducer:

```python
from mpi4py import MPI
import numpy as np
from numpy.testing import assert_array_equal
import h5py

def main():
    comm = MPI.COMM_WORLD
    rank = comm.Get_rank()

    # round trip HDF5 IO test
    file_path = f"./test_rank_{rank}.hdf5"
    n_bytes = 10 * (rank + 1)

    bytes_to_write = np.ones(shape=n_bytes, dtype=np.int8)

    #with h5py.File(file_path, "w") as f:
        #f.create_dataset("dataset", data=bytes_to_write)

    print("before h5py.File()! on rank:", rank)
    with h5py.File(file_path, "r") as g:
        print("after h5py.File()! on rank:", rank)
        retrieved_data = np.asarray(g['dataset'])
        print("after retrieved_data on rank:", rank)
        assert_array_equal(retrieved_data, bytes_to_write)

if __name__ == "__main__":
    main()
```

* on this branch we now pick up the `H5D` read event with 3 ranks,
though the dataset ids or hashes darshan assigns need checking:

```
   rank                    id  H5F_OPENS  H5F_FLUSHES  H5F_USE_MPIIO
0     0   3991442081443699295          1            0              0
1     1  13842144362321976876          1            0              0
2     2  13503993907200906631          1            0              0
   rank                    id  H5D_OPENS  H5D_READS  H5D_WRITES  H5D_FLUSHES  H5D_BYTES_READ  ...  H5D_CHUNK_SIZE_D5  H5D_USE_MPIIO_COLLECTIVE  H5D_USE_DEPRECATED  H5D_FASTEST_RANK  H5D_FASTEST_RANK_BYTES  H5D_SLOWEST_RANK  H5D_SLOWEST_RANK_BYTES
0     0  17846303746652554285          0          1           0            0               0  ...                  0                         0                   0                 0                       0                 0                       0
1     1  13677892646774148885          0          1           0            0               0  ...                  0                         0                   0                 0                       0                 0                       0
2     2  17181974452690724755          0          1           0            0               0  ...                  0                         0                   0                 0                       0                 0                       0

[3 rows x 96 columns]
```

* on `main`:

```
   rank                    id  H5F_OPENS  H5F_FLUSHES  H5F_USE_MPIIO
0     0   3991442081443699295          1            0              0
1     1  13842144362321976876          1            0              0
2     2  13503993907200906631          1            0              0
  File "/home/tyler/LANL/rough_work/darshan/issue_690/check_for_h5dread.py", line 13, in <module>
    main(log_path=log_path)
  File "/home/tyler/LANL/rough_work/darshan/issue_690/check_for_h5dread.py", line 9, in main
    print(report.records["H5D"].to_df()["counters"])
KeyError: 'H5D'
```

* using the probe code:

```python
import sys

import darshan

def main(log_path: str):
    print(report.records["H5F"].to_df()["counters"])
    print(report.records["H5D"].to_df()["counters"])

if __name__ == "__main__":
    log_path = sys.argv[1]
    main(log_path=log_path)
```

### DIFF
--- a/darshan-runtime/lib/darshan-hdf5.c
+++ b/darshan-runtime/lib/darshan-hdf5.c
@@ -730,6 +730,44 @@ herr_t DARSHAN_DECL(H5Dread)(hid_t dataset_id, hid_t mem_type_id, hid_t mem_spac
                 rec_ref->dataset_rec->fcounters[H5D_F_READ_TIME],
                 tm1, tm2, rec_ref->last_read_end);
         }
+        else {
+            /* 
+             * Any programming language with its own bindings to HDF5
+             * is, in theory, capable of tracking the HDF5 dataset information
+             * internally and calling i.e., H5Dread() without a prior call to
+             * H5Dopen*(), H5Dcreate*(), etc. when working with an already-created
+             * HDF5 file.
+             *
+             * Therefore, if darshan intercepts H5Dread() without a prior HDF5 dataset
+             * "registration," we record the event and decrement H5D_OPENS because
+             * of the unusual route to dataset access--there is no API "open" event
+             * and we artificially use an open start/stop time of 0.0.
+             * 
+             */
+            darshan_core_fprintf(stderr, "start special branch for h5py H5Dread!\n");
+            hid_t dtype_id = H5Dget_type(dataset_id);
+            hid_t space_id = H5Dget_space(dataset_id);
+            hid_t dcpl_id = H5Dget_create_plist(dataset_id);
+            darshan_core_fprintf(stderr, "special branch for h5py H5Dread checkpoint A!\n");
+            const char *name = "";
+            darshan_core_fprintf(stderr, "special branch for h5py H5Dread checkpoint A2!\n");
+            H5D_RECORD_OPEN(dataset_id, dataset_id, name, dtype_id, space_id, dcpl_id, 0, 0, 0);
+            darshan_core_fprintf(stderr, "special branch for h5py H5Dread checkpoint B!\n");
+            rec_ref = darshan_lookup_record_ref(hdf5_dataset_runtime->hid_hash,
+                &dataset_id, sizeof(hid_t));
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Note that I've also queried about the control flow in `h5py` upstream: https://github.com/h5py/h5py/issues/2107

tylerjereddy:
We've now heard back from the folks upstream and I've added some detailed thoughts here: https://github.com/darshan-hpc/darshan/issues/690#issuecomment-1146837894

tylerjereddy:
I'll close this and leave it to the team given the above-linked discussion--my main objective was to emphasize that I think we should have a more modular way to capture HDF5 symbols that is not reliant on specific API event chains.

PR #751, opened 2022-06-08
Add support for H5Oopen

shanedsnyder:
It's possible for HDF5 users to open datasets using a more generic `H5Oopen` call (see discussion linked in #690), which is the case for h5py read path, for instance.

Since Darshan does not intercept `H5Oopen`, it does not store any state about the open dataset, causing subsequent dataset read/write calls to bail out, since they do not know which open dataset to attribute the I/O activity to.

This PR adds new wrappers for `H5Oopen`/`H5Oclose` to support instrumentation of datasets accessed in this manner. Code is nearly identical to `H5Dopen`/`H5Dclose`, sans an additional check here to only store state on opened dataset objects.

### DIFF
--- a/darshan-runtime/lib/darshan-hdf5.c
+++ b/darshan-runtime/lib/darshan-hdf5.c
@@ -48,7 +48,7 @@ DARSHAN_FORWARD_DECL(H5Dflush, herr_t, (hid_t dataset_id));
 DARSHAN_FORWARD_DECL(H5Dclose, herr_t, (hid_t dataset_id));
 
 /* H5Oopen prototype -- can be used as an indirect way of opening a dataset */
-DARSHAN_FORWARD_DECL(H5Oopen, herr_t, (hid_t loc_id, const char *name, hid_t lapl_id));
+DARSHAN_FORWARD_DECL(H5Oopen, hid_t, (hid_t loc_id, const char *name, hid_t lapl_id));
 DARSHAN_FORWARD_DECL(H5Oclose, herr_t, (hid_t object_id));
 
 /* structure that can track i/o stats for a given HDF5 file record at runtime */
--- a/darshan-runtime/lib/darshan-hdf5.c
+++ b/darshan-runtime/lib/darshan-hdf5.c
@@ -926,7 +926,6 @@ hid_t DARSHAN_DECL(H5Oopen)(hid_t loc_id, const char *name, hid_t lapl_id)
     hid_t dcpl_id;
     double tm1, tm2;
     hid_t ret;
-    herr_t herr;
 
     MAP_OR_FAIL(H5Oopen);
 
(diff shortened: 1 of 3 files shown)


darshan-util/pydarshan/pyproject.toml
```diff
@@ -3,3 +3,24 @@ requires = [
...
+skip = [
+    "*musllinux*",
+    "*i686*"
+]
+test-requires = [
+    "pytest",
+    "lxml",
+    "importlib_resources;python_version<'3.9'"
+]
+test-command = "pytest {package}"
```
tylerjereddy, line 26: may need a rebase to filter out these changes already present in `main`

darshan-runtime/lib/darshan-hdf5.c
```diff
@@ -922,6 +926,103 @@ herr_t DARSHAN_DECL(H5Dclose)(hid_t dataset_id)
     return(ret);
 }
 
+/* NOTE: we have to intercept this generice H5Oopen call, since it allows
```
tylerjereddy, line 929:
```suggestion
/* NOTE: we have to intercept this generic H5Oopen call, since it allows
```

tylerjereddy (commented):
I'll give this a test with the toy `h5py` example in a minute and report back.

darshan-util/pydarshan/pyproject.toml
shanedsnyder: oops, my mistake. this is fixed now.

shanedsnyder:
Heh, okay, thanks for checking. I suppose that's better than "no affect at all", but not quite as good as "works as expected".

I'll try to think more on this, but ultimately might just need to get everything built and step through things. Hopefully it's just something subtle I missed. I don't think anything I did should affect the ID returned by `H5Oopen`, it just may be that some of the additional HDF5 calls I'm making within the wrapper are somehow tripped up.

tylerjereddy:
@shanedsnyder see gh-753 for end-to-end functional CI testing that reproduces the runtime error I note above (when cherry-picked to this branch); that seems useful for building up guards against runtime regressions I think

shanedsnyder:
FYI, I just pushed new commits here that address a couple of bugs I encountered in a simple test program I wrote to try to mimc h5py pattern. Would you mind trying again now? I suspect one of them for sure was causing the traceback you shared.

I'll also try to have a closer look at the CI PR soon.

tylerjereddy:
@shanedsnyder See the result of cherry-picking the commits from my regression test PR into the latest version of your branch here: https://github.com/tylerjereddy/darshan/pull/16

Looks like the regression test is passing there, so that's a good sign.

tylerjereddy:
Looks like the new runtime tests are passing again with the fix, that's nice

shanedsnyder:
> We probably want to just go ahead and knock out support for the following calls while we're here, too:
> 
>     * `H5Oopen_by_addr()`
> 
>     * `H5Oopen_by_idx()`
> 
>     * `H5Oopen_by_token()`

I just pushed in support for these. They each just require one small additional step to the original `H5Oopen` wrapper in that the dataset name must be queried via HDF5 API (`H5Iget_name`), since it's not given in the arguments to the call.

There's one snag, though. `H5Oopen_by_token` is only available in 1.12.0+ versions of HDF5, so we need to conditionally compile in support for that. The problem is that we need to tackle #754 first since our current autoconf logic only tells us whether the configured version is 1.10+ or not (i.e., we need more fine-grained version info). That should be easy to straighten out, so I'll take care of that first and then rebase this work on top of it to ensure it's properly handled.

shanedsnyder:
(You can see the outstanding issue in question in current test failures -- looks like whatever h5py we are using is not 1.12+ so hits the same issue.)

carns (approved):
Looks good

Issue #752, opened 2022-06-09
TST: unfavorable interaction with pytest-mpi


tylerjereddy:
Feature branch `treddy_functional_testing`, which I've pushed up to my fork, contains an example `pytest-mpi` test that succeeds on its own but fails when under simultaneous instrumentation from both `darshan` runtime and `pytest-mpi`.

This works:

- `mpirun -n 1 python -m pytest --with-mpi darshan-test/python_runtime_tests.py`

```
1 passed in 0.17s
```

This fails:
- `mpirun -n 1 -x LD_PRELOAD=/home/tyler/darshan_install/lib/libdarshan.so -x DARSHAN_LOGPATH=/home/tyler python -m pytest --with-mpi darshan-test/python_runtime_tests.py`

```
================================================================================================================================ test session starts =================================================================================================================================
platform linux -- Python 3.10.4, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/tyler/github_projects/darshan
plugins: forked-1.4.0, mpi-0.6, xdist-2.5.0
collected 1 item                                                                                                                                                                                                                                                                     

darshan-test/python_runtime_tests.py --------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[54384,1],0]
  Exit code:    127
--------------------------------------------------------------------------

```

The `127` code is for a  missing executable I think, so may not be too hard to debug, though I'll probably just avoiding using `pytest-mpi` here for the moment.

PR #753, opened 2022-06-09
CI, TST: early draft of functional testing

tylerjereddy:
* this adds end-to-end functional CI testing from
the darshan runtime monitoring MPI IO activity
right through to the Python bindings checking
for the expected result; you can run this locally (containerized)
with the `act` tool with i.e.,
`act -j test_runtime`

* when cherry-picked into gh-751, this reproduces
the exact error I was hoping CI could catch as
described here:
https://github.com/darshan-hpc/darshan/pull/751#issuecomment-1150197393

* otherwise, running this locally produces the expected failure
to capture `H5D` events when reading with `h5py`, with
only the write, but not the read event going through
the expected API path until Shane's PR (gh-751) is
ready

```
| >           assert h5d_df_counters["H5D_OPENS"].to_numpy() == n_ranks * 2
| E           assert array([1]) == (1 * 2)
| E            +  where array([1]) = <bound method IndexOpsMixin.to_numpy of 0    1\nName: H5D_OPENS, dtype: int64>()
| E            +    where <bound method IndexOpsMixin.to_numpy of 0    1\nName: H5D_OPENS, dtype: int64> = 0    1\nName: H5D_OPENS, dtype: int64.to_numpy
|
| darshan-test/python_runtime_tests.py:50: AssertionError
```

* there are many reasons I've opted not to use `pytest-mpi` at this
time, and gh-752 is only one of them

* if you'd like me to write an example test using C code, that should
be quite doable as well, we'd just have the `pytest` test compile
the source appropriately and then use the same approach; developing
this infrastructure was fairly time consuming though so I may want to
defer that to a follow-up if it is "ok"

* I'm not sure how many ranks we can safely use simultaneously
on the same free open source virtual machines, but this still
seems like a pretty big step up from "no runtime testing whatsoever"
in the CI

* this also opens up the possibility for i.e., using `gcov` to
produce reports of coverage of the C runtime code in CI, and so on..

* we may benefit from caching some steps.. building `h5py` from
source is time consuming in my local tests with `act`

### DIFF
--- a/.github/workflows/runtime_ci.yml
+++ b/.github/workflows/runtime_ci.yml
@@ -23,8 +23,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-          apt-get update -y
-          apt-get install -y hdf5-tools libhdf5-openmpi-dev openmpi-bin
+          sudo apt-get update -y
+          sudo apt-get install -y hdf5-tools libhdf5-openmpi-dev openmpi-bin
           python -m pip install --upgrade pip
           python -m pip install --upgrade pytest mpi4py
           # the organization of HDF5 files from apt isn't
--- a/.github/workflows/runtime_ci.yml
+++ b/.github/workflows/runtime_ci.yml
@@ -48,6 +48,7 @@ jobs:
           make install
       - name: Install darshan-util
         run: |
+          export DARSHAN_INSTALL_PATH=$PWD/darshan_install
           cd darshan-util
           mkdir build
           cd build
@@ -60,6 +61,7 @@ jobs:
           python -m pip install .
       - name: Test with pytest
         run: |
+          export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/darshan_install/lib
           export DARSHAN_INSTALL_PATH=$PWD/darshan_install
           export DARSHAN_ROOT_PATH=$PWD
           export HDF5_LIB=/usr/lib/x86_64-linux-gnu/hdf5/openmpi/libhdf5.so
--- a/.github/workflows/runtime_ci.yml
+++ b/.github/workflows/runtime_ci.yml
@@ -1,0 +1,66 @@
+name: Runtime Testing
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test_runtime:
+    strategy:
+      matrix:
+        platform: [ubuntu-latest]
+        python-version: ["3.10"]
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          apt-get update -y
+          apt-get install -y hdf5-tools libhdf5-openmpi-dev openmpi-bin
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade pytest mpi4py
+          # the organization of HDF5 files from apt isn't
+          # ideal for usage with darshan, but we can cheat
+          # a little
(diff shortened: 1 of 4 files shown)


tylerjereddy:
One annoying thing here is that a few commands seem to require `sudo` in GitHub Actions CI proper, but not locally with `act`. Hopefully there is a way to have them both working symmetrically eventually, but for now I'll prioritize the CI version. There is a way to check if you are running locally with a variable so we could add that later maybe, to use/not use `sudo` depending on that.

tylerjereddy:
Ok, CI seems to be running properly here now for the runtime testing, and picking up the current bug in the runtime. We could temporarily `xfail` the test until gh-751 is ready perhaps.

The 6 minute CI time is pretty reasonable as well, considering how much stuff gets compiled/downloaded.

.github/workflows/runtime_ci.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y hdf5-tools libhdf5-openmpi-dev openmpi-bin
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade pytest mpi4py
+          # the organization of HDF5 files from apt isn't
+          # ideal for usage with darshan, but we can cheat
+          # a little
+          sudo cp /usr/include/hdf5/openmpi/*.h /usr/include
```
shanedsnyder, line 33: you don't have to change this, but i confirmed you can work around the header location issue above by just adding `/usr/include/hdf5/openmpi/` to C_INCLUDE_PATH, i.e.,:

`export C_INCLUDE_PATH=$C_INCLUDE_PATH:/usr/include/hdf5/openmpi/`

in the "Install darshan-runtime step". Still an annoying additional step, but maybe less hack-y.

shanedsnyder (changes requested):

shanedsnyder:
I'm real late to the game here, but just had a look and think this looks great! I was able to get it working on my laptop using `act` and things appear to be working as expected.

I reviewed a couple of things, both really nitpicky, but at least the job ID env variable thing should probably be fixed. I'd probably be fine to merge after that.

A few additional comments:
- I'm fine with keeping the tests as failed for now rather than explicitly xfailing -- I plan to merge in the PR that fixes the underlying issue soon after this is merged, so that will avoid some churn in the repo.
- What exactly do you mean by caching some steps in the original PR description? Could that be used to cache some of the shared steps (like building darshan-util, installing pydarshan) from this CI workflow and the generic CI workflow we already have? Or did you have something else in mind? It would be nice if our different CI workflows could either reuse some of the same cached artifacts, or at the very least, could reuse some of the shared install scripts. That's probably out-of-scope here, I'm just asking to see if it's something worth following up on in an additional issue.

shanedsnyder:
> One annoying thing here is that a few commands seem to require `sudo` in GitHub Actions CI proper, but not locally with `act`. Hopefully there is a way to have them both working symmetrically eventually, but for now I'll prioritize the CI version. There is a way to check if you are running locally with a variable so we could add that later maybe, to use/not use `sudo` depending on that.

I assume it's not really a big deal to let `act` run these commands locally using `sudo`, since it's all containerized? In fact, I kind of expected `sudo` would need to be used with `act` since, in my experience, you usually need to use `sudo` pretty liberally inside of containers?

tylerjereddy:
I think I pushed in changes for the two requests from Shane.

- re: caching, yes I meant https://github.com/marketplace/actions/cache for example; a number of OSS projects use this to save time for stuff that gets repeated; not sure about cross-job cache, but certainly saving time on a given job; there's an example in SciPy (https://github.com/scipy/scipy/blob/main/.github/workflows/linux_meson.yml) though it is sufficiently complex that I'd probably agree we should do that in a separate PR
- re: `sudo` and `act`; ok, it works for me locally now too, so I guess that's not a problem anymore--I think my docker stuff got all messed up a few months ago when LANL changed some security settings..

shanedsnyder (approved):

PR #756, opened 2022-06-14
WIP, TST: Regression test for runtime integer overflow in HEATMAP

tylerjereddy:
* this builds on the new testing infrastructure drafted in gh-753, which should be
reviewed/dealt with before this PR

* add a C-language regression test based on Phil's comment here:
https://github.com/darshan-hpc/darshan/pull/737#issuecomment-1126148539

* at the time of writing, however, the test passes--there is no evidence of
overflow--I'm open to suggestions from the "C folks" on the team..

### DIFF
--- a/darshan-test/c_mpi_progs/runtime_prog_issue_730.c
+++ b/darshan-test/c_mpi_progs/runtime_prog_issue_730.c
@@ -9,7 +9,7 @@ int main(int argc, char** argv)
 {
     int  rank;
     int  fd;
-    char c = 'a';
+    char c[5000000];
     int  ret;
     char template[] = "testfileXXXXXX";
 
@@ -22,8 +22,8 @@ int main(int argc, char** argv)
         perror("mkstemp");
         return (-1);
     }
-    /* write one byte, to help ensure heatmap is instantiated */
-    ret = write(fd, &c, 1);
+    /* write bytes, to help ensure heatmap is instantiated */
+    ret = write(fd, &c, 5000000);
     if (ret < 0) {
         perror("write");
         return (-1);
@@ -36,8 +36,8 @@ int main(int argc, char** argv)
             return (-1);
         }
     }
-    /* write another byte, to help ensure heatmap is instantiated */
-    ret = write(fd, &c, 1);
+    /* write bytes, to help ensure heatmap is instantiated */
+    ret = write(fd, &c, 5000000);
     if (ret < 0) {
         perror("write");
         return (-1);
--- a/darshan-test/c_mpi_progs/runtime_prog_issue_730.c
+++ b/darshan-test/c_mpi_progs/runtime_prog_issue_730.c
@@ -3,21 +3,49 @@
 #include <stdio.h>
 #include <mpi.h>
 #include <stdlib.h>
+#include <errno.h>
 
-
-int main(int argc, char **argv)
+int main(int argc, char** argv)
 {
-   int rank;
-   int fd;
-   char c;
-   
-   MPI_Init(&argc, &argv);
-   MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+    int  rank;
+    int  fd;
+    char c = 'a';
+    int  ret;
+    char template[] = "testfileXXXXXX";
 
-   fd = open("test.txt", O_RDONLY);
-   read(fd, &c, 1);
-   for(int i = 0; i < 5000; i++) {
-       read(fd, &c, 0);
-   }
-   MPI_Finalize();
+    MPI_Init(&argc, &argv);
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+    /* create temporary file in current directory */
+    fd = mkstemp(template);
+    if (fd < 0) {
+        perror("mkstemp");
+        return (-1);
+    }
+    /* write one byte, to help ensure heatmap is instantiated */
+    ret = write(fd, &c, 1);
+    if (ret < 0) {
+        perror("write");
+        return (-1);
+    }
+    /* issue a bunch of zero byte reads */
+    for (int i = 0; i < 5000; i++) {
+        ret = read(fd, &c, 1);
+        if (ret < 0) {
+            perror("read");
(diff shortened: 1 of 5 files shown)


darshan-test/c_mpi_progs/runtime_prog_issue_730.c
```diff
@@ -0,0 +1,23 @@
...
+   MPI_Init(&argc, &argv);
+   MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+   fd = open("test.txt", O_RDONLY);
+   read(fd, &c, 1);
+   for(int i = 0; i < 5000; i++) {
+       read(fd, &c, 0);
+   }
+   MPI_Finalize();
+}
```
tylerjereddy, line 23: @carns @shanedsnyder Here is an example of a "C-based" regression test in the end-to-end functional testing infrastructure. At the time of writing, this passes instead of causing the issue Phil suggested here:
https://github.com/darshan-hpc/darshan/pull/737#issuecomment-1126148539

I'm open to suggestions, or feel free to push in to this PR if you think this can be made to repro the `0/0`.

carns:
I didn't push this straight into the PR because I'm not exactly sure how to run it, but this variation reliably reproduces the problem on my laptop:

```
#include <fcntl.h>
#include <unistd.h>
#include <stdio.h>
#include <mpi.h>
#include <stdlib.h>
#include <errno.h>

int main(int argc, char** argv)
{
    int  rank;
    int  fd;
    char c = 'a';
    int  ret;
    char template[] = "testfileXXXXXX";

    MPI_Init(&argc, &argv);
    MPI_Comm_rank(MPI_COMM_WORLD, &rank);

    /* create temporary file in current directory */
    fd = mkstemp(template);
    if (fd < 0) {
        perror("mkstemp");
        return (-1);
    }
    /* write one byte, to help ensure heatmap is instantiated */
    ret = write(fd, &c, 1);
    if (ret < 0) {
        perror("write");
        return (-1);
    }
    /* issue a bunch of zero byte reads */
    for (int i = 0; i < 5000; i++) {
        ret = read(fd, &c, 1);
        if (ret < 0) {
            perror("read");
            return (-1);
        }
    }
    /* write another byte, to help ensure heatmap is instantiated */
    ret = write(fd, &c, 1);
    if (ret < 0) {
        perror("write");
        return (-1);
    }

    /* close and unlink test file */
    close(fd);
    unlink(template);

    MPI_Finalize();
}
```

It is self-contained and makes its own temporary file to access.  I'm not exactly sure what the key element is/was, but I also added some error checking and wrote a few bytes to the file (rather than read) to make sure that some of the bins have valid positive values.  That might make a difference when the heatmaps bin widths are normalized at the end of execution?

@tylerjereddy can you help integrating this test case in this PR and see if it does what you need?

carns:
Oh, and just to clarify, this is what I see in the resulting log to indicate a problem; I haven't looked to see if the test harness is checking for this or not:

```
> ~/working/install/bin/darshan-parser ~/carns_runtime_prog_issue_730_id167563-167563_6-16-51714-17086941444664511151_1.darshan |grep HEATMAP_READ_BIN
HEATMAP	0	16592106915301738621	HEATMAP_READ_BIN_0	-9223372036854775808	heatmap:POSIX	UNKNOWN	UNKNOWN
```
(note the large negative value)

FWIW I would have thought this would have produced a floating point exception, but I guess that is compiler/system dependent.

tylerjereddy:
Hmm, the Python layer showns two written bytes and no read bytes, and only a single time bin, so that's why there's no assertion triggered.

```
----------------------------- Captured stdout call -----------------------------
*** hmap_df_write:
       (0.0, 0.1]
rank            
0              2
*** hmap_df_read:
       (0.0, 0.1]
rank            
0              0
*** hmap_df:
       (0.0, 0.1]
rank            
0              2
```

The next steps for me are probably to confirm that I can reproduce the `darshan-parser` detection of the overflow, and then try to figure out why it doesn't get propagated through. There may be other transformations in getting to the Python layer such that an end-to-end pass-through of the overflow is even trickier.

tylerjereddy:
I don't see the negative number in the `darshan-parser` output here: https://github.com/tylerjereddy/darshan/runs/6927438842?check_suite_focus=true (although it is hard to read with the tabs/newlines escaped like that).

Looks like the pass-through to the Python layer is genuinely `0` from `darshan-parser`:

`HEATMAP\t0\t16592106915301738621\tHEATMAP_READ_BIN_0\t0`

Is your compile line just a plain `mpicc source.c` or similar? More optimization flags or something?

tylerjereddy:
@carns ok, the test failure is reproduced now--you can see the assertion with the integer overflow clearly in the CI testsuite traceback. As we discussed on Slack, I had to use a separate compilation of the runtime for that specific test where `CLOCK_REALTIME_COARSE` is substituted in to slow the timers down a bit.

While this is a bit awkward, it isn't too terrible I guess. Some runtime bugs are indeed likely going to require some creativity to get regression tests for.

Note that the approach of using `/dev/shm` was not sufficient to zero out the timers in my hands.

tylerjereddy:
I should mention, I did a bunch of fast iterating on this and it could probably use some cleanup, but I am pretty confident it is reproducing the issue.

carns:
That's interesting :)  If it works, it works.  I can't think of any better way to force the behavior.  I considered that we could preload an alternative clock_gettime() implementation in front of the real one, but that's probably even more fragile.

I guess we need to re-kick this test now that Shane merged the fix, to confirm that it passes clean?  Assuming that's all good then I'm on board with this.

tylerjereddy:
I'll close/reopen to see if it passes now, while I'm stuck with crazy air travel delays in the US..

tylerjereddy:
I think it is passing now--1/2 tests is still expected to fail because of the lower-priority HDF5 stuff I think.

shanedsnyder (approved):

shanedsnyder:
Looks good to me, the only failure is the HDF5 xfail unrelated to this PR. Thanks!

PR #761, opened 2022-06-16
make `cuserid()` usage opt-in

### LINKED ISSUE #699: cuserid() calls should be opt-in, due to recurring crashes on Crays
Darshan runtime library will first check `cuserid()` for the username to associate the log file with, then fallback to `getenv("LOGNAME")`, before falling back to an integer UID. Calls to `cuserid()` can already be avoided via configure arg `--disable-cuserid`, originally added to work around a bug on an old Cray system from around a decade ago. Recent tests on the Cori Cray system at NERSC trigger similar crashes (seg fault) in statically-linked executables.

To avoid recurring crashes related to `cuserid()` we should only use it as a fallback if the user explicitly asks configure to enable it (`--enable-cuserid`) -- by default, Darshan should call `getenv("LOGNAME")` and if that fails get the integer UID associated with the user.

shanedsnyder:
PR contains the following:

- updates to `darshan-runtime` autoconf and core code to make `cuserid()` usage opt-in
- corresponding changes/simplifications to docs

Fixes #699

### DIFF
--- a/darshan-runtime/configure.ac
+++ b/darshan-runtime/configure.ac
@@ -98,13 +98,13 @@ if test "x$enable_darshan_runtime" = xyes ; then
    fi
 
    AC_ARG_ENABLE([cuserid],
-      [AS_HELP_STRING([--disable-cuserid],
-                      [Disables attempted use of cuserid() at run time])],
-      [], [enable_cuserid=yes]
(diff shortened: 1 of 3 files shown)


carns (approved):
looks good to me!

Issue #698, opened 2022-03-22
MPI_Type_size() is called with MPI_DATATYPE_NULL and breaks some applications


bebosudo:
Hi, a user at NERSC reported that darshan is calling the MPI_Type_size() function with MPI_DATATYPE_NULL; they protect all calls to MPI_Type_size in their application with
```
if (mpi_datatype == MPI_DATATYPE_NULL) {
  size = 0;
} else {
  size = MPI_Type_size(mpi_datatype);
}
```

And he suggests to do the same in Darshan.

The following is an example traceback, but he couldn't provide more details, so I don't have a piece of code to reproduce the issue:

```
3: Test command: /usr/bin/srun "-n" "5" "/path/to/test/app"
3: Test timeout computed to be: 600
3: Rank 1 [Tue Mar 22 10:03:09 2022] [c1-0c0s6n1] Fatal error in PMPI_Type_size: Invalid datatype, error stack:
3: PMPI_Type_size(119): MPI_Type_size(MPI_DATATYPE_NULL) failed
3: PMPI_Type_size(69).: Datatype for argument datatype is a null datatype
3: Rank 2 [Tue Mar 22 10:03:09 2022] [c1-0c0s6n1] Fatal error in PMPI_Type_size: Invalid datatype, error stack:
3: PMPI_Type_size(119): MPI_Type_size(MPI_DATATYPE_NULL) failed
3: PMPI_Type_size(69).: Datatype for argument datatype is a null datatype
3: Rank 3 [Tue Mar 22 10:03:09 2022] [c1-0c0s6n1] Fatal error in PMPI_Type_size: Invalid datatype, error stack:
3: PMPI_Type_size(119): MPI_Type_size(MPI_DATATYPE_NULL) failed
3: PMPI_Type_size(69).: Datatype for argument datatype is a null datatype
3: srun: error: nid00217: tasks 1-3: Aborted
3: srun: launch/slurm: _step_signal: Terminating StepId=56572294.33
3: slurmstepd: error: *** STEP 56572294.33 ON nid00217 CANCELLED AT 2022-03-22T10:03:11 ***
3: srun: error: nid00217: tasks 0,4: Terminated
3: srun: Force Terminated StepId=56572294.33
1/1 Test #3: test_rearr .......................***Failed 5.43 sec
```

I only see occurrences of `PMPI_Type_size` in the darshan source code; I built darshan with support to autoperf with `--enable-apmpi-mod`.
[...]

PR #766, opened 2022-06-28
don't call `MPI_Type_size()` on `MPI_DATATYPE_NULL` in MPI-IO module

shanedsnyder:
As discussed in #698, we should be careful in Darshan not to call `MPI_Type_size()` when applications are using `MPI_DATATYPE_NULL` type.

This PR adds a couple of checks to the two locations in the MPI-IO module where we use this call, to make sure we don't call if the app uses a zero value for `count` or if the app is using `MPI_DATATYPE_NULL`.

There is a similar PR open on the AutoPerf repo which fixes a similar issue in the APMPI module, which was the ultimate source of the reported bug: https://github.com/argonne-lcf/autoperf/pull/15


carns (approved):
looks good to me.

PR #778, opened 2022-07-22
Changed get_module_table() function

saudahmed202:
Remove log size and replaced it with no. of unique file count, bytes read & written in each module

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -244,21 +244,39 @@ class ReportData:
             "Runtime Library Version": [job_data["metadata"]["lib_ver"], ""],
             "Log Format Version": [job_data["log_ver"], ""],
         }
-
         for mod in self.report.modules:
             # retrieve the module version and buffer sizes
             mod_version = self.report.modules[mod]["ver"]
-            # retrieve the buffer size converted to KiB
-            mod_buf_size = self.report.modules[mod]["len"] / 1024
+            # retrieve all the record and convert to pandas df
+            self.report.mod_read_all_records(mod)
+            df = self.report.records[mod].to_df()["counters"]
             # create the key/value pairs for the dictionary
             key = f"{mod} (ver={mod_version}) Module Data"
-            val = f"{mod_buf_size:.2f} KiB"
-            flag = ""
+            #specifying behaviors for each module
+            if mod in {"POSIX", "MPI-IO", "STDIO"}:
+                if mod == "MPI-IO":
+                    read = df["MPIIO_BYTES_READ"].sum()
+                    write = df["MPIIO_BYTES_WRITTEN"].sum()
+                else:
+                    read = df[f"{mod}_BYTES_READ"].sum()
+                    write = df[f"{mod}_BYTES_WRITTEN"].sum()
+                
+                if mod == "H5D":
+                    val = f"{df['id'].nunique()} unique datasets ({read} total bytes read, {write} total bytes written)"
+                    flag = ""
+                else:
+                    val = f"{df['id'].nunique()} unique file ({read} total bytes read, {write} total bytes written)"
+                    flag = ""
+            elif mod in {"H5F", "PNETCDF", "LUSTRE"}:
+                val = f"{df['id'].nunique()} unique file"
+                flag = ""
+            else:
+                val = "-"
+                flag = ""
             if self.report.modules[mod]["partial_flag"]:
                 msg = "Module data incomplete due to runtime memory or record count limits"
                 flag = f"<p style='color:red'>&#x26A0; {msg}</p>"
             module_dict[key] = [val, flag]
-
         # convert the module dictionary into a dataframe
         module_df = pd.DataFrame.from_dict(data=module_dict, orient="index")
         # write out the table in html


tylerjereddy:
Is this related to a particular issue? Or a discussion with the team somewhere? Maybe you can summarize the motivation/context a bit for reviewers.

We usually add test(s) for summary report changes--I can help with that if this is indeed what the team wants.

tylerjereddy:
Looks like some summary report tests are failing, yeah.

shanedsnyder:
I can summarize the context briefly.

I've been working with Saud on how to get a little more useful information included in the module summary portion of this "Darshan Log Information" table -- currently, it only includes the module version, total size, and whether the module has partial data. 

The size in particular is not really useful information. It's in terms of compressed records, so there's no intuition that sort of information gives beyond "this module has _some_ data". These changes are supposed to provide some high-level characteristics about the module, particularly the total number of records in each and, in cases where it makes sense, the total number of bytes read/written using the module.

We probably still need to tweak the exact formatting of the table and make sure tests are passing, but I encouraged Saud to go ahead and submit a PR so he can get some experience with GitHub and working through code review.

shanedsnyder:
Here's some example output Saud shared with me offline:
image

A bit of feedback on my end is we should probably come up with a helper routine to convert raw byte counts into human readable form. E.g., the numbers above I think come out to around 2 TiB (terabytes). Would look cleaner to express in terms of terabytes. I looked briefly and surprisingly don't see any python helpers for this. There are some examples floating around that we could probably borrow from: https://stackoverflow.com/questions/14996453/python-libraries-to-calculate-human-readable-filesize-from-bytes

We could probably get some decent mileage out of a helper routine like that in PyDarshan --- I'm sure we'll use it frequently. Could you try and implement something like that, @saudahmed202 ? Or alternatively, maybe someone else is aware of how to do this using standard python libraries?

darshan-util/pydarshan/darshan/cli/summary.py
```diff
@@ -244,21 +244,39 @@ def get_module_table(self):
...
         }
-
         for mod in self.report.modules:
             # retrieve the module version and buffer sizes
             mod_version = self.report.modules[mod]["ver"]
-            # retrieve the buffer size converted to KiB
-            mod_buf_size = self.report.modules[mod]["len"] / 1024
+            # retrieve all the record and convert to pandas df
+            self.report.mod_read_all_records(mod)
+            df = self.report.records[mod].to_df()["counters"]
```
tylerjereddy, line 252: One problem you'll have here, and it is reflected in some of the test failures, is that the result of `to_df()` can also be a list of DXT records, which are formatted quite differently, and you can't treat that as a simple dictionary.

Iterating through the full set of DXT data may also be time consuming, since it can be a huge Python list of records.

tylerjereddy (commented):
I haven't checked for correctness too much yet, but a few comments to start. Note that we have tests like `test_module_table` that test this exact table, so they will need to be updated of course.

For Shane's natural file sizing suggestion maybe you can use humanize: https://python-humanize.readthedocs.io/en/latest/#file-size-humanization

Example below. I think it is just pure Python dependency, so maybe reasonable enough.

image

```diff
diff --git a/darshan-util/pydarshan/darshan/cli/summary.py b/darshan-util/pydarshan/darshan/cli/summary.py
index 01b69923..a66c44ce 100644
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -12,6 +12,7 @@ else:
 
 from typing import Any, Union, Callable
 
+import humanize
 import pandas as pd
 from mako.template import Template
 
@@ -255,17 +256,17 @@ class ReportData:
             #specifying behaviors for each module
             if mod in {"POSIX", "MPI-IO", "STDIO"}:
                 if mod == "MPI-IO":
-                    read = df["MPIIO_BYTES_READ"].sum()
-                    write = df["MPIIO_BYTES_WRITTEN"].sum()
+                    read = humanize.naturalsize(df["MPIIO_BYTES_READ"].sum())
+                    write = humanize.naturalsize(df["MPIIO_BYTES_WRITTEN"].sum())
                 else:
-                    read = df[f"{mod}_BYTES_READ"].sum()
-                    write = df[f"{mod}_BYTES_WRITTEN"].sum()
+                    read = humanize.naturalsize(df[f"{mod}_BYTES_READ"].sum())
+                    write = humanize.naturalsize(df[f"{mod}_BYTES_WRITTEN"].sum())
                 
                 if mod == "H5D":
-                    val = f"{df['id'].nunique()} unique datasets ({read} total bytes read, {write} total bytes written)"
+                    val = f"{df['id'].nunique()} unique datasets ({read} read, {write} written)"
                     flag = ""
                 else:
-                    val = f"{df['id'].nunique()} unique file ({read} total bytes read, {write} total bytes written)"
+                    val = f"{df['id'].nunique()} unique file ({read} read, {write} written)"
                     flag = ""
             elif mod in {"H5F", "PNETCDF", "LUSTRE"}:
                 val = f"{df['id'].nunique()} unique file"
```

The team might prefer to use the `binary=True` option to get the powers-of-two values as well.

darshan-util/pydarshan/darshan/cli/summary.py
shanedsnyder: I'm not sure it's super helpful to display much information on the DXT modules, beyond what the version is and whether the module has partial data. Unless there's a cheap way to calculate the total number of DXT file records?

If it's too expensive to calculate that, I'd say let's just avoid it altogether. The DXT modules are ultimately controlled by the POSIX and MPI-IO modules at runtime (i.e., DXT will likely have data on the same exact files as POSIX and MPI-IO modules), so there's not really a need for lots of DXT details since you can just infer that info from POSIX/MPI-IO modules.

tylerjereddy:
For DXT you could just have a separate code path that does what we currently do I suppose, depending on the consequences for table formatting. 

I thought the length of the DXT event lists was effectively the number of recorded segments, so maybe not too hard to calculate, but would need to double check on if there is nesting in that data structure. Jakob probably remembers more accurately, but it isn't hard to check with some of the DXT-containing examples in logs repo.

shanedsnyder:
> For DXT you could just have a separate code path that does what we currently do I suppose, depending on the consequences for table formatting.
> 

This would be my current vote. We will need special handling for more modules beyond DXT, so not unreasonable. E.g., APMPI and APXC modules use at most one record per process, so we probably don't output any additional details (beyond version and partial_flag) for either of those.

> I thought the length of the DXT event lists was effectively the number of recorded segments, so maybe not too hard to calculate, but would need to double check on if there is nesting in that data structure. Jakob probably remembers more accurately, but it isn't hard to check with some of the DXT-containing examples in logs repo.

If we did decide to include some details on DXT, I think it would just be total number of unique DXT file records, not the total number of trace segments. The file records have a hierarchical structure (the file record embeds an array of trace segments) as you point out, but total number of files traced is probably more useful info than total number of operations traced.

shanedsnyder:
I already shared this with @saudahmed202, but here's what I had in mind for our existing modules:

- NULL
  - skip completely, no log will ever have this module
- POSIX, MPI-IO, STDIO
  - version, unique files, bytes read/written, partial flag
- H5F, PNETCDF, LUSTRE
  - version, unique files, partial flag
- H5D
  - version, unique datasets, bytes read/written, partial flag
- DXT_POSIX, DXT_MPIIO, BG/Q, MDHIM, APXC, APMPI, HEATMAP
  - version, partial flag

tylerjereddy:
And would you want to keep it a single table and have sort of blank/greyed-out entries for the modules with no info for some columns?

shanedsnyder:
> And would you want to keep it a single table and have sort of blank/greyed-out entries for the modules with no info for some columns?

I had originally told Saud to just combine into the existing log info table, but maybe it's just cleaner to add a new "Darshan Module Info" table that only includes this data. I think that would look cleaner now that I think about it.

I think this is just a 2 column table like this:
|   |   |
| --- | ----------- |
| **POSIX module** (ver=7)    | 10 unique files accessed (10 MB read, 8 MB written)     |
| **H5F module** (ver=3) | 1 unique file accessed | 
| **H5D module** (ver=3)   | 8 unique datasets accessed (8 MB read, 0 MB written)        |

(Still needs to have the logic to handle the partial flag, obviously)

There are sufficient differences between modules that just having an arbitrary text block describing the module is probably cleaner than trying to break down into different columns.

saudahmed202:
Screenshot (97)

This the what the tables look like with the partial flag.

Issue #781, opened 2022-07-26
ENH: mechanism to collapse partially-shared file records


carns:
This could be an option to  `darshan-convert`, like `--collapse-shared`.  The idea would be to take take records that refer to the same file and combine them into one, even if those files are not globally shared across all ranks.  The resulting format would be easier to analyze so that users don't have to account for duplicate entries.

Unfortunately we don't presently have a way to represent the process count for partially shared files in the on-disk log format.  There is a `rank` field for each record that is usually set to the rank number of the process that generated the record, or -1 for records that are shared across all ranks.  Ideally would would have a way to represent (for example) that 4 ranks out of 16 opened a file.  Maybe even a bitmap that could indicate precisely which 4 ranks it was.

carns:
Not sure it's a good idea, but I'll note it here for posterity just in case.  We could use negative values for rank besides -1 to encode additional information.

PR #783, opened 2022-07-26
Remove  --file-list and --file-list-detailed options from darshan-parser

### LINKED ISSUE #782: MAINT: deprecate --file-list and --file-list-detailed options in darshan-parser
These two options are not frequently used; they produce a text formatted table with a row for each unique file and columns for some select metrics of interest.  The maintenance problem is that those options significantly increase the complexity of the `darshan-parser` utility, particularly in terms of what capabilities could be refactored in #677 .

In the long run it would be better to handle this as described in #781 ; you would first convert a log to collapse partially shared files and then use Python scripts or other conventional analysis tools on the resulting log.

carns:
See discussion in #677 and possible long-term solution in #781 .  The --file-list and --file-list-detailed options are not frequently used (especially given other options available for analysis now), but their presence was making it difficult to refactor parser to use accumulator API because the --file-list features requires the parser to construct it's own explicit hash table of file records.

It would make more sense to maintain a more generalized version of this functionality elsewhere and reduce complexity in darshan-parser for now.

Fixes #782 .

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,10 @@
 Darshan Release Change Log
 --------------------------
 
+Darshan-3.4.1-pre1
+=============
+* Deprecated --file-list and --file-list-detailed options in darshan-parser
+
 Darshan-3.4.0
 =============
 * Fix segfault affecting new DARSHAN_MOD_DISABLE/ENABLE environment variables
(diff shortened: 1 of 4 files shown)


shanedsnyder (approved):

shanedsnyder:
Looks great, thanks Phil!

PR #787, opened 2022-08-11
HDF5 autoconf changes + hyperslab instrumentation fixes

shanedsnyder:
This PR contains a few changes to the runtime component of our HDF5 module:

* autoconf changes to conditionally compile in support for various functions (e.g., `H5Dflush`, `H5Sget_regular_hyperslab`) based on `AC_CHECK_FUNCS` checks rather than checking the HDF5 version number
* don't attribute read/write accesses to entire HDF5 datasets (i.e., 'ALL' selections) to `H5D_REGULAR_HYPERSLAB_SELECTS` counter -- there is no `ALL` variant of Darshan counters, but it can be deduced from current counters (total number of reads/writes minus total point/regular-hyperslab/irregular-hyperslab counts)
* compiler warning fixes for HDF5 versions greater than or equal to 1.8 and less than 1.10

### DIFF
--- a/darshan-runtime/configure.ac
+++ b/darshan-runtime/configure.ac
@@ -370,12 +370,23 @@ if test "x$enable_darshan_runtime" = xyes ; then
       if test $hdf5_version_major -gt 1 || (test $hdf5_version_major -eq 1 && test $hdf5_version_minor -ge 10) ; then
          AC_DEFINE([DARSHAN_HDF5_VERS_1_10_PLUS], 1,
                    [Define if HDF5 module built with version 1.10+])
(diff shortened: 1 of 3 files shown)


carns (approved):
Looks good to me.  I tested on a local build with HDF5 1.12.2.

carns:
oh actually, is there something to be worried about with the PR test failure?  I just reviewed the code and didn't look at the CI results.

shanedsnyder:
Great, thanks for the review!

The 1 CI failure is expected for now, and it should be fixed when #751 is merged. You can see that PR passes CI and is actually rebased on top of this PR, so should be good.

PR #796, opened 2022-08-20
CI: native M1 mac testing

tylerjereddy:
* add native M1 mac testing using
https://cirrus-ci.org/guide/macOS/#macos-virtual-machines
(because M1 is becoming more prominent, and the
port is easy for us it seems)

* because of the concurrency limits, only
add a single job on this platform/service
for now

* you can see the succeeding job log in a test on my fork
here: https://github.com/tylerjereddy/darshan/pull/21
if the team is "ok" with this, someone with appropriate permissions
can add the Cirrus CI app to this project (and then close/reopen
this PR I suppose)

* note that I had to use `DYLD_FALLBACK_LIBRARY_PATH` instead
of `LD_LIBRARY_PATH` for the shared `.dylib` on this platform for
whatever reason

* for some related discussion about this service see:
https://github.com/FFY00/meson-python/issues/131
https://github.com/MDAnalysis/mdanalysis/pull/3780

* the job takes 6-7 minutes at the moment, which seems fine,
even if we start to get say 3-4 minute queues if that service
gets popular (since M1 CI runners are currently rare)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,28 @@
+macos_instance:
+  image: ghcr.io/cirruslabs/macos-monterey-base:latest
+
+task:
+  script: |
+    brew install python@3.10
+    brew install python-tk@3.10
+    /opt/homebrew/opt/python@3.10/bin/python3 -m venv ~/py_310
+    source ~/py_310/bin/activate
+    python -m pip install --upgrade pip
+    python -m pip install --upgrade pytest lxml
+    python -m pip install 'matplotlib<3.5'
+    brew install automake libtool binutils
+    mkdir -p /tmp/darshan_install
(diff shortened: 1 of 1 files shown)


tylerjereddy:
I don't think it can be trigger related because the PR on my fork runs cirrus just fine with just the CI config change, right?

Assuming you have something like below in the settings, maybe what happens is that the branch was already flushed through Cirrus on my fork so it decides not to run it yet again? Maybe try merging if you agree that's a sensible hypothesis. We can always delete it if it gets annoying.

image

shanedsnyder:
Yeah, that theory makes sense, given that your fork ran the CI no problem. I'll go ahead and merge, it all seems noncontroversial.

shanedsnyder (approved):

Issue #692, opened 2022-03-16
ENH, CI: `e3sm_io_heatmap_and_dxt.darshan` memory usage causes CI error


nawtrey:
### Background

As discussed in gh-691, `e3sm_io_heatmap_and_dxt.darshan` is one of the larger logs in the `darshan-logs` repo, and when generating a summary report it yields an error in the CI:

```
/home/runner/work/_temp/67ef905d-ab26-4af3-a5d0-5d8bf0d04f00.sh: line 7:  6525 Killed                  pytest --pyargs darshan --cov-report xml --cov=$site_packages/darshan
tests/test_summary.py ...................
Error: Process completed with exit code 137.
```

This is due to the peak memory usage exceeding the Linux runner cap of 7 GB, which Tyler highlighted pretty clearly [here](https://github.com/darshan-hpc/darshan/pull/691#issuecomment-1066184467). This (unsurprisingly) occurs during the `get_heatmap_df()` call when generating the darshan summary report, specifically for the `DXT_POSIX` module, which has over 600,000 DXT segments. 

### Solutions

The current solution is to leave the `e3sm_io_heatmap_and_dxt.darshan` xfailed in the CI, but other long-term solutions were proposed that should allow us to remove the `xfail` status for this log:

- Tyler suggested the following solution [here](https://github.com/darshan-hpc/darshan/pull/691#issuecomment-1068406464):

  >There's probably a way to use a smaller memory buffer to accumulate/bin/combine/process DXT records by rank as they get read in, as long as the time bins/bounds are established early on. pd.get_dummies also supports sparse data, though that may not be the right choice b/c it limits the valid ops that can be used on the data structures.
  >
  >Avoiding the generation of intermediate data structures with > 600,000 rows x 200 columns for an input with only 512 ranks seems desirable, at least at some point. Switching from single to multiple reduction operations (summing/combining a few DXT segments instead of all of them at once) is likely slower in Python, so a low-level + concurrent solution may be warranted.

- Phil proposed an alternate solution [here](https://github.com/darshan-hpc/darshan/pull/691#issuecomment-1069189012):

  >In the long run, for logs with DXT data, we might have to decide if the DXT data should be used automatically (because of it's higher fidelity) or if it should be an optional argument, at least when the module data is above some threshold. We could put a message in the warning box along the lines of "DXT data is present in the log but was not used to generate this summary. Rerun the report with --dxt (this may be more memory and compute intensive) to ..."

[...]

PR #802, opened 2022-09-03
WIP, ENH: DXT plot threshold

tylerjereddy:
* related to gh-729 and gh-692

* draft infrastructure for skipping the processing
of DXT data (for Python html report generation) above a certain module compressed size
threshold, in cases where runtime `HEATMAP` data is
available

* note that for the vast majority of log files that
have been provided/problematic in this regard, including
the large ones from NERSC, this is of no help, because there
is no `HEATMAP` data to fall back on

* for a case where this does help, on this branch:
`time python -m darshan summary e3sm_io_heatmap_and_dxt.darshan`
`real   0m12.415s`
vs. `main`: `real       0m47.470s`

* so, that's not a bad improvement, but there are still many
things to decide/do here:

- [ ] test the size threshold empirically with more appropriate logs
- [ ] decide if we want to use the current approach of summing
      the sizes of each DXT module together vs. having per-DXT
      module thresholds
- [ ] decide if we'd also want a way to disable DXT handling even
      if `HEATMAP` is not available (otherwise, all the sample NERSC
      logs will use > 100 GB memory and be unusable with current report
      generation machinery)
- [ ] add a warning mechanism/message somewhere on the report when the
      threshold is reached to disable DXT parsing
- [ ] add a command line argument to force an override of the disable
      (if i.e., the user is working on a high memory node and really
       wants to see DXT results)
- [ ] add regression tests for the new machinery

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -345,6 +345,18 @@ class ReportData:
         )
         modules_avail = set(self.report.modules)
         hmap_modules = ["HEATMAP", "DXT_POSIX", "DXT_MPIIO"]
+        # see gh-729 and gh-692
+        # don't ingest DXT data if it appears to be
+        # too large as a compressed entity
+        if "HEATMAP" in self.report.modules:
+            max_allowed_dxt_mib = 2
+            dxt_mod_size_mib = 0
+            for mod in self.report.modules:
+                if "DXT" in mod:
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Feedback from meeting:

- Phil is in favor of checking the threshold on a per-module basis rather than summing
- Phil is also in favor of not displaying heatmaps when the threshold is exceeded even if runtime HEATMAP is absent, but we should have an option to force generation of heatmaps at your own peril

PR #799, opened 2022-08-24
ENH: faster DXT interpolation

tylerjereddy:
* apply the faster interpolation algorithm
described at:
https://github.com/darshan-hpc/darshan/issues/798#issuecomment-1226383841

* as I note in the source, this could likely be even
faster if we used Cython or numba, but I decided against
the complexity of a new build or runtime dependency just yet..

* this also has the advantage of removing a version
switch for Python (which actually should have been
a version switch for `pandas`)

* the HTML report generation time for `e3sm_io_heatmap_and_dxt.darshan` drops from `47` to `33` seconds on my machine, while the `snyder_acme..` log timing remains unchaged (dominated by "data ingestion" there); also, our internal benchmark suite shows this branch as decisively faster than `main`, so I don't really see a drawback:

`asv continuous -e -b ".*heatmap.*" main treddy_faster_dxt_interpolate`

```
       before           after         ratio
     [3fb01f49]       [9d11b152]
     <main>           <treddy_faster_dxt_interpolate>
-     13.0±0.09ms      9.88±0.09ms     0.76  dxt_heatmap.GetHeatMapDf.time_get_heatmap_df(50, 10, 0.01)
-      13.1±0.2ms       9.83±0.1ms     0.75  dxt_heatmap.GetHeatMapDf.time_get_heatmap_df(50, 10, 0.001)
-      13.2±0.1ms      9.80±0.04ms     0.74  dxt_heatmap.GetHeatMapDf.time_get_heatmap_df(1000, 10, 0.001)
-      14.5±0.1ms      10.0±0.09ms     0.69  dxt_heatmap.GetHeatMapDf.time_get_heatmap_df(10000, 10, 0.001)
-      14.6±0.1ms      9.87±0.03ms     0.68  dxt_heatmap.GetHeatMapDf.time_get_heatmap_df(1000, 10, 0.01)
-      31.1±0.2ms      16.6±0.08ms     0.53  dxt_heatmap.GetHeatMapDf.time_get_heatmap_df(50, 50, 0.001)
-      31.6±0.4ms       16.9±0.2ms     0.53  dxt_heatmap.GetHeatMapDf.time_get_heatmap_df(1000, 50, 0.001)
-      31.4±0.3ms      16.7±0.04ms     0.53  dxt_heatmap.GetHeatMapDf.time_get_heatmap_df(50, 50, 0.01)
[...]

PR #803, opened 2022-09-06
DOC: update docs to reflect new location of prepare.sh

shanedsnyder:
darshan-runtime and darshan-util docs needed to be updated to reflect the move of `prepare.sh` script to the top-level of the repo.


tylerjereddy:
The Python CI failure may be caused by the new release of `pandas` a few days ago, perhaps: https://github.com/pandas-dev/pandas/releases/tag/v1.4.4.

It may actually go away after we merge gh-799, but we could add `scipy` as an explicit dependency until we merge that and no longer need it at runtime again.

carns (approved):

shanedsnyder:
I'll just go ahead and merge since these only touch docs, but will see if we can clear up the CI failures by merging in #799 or if more is needed.

PR #805, opened 2022-09-07
BUG: logutils up-conversin of partial_flags is broken for darshan versions prior to 3.2.0

shanedsnyder:
The logutils library is not properly up-converting the module partial flags contained in the header to account for the H5D module placed after the original HDF5 file module in version 3.2.0. This means that logs prior to 3.2.0 that are parsed by newer versions of Darshan will have an "off-by-1" bug in terms of module partial flags, with this bug affecting all modules following HDF5 in Darshan's module list.

This PR properly shifts the partial flags bitfield for all modules following the HDF5 module on pre 3.2.0 logs so the flags are properly up-converted.


carns (approved):

tylerjereddy:
The Apple M1 job ("main") we may have to ignore here b/c that service didn't seem to restart on close/reopen, but it was just the same `scipy` issue that should be resolved by gh-799

shanedsnyder:
Good deal. I was curious about CI here since this bug has been around for a while, but looks like none of our tests need to be updated for it.

PR #810, opened 2022-09-13
MAINT: unpin matplotlib

### LINKED ISSUE #479: MAINT: `matplotlib==3.5b1` causes DXT heatmap code failures
The failures are caused by `python setup.py install` pulling in a bleeding-edge version of `matplotlib` (`3.5b1`). Longer-term we should probably modernize the build system per gh-476, but that's a separate matter (to `pip install .` with the `autoperf` stuff enabled, and how much "Python stuff" should still be done by the automake machinery).

I'll place the failures below the fold. Questions to answer might be:

1) Is this caused by `seaborn` not yet having shims for this bleeding-edge version of `matplotlib`?
2) Is this something we need to/should handle ourselves because `seaborn` gives us different handles with different versions of `matplotlib`?
(that kind of thing..)

For now, I will pin the `matplotlib` version we use in CI, since it is still effectively labelled as a beta for example.

tylerjereddy:
* Fixes #479

* `test_adjust_for_colorbar()` has been failing with `matplotlib >= 3.5.0` for almost a year now, which is pretty annoying because building old versions of `matplotlib` with more recent Python versions can be painful (there is no binary sometimes)

* since I don't see any actual problems with the heatmap/colorbar layouts with the newer matplotlib, add a
[PEP440](https://peps.python.org/pep-0440/)-compliant version conditional inside `test_adjust_for_colorbar()` that uses a different colorbar coordinate value with newer versions of `matplotlib`; I think it is encouraging that the new values are offset by the same value (`0.3`) as the old ones; I don't think it is worth trying to figure out which `matplotlib` commit changed the coord layout values, but we could probably do that if we really wanted..

* so, unpin `matplotlib` version in a few CI/build system locations, and add `packaging` for the PEP-compliant version comparison at test time

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -8,8 +8,7 @@ task:
     /opt/homebrew/opt/python@3.10/bin/python3 -m venv ~/py_310
     source ~/py_310/bin/activate
     python -m pip install --upgrade pip
-    python -m pip install --upgrade pytest lxml
-    python -m pip install 'matplotlib<3.5'
+    python -m pip install --upgrade pytest lxml matplotlib packaging
     brew install automake libtool binutils
     mkdir -p /tmp/darshan_install
     export DARSHAN_INSTALL_PATH=/tmp/darshan_install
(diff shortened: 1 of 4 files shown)


shanedsnyder (approved):

shanedsnyder:
Looks good, thanks!

Issue #804, opened 2022-09-06
MAINT, ENH: data access by category section requests


tylerjereddy:
@kevin-harms shared some feedback about the "data access by category" section in the Python reports, from an end user:

> In the last graphs, I would put the # files read/written in normal integer (e.g. 1, 3, 10, 200...) format instead of X.XXE+xx format. It would be easier to read and that is a number that should always be a "relatively low" number (and if not there is a problem ...)
If possible, make the font bigger.

Shane seemed to agree with the feedback--if the integer number becomes huge we just let it become visually messy I guess.

tylerjereddy:
There was also some interest in removing the scientific notation for the data volume and using i.e., MiB/GiB units to "condense" the byte counts instead. However, I'm not clear on how well that would work with a log-scaled x-axis for data volume, or at least how the log scale would work in a way that may not confuse some. You could try adding "log of bytes volume" to the x-axis but use the closest data unit `MiB`/`GiB` on the respective rows maybe..

PR #812, opened 2022-09-14
MAINT: integer r/w vals for data access

tylerjereddy:
* deals with one of the user requests in gh-804--display "Data Access by Category" # files read/written using plain integers instead of scientific notation

* it may also seem weird to prefix a `0` value and append all other integer values, but there
was no specific feedback on this aspect of the display so I've left it untouched

Before/after samples for `ior_hdf5_example.darshan` and `treddy_runtime_heatmap_inactive_ranks.darshan` are below.

image
image
image
image


shanedsnyder:
This was cherrypicked and eventually merged via #911

Issue #811, opened 2022-09-14
Memory leak in PyDarshan


jesse-r-s-hines:
There appears to be a memory leak in the PyDarshan package, causing the Python process to increase memory usage until it is killed, usually in under a minute.

## Steps to reproduce
The leak occurs when either trying to parse a very large `.darshan` file (≈1G), or simply opening and closing many smaller `.darshan` files in sequence.

```python
import darshan
import darshan.backend.cffi_backend as backend

def low_level(report_file):
    log = backend.log_open(report_file)
    rec = backend.log_get_generic_record(log, 'POSIX')
    while rec != None:
        rec = backend.log_get_generic_record(log, 'POSIX')
    backend.log_close(log)

def high_level(report_file):
    with darshan.DarshanReport(report_file, read_all=False, lookup_name_records=False) as report:
        for row in report.mod_records('POSIX'):
            pass

# These loops will leak memory.
# If the darshan files are large it take much fewer iterations to run out of RAM
for i in range(1_000_000):
    low_level('darshan_files/sample.darshan')
for i in range(1_000_000):
    high_level('darshan_files/sample.darshan')

# Actually, just opening the file will leak some as well.
for i in range(1_000_000):
    with darshan.DarshanReport('darshan_files/sample.darshan',
        read_all=False, lookup_name_records=False
    ) as report:
        pass

# Breaking down the backend method calls in `DarshanReport.open`

# Leaks
for i in range(1_000_000):
    log = backend.log_open('darshan_files/sample.darshan')
    backend.log_get_mounts(log)
    backend.log_close(log)

# Very slow leak here
for i in range(1_000_000):
    log = backend.log_open('darshan_files/sample.darshan')
    backend.log_get_modules(log)
    backend.log_close(log)

# These don't leak
for i in range(1_000_000):
    log = backend.log_open(report_file)
    backend.log_get_job(log)
    backend.log_get_exe(log)
    backend.log_close(log)
```

I don't see the issue when calling `darshan-parser` CLI, even on very large `.darshan` files `darshan-parser` will stay under 1-2G of memory usage.

I'm running on an Ubuntu 20.04 machine with 16G of RAM, with Python3.9.14 and PyDarshan 3.4.0.1. I've also reproduced the issue on PyDarshan 3.3.1.1.

The issue appears to occur with any `.darshan` file, but I've still uploaded a small [sample darshan log](https://github.com/darshan-hpc/darshan/files/9568265/sample.darshan.zip).

[...]

PR #813, opened 2022-09-15
BUG: fix mem leak across C/CFFI boundary

tylerjereddy:
* a few different memory leaks are described in gh-811, and this branch aims to deal with one
surrounding `darshan_log_get_modules()`, where
an internal `malloc` is effectively making a memory decision that is invisible to CFFI, so we can't free it externally--for example, CFFI would end up trying to free a subset of the original memory on the heap, which is prohibited

* this branch refactors `darshan_log_get_modules()` so that an external caller of this function, for example our CFFI use case, can allocate the memory needed for the module information on its own, which means that it has ownership for freeing that memory (which now happens automatically when `mods` drops out of scope in CPython)

* this substantially reduces the memory leak described in the profiling code (https://github.com/darshan-hpc/darshan/issues/811#issuecomment-1248290837), but not completely, since there appear to be other leaks related to opening log files, etc., that I'd prefer to investigate in separate PRs

Here is the somewhat less serious memleak with the same benchmark on this branch (1/2 the total footprint):

image

### DIFF
--- a/darshan-util/darshan-logutils.c
+++ b/darshan-util/darshan-logutils.c
@@ -2069,7 +2069,7 @@ static int darshan_log_get_namerecs_3_00(void *name_rec_buf, int buf_len,
  * Gets list of modules present in logs and returns the info
  */
 void darshan_log_get_modules(darshan_fd fd,
-                             struct darshan_mod_info **mods,
+                             struct darshan_mod_info mods[DARSHAN_MAX_MODS],
                              int* count)
 {
     int i;
@@ -2082,18 +2082,15 @@ void darshan_log_get_modules(darshan_fd fd,
         return;
     }
 
-    *mods = malloc(sizeof(**mods) * DARSHAN_MAX_MODS);
-    assert(*mods);
-
     for (i = 0, j = 0; i < DARSHAN_KNOWN_MODULE_COUNT; i++)
     {
         if (fd->mod_map[i].len)
         {
-            (*mods)[j].name          = darshan_module_names[i];
-            (*mods)[j].len           = fd->mod_map[i].len;
-            (*mods)[j].ver           = fd->mod_ver[i];
-            (*mods)[j].partial_flag  = DARSHAN_MOD_FLAG_ISSET(fd->partial_flag, i);
-            (*mods)[j].idx           = i;
+            (mods)[j].name          = darshan_module_names[i];
+            (mods)[j].len           = fd->mod_map[i].len;
+            (mods)[j].ver           = fd->mod_ver[i];
+            (mods)[j].partial_flag  = DARSHAN_MOD_FLAG_ISSET(fd->partial_flag, i);
+            (mods)[j].idx           = i;
             j += 1;
         }
     }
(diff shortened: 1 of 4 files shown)


tylerjereddy:
I should also add that I'd like to add a regression guard here, probably in the form of an `asv` benchmark suite peak memory usage benchmark, though we could open an issue to remind us to add that if we don't want to delay patching the leak for too long.

darshan-util/darshan-logutils.c
```diff
@@ -2069,7 +2069,7 @@ static int darshan_log_get_namerecs_3_00(void *name_rec_buf, int buf_len,
  * Gets list of modules present in logs and returns the info
  */
 void darshan_log_get_modules(darshan_fd fd,
-                             struct darshan_mod_info **mods,
+                             struct darshan_mod_info mods[DARSHAN_MAX_MODS],
```
carns, line 2072: I'm not sure what putting a static sized array in the definition does here actually?  C doesn't know how to do bounds checks.  It might be more conventional to just match the logutils.h function prototype (with `struct darshan_mod_info *mods`).

More generally, this API doesn't do anything that obligates the caller to pass in a sufficiently large array.  We could make the `count` argument an input/output argument as one option (on the way in it is the size of the array, on the way out it is how many array elements were populated).  Or make another function to report how many modules are present, since it won't change once the log exists.

carns (changes requested):
I like the spirit of the change, just not sure about how the C compiler will interpret the array bounds on the definition,  and technically we probably need a mechanism (even if it is just the count argument) for caller and function to negotiate array size.

darshan-util/darshan-logutils.c
carns: Actually I retract the last sentence above.  Even if there were a separate helper fn to inform the caller of the module count, this function still can't know if the caller made the array big enough :)  Maybe just make the count an in/out and call it a day.

tylerjereddy: It just decays to a pointer yeah, although in 2 dimensions I believe the compiler will actually barf on misuse. Can't you use an assert to enforce the static size if you don't trust the caller? It should always be `DARSHAN_MAX_MODS` rather than negotiable, no?

carns: By convention the caller should make it DARSHAN_MAX_MODS for sure.  I don't think there is any technical way to enforce or assert that it's that big, though.  I could be proven wrong :)

I think the more C-like strategy is for the caller to tell the function how big the buffer is rather than making its size implied.  The former is a little more explicit/pedantic from an API perspective and makes sure that (for example) you don't somehow accidentally get a mismatch on what the value of DARSHAN_MAX_MODS is between the function and the caller.

tylerjereddy: arrghh, `sizeof` is going to resolve to the size of a single struct b/c you feed in a pointer... fat pointers like Rust/C++ would be nice. You could make your own "smart type" (augment the `struct` I suppose) by using a termination sentinel and iterating over the input array I guess. C isn't memory safe, so I'm pretty sure a sufficiently creative person can cause trouble with pointers into the abyss anyway.

Or maybe even mark/name a new function for CFFI usage only. Either approach seems better than the leak in production. Feel free to push in whatever changes you like.

tylerjereddy:
superseded by gh-816

PR #815, opened 2022-09-15
BUG: log_get_mounts mem leak

tylerjereddy:
* fix a memory leak at the CFFI interface to `darshan_log_get_mounts()` by allowing CFFI
to first count the number of moint points,
then allocate its own struct of the appropriate
size, and then fill that struct with the mount
points

* solution here is a bit more complex b/c the C code is used elsewhere--I didn't check
that in great detail and instead skirted the issue by writing new C functions I actually
wanted
 
* this is done by adding two new functions that split the logic into bits that allow CFFI to
properly drop unused memory promptly, which
reduces our memory footprint by 3.5 GiB vs.
`main` and gh-813 for the `2.3 KiB` reproducer
log (see comparison memtrace plot at:
https://github.com/darshan-hpc/darshan/issues/811#issuecomment-1248514526)

New trace is nice and stable for that reproducer:
image

* we may want to add `asv` peak memory usage benchmarks, or open an issue to remind us to do that for regression guard of sorts..

### DIFF
--- a/darshan-util/darshan-logutils.c
+++ b/darshan-util/darshan-logutils.c
@@ -2207,6 +2207,104 @@ int darshan_log_get_record(darshan_fd fd,
     return r;
 }
 
+/* count the mount points for i.e., usage with
+ * darshan_log_get_mounts_cffi
+ */
+
+int darshan_log_get_mount_count(darshan_fd fd, int* count)
+{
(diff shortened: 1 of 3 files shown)


tylerjereddy:
superseded by gh-818

PR #816, opened 2022-09-22
BUG: memleak fix CFFI only pt1

tylerjereddy:
* this achieves identical memory leak improvements as gh-813 without touching any of our existing util C API functions (300 MB -> 150 MB footprint for the same benchmark test as there)

### DIFF
--- a/darshan-util/darshan-logutils.c
+++ b/darshan-util/darshan-logutils.c
@@ -2207,6 +2207,11 @@ int darshan_log_get_record(darshan_fd fd,
     return r;
 }
 
+void darshan_free(void *ptr)
+{
+    free(ptr);
+}
+
 /*
  * Local variables:
  *  c-indent-level: 4
--- a/darshan-util/darshan-logutils.c
+++ b/darshan-util/darshan-logutils.c
@@ -2207,6 +2207,13 @@ int darshan_log_get_record(darshan_fd fd,
     return r;
 }
 
+/*
+ * darshan_free
+ *
+ * We expose free() in this manner so that i.e.,
+ * CFFI can free memory that has been malloc'd
+ * inside the scope of C functions.
+ */
 void darshan_free(void *ptr)
 {
     free(ptr);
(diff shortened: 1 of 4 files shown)


darshan-util/darshan-logutils.c
```diff
@@ -2207,6 +2207,11 @@ int darshan_log_get_record(darshan_fd fd,
     return r;
 }
 
+void darshan_free(void *ptr)
```
carns, line 2210: maybe a comment here (or in the header, whatever is most like the rest of the code) to indicate that this mostly exists as a helper routine for the python bindings in case anyone sees it in the C API and wonders what it does.

tylerjereddy: I added a short explanation above the function, following the convention in this file. The header file doesn't seem to have comments for the functions in this source file.

carns: Great, thanks!

tylerjereddy:
I also checked that this doesn't work as an alternative to manual `free()`:

```diff
--- a/darshan-util/pydarshan/darshan/backend/cffi_backend.py
+++ b/darshan-util/pydarshan/darshan/backend/cffi_backend.py
@@ -201,7 +201,8 @@ def log_get_modules(log):
 
     modules = {}
 
-    mods = ffi.new("struct darshan_mod_info **")
+    mods_ptr = ffi.new("struct darshan_mod_info *")
+    mods = ffi.new("struct darshan_mod_info **", mods_ptr)
     cnt    = ffi.new("int *")
     libdutil.darshan_log_get_modules(log['handle'], mods, cnt)
     for i in range(0, cnt[0]):

```

i.e., it doesn't appear to simply be an issue that CFFI can't deal with the pointer dereferencing when dropping memory. So, I'm not aware of any simpler solution than re-exporting `free()`.

shanedsnyder (approved):

tylerjereddy:
Alright, this has an approval and Phil was "ok" with it on Slack as well

Issue #817, opened 2022-09-22
TST, BENCH: memory profiling/testing


tylerjereddy:
This is just a reminder to check if our current `asv` suite is sufficient to detect the memory footprint improvements related to fixing gh-811, and if not we should add a peak memory benchmark or two that are sensitive to that suite.

That said, it is quite common *not* to have CI tests for memory leaks, even in quite large projects, so we can probably delay the actual mem profiling code for a little bit as long as we fix soon-ish.

PR #819, opened 2022-09-22
BUG: memleak part 3 (CFFI only)

tylerjereddy:
* this should be reviewed after gh-816, and fixes
    the residual memory leak there so that we get
    a flat memory profile with repeated calls
    to `log_get_generic_record()`; after rebasing, this PR
   should simplify to a few CFFI backend changes to free
   `buf[0]` in a few places and make explicit copies for NumPy
    where it is pointing to the dropped memory buffer
    
 * note that `RUNTIME` heatmap doesn't seem to like
    the extra `free()`, it causes some testing issues,
    so I've excluded that here--hopefully that module
    just doesn't have the same problem..

Gradual mem spike -> flat profile:

image

### DIFF
--- a/darshan-util/pydarshan/darshan/backend/cffi_backend.py
+++ b/darshan-util/pydarshan/darshan/backend/cffi_backend.py
@@ -638,10 +638,11 @@ def _log_get_heatmap_record(log):
     # write/read bins
     sizeof_64 = ffi.sizeof("int64_t")
     
-    write_bins = np.frombuffer(ffi.buffer(filerec[0].write_bins, sizeof_64*nbins), dtype = np.int64)
+    write_bins = np.copy(np.frombuffer(ffi.buffer(filerec[0].write_bins, sizeof_64*nbins), dtype = np.int64))
     rec['write_bins'] = write_bins
 
-    read_bins = np.frombuffer(ffi.buffer(filerec[0].read_bins, sizeof_64*nbins), dtype = np.int64)
+    read_bins = np.copy(np.frombuffer(ffi.buffer(filerec[0].read_bins, sizeof_64*nbins), dtype = np.int64))
     rec['read_bins'] = read_bins
+    libdutil.darshan_free(buf[0])
     
     return rec
--- a/darshan-util/pydarshan/darshan/backend/cffi_backend.py
+++ b/darshan-util/pydarshan/darshan/backend/cffi_backend.py
@@ -341,7 +341,6 @@ def log_get_generic_record(log, mod_name, dtype='numpy'):
     buf = ffi.new("void **")
     r = libdutil.darshan_log_get_record(log['handle'], modules[mod_name]['idx'], buf)
     if r < 1:
-        libdutil.darshan_free(buf[0])
         return None
     rbuf = ffi.cast(mod_type, buf)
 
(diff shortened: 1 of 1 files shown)


darshan-util/pydarshan/darshan/backend/cffi_backend.py
```diff
@@ -340,6 +341,7 @@ def log_get_generic_record(log, mod_name, dtype='numpy'):
     buf = ffi.new("void **")
     r = libdutil.darshan_log_get_record(log['handle'], modules[mod_name]['idx'], buf)
     if r < 1:
+        libdutil.darshan_free(buf[0])
```
shanedsnyder, line 344: Probably not super obvious from the code, but `darshan_log_get_record` will not return with any data malloced if `r < 1` (i.e., there's no more data or an error was encountered). I think we need to drop this `free()`. I started to wonder why you didn't see double free errors for this, but realized on this code path the output pointer is set to `NULL`, so you are just freeing a `NULL` pointer which isn't technically illegal.

darshan-util/pydarshan/darshan/backend/cffi_backend.py
```diff
@@ -460,6 +463,7 @@ def _log_get_lustre_record(log, dtype='numpy'):
     buf = ffi.new("void **")
     r = libdutil.darshan_log_get_record(log['handle'], modules['LUSTRE']['idx'], buf)
     if r < 1:
+        libdutil.darshan_free(buf[0])
```
shanedsnyder, line 466: Drop this `free()` as well.

darshan-util/pydarshan/darshan/backend/cffi_backend.py
```diff
@@ -545,6 +550,7 @@ def log_get_dxt_record(log, mod_name, reads=True, writes=True, dtype='dict'):
     buf = ffi.new("void **")
     r = libdutil.darshan_log_get_record(log['handle'], modules[mod_name]['idx'], buf)
     if r < 1:
+        libdutil.darshan_free(buf[0])
```
shanedsnyder, line 553: Drop this `free()` as well.

shanedsnyder (changes requested):

shanedsnyder:
>note that RUNTIME heatmap doesn't seem to like
the extra free(), it causes some testing issues,
so I've excluded that here--hopefully that module
just doesn't have the same problem..

I'll see if I can figure out what's going on here, that seems like it could be a problem

tylerjereddy:
Ok, I think I addressed the comments so far. I also rebased to vastly simplify the displayed diff.

shanedsnyder:
> > note that RUNTIME heatmap doesn't seem to like
> > the extra free(), it causes some testing issues,
> > so I've excluded that here--hopefully that module
> > just doesn't have the same problem..
> 
> I'll see if I can figure out what's going on here, that seems like it could be a problem

I think I fixed the memory leak and the test errors via [b21d644](https://github.com/darshan-hpc/darshan/pull/819/commits/b21d644024ed813c62b6130df692ad2c4d9f04cf), which looked like they were ultimately caused by incorrect data being stored in heatmap records when we add in the necessary `free()` statement to avoid mem leaks. Looks like we just needed a `np.copy()` as in other places (e.g., DXT segment extraction) so record data isn't freed out from under heatmap records.

@tylerjereddy if it looks good to you, I think we can go ahead and merge. I'm not sure if we had a good reproducer for the heatmap module-specific memory leak, but might be good to confirm this is working as expected in terms of leaking if we do.

tylerjereddy:
@shanedsnyder Probably not HEATMAP reproducer yet, but may we can include that separately as part of https://github.com/darshan-hpc/darshan/issues/817. I'd probably just lift the reproducers from gh-811, and supplement for any modules not covered there, if needed. I guess that means putting stuff in the `asv` suite, which takes a bit more testing.

Once this is merged (it looks sensible to me as well) maybe we should ping in gh-811 for user confirmation of the fix if they are comfortable recompiling from `main`.

shanedsnyder (approved):

PR #823, opened 2022-09-28
TST: Python 3.11

### LINKED ISSUE #795: CI: more inclusive pydarshan paths
The current `pydarshan` CI testing has `paths` triggers that are too restrictive (see below). Case in point, if you change the CI configuration in the `.github/workflows/main_ci.yml` file, it should rerun the CI steps specified in that `yml` file, otherwise there is no sane way for a reviewer to consider the correctness of the CI configuration changes in practice.

```yml
name: Python Testing

on:
  push:
    branches:
      - main
    paths:
      - darshan-util/**
      - include/**
  pull_request:
    branches:
      - main
    paths:
      - darshan-util/**
      - include/**
```

tylerjereddy:
Fixes #795 as well

* Python `3.11.0rc2` is not only API stable with the final release coming in a few weeks, but is also guaranteed ABI stable: https://www.python.org/downloads/release/python-3110rc2/

* let's start testing it in our CI--many upstream projects are already providing pre-built `3.11` binaries on PyPI (`pandas`, `numpy`, `matplotlib`..)

* the run time for this new job using `act` locally seems to be about 10 minutes, which is reasonably in line with the other Python version jobs I'd say

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -13,6 +13,7 @@ on:
     paths:
       - darshan-util/**
       - include/**
+      - .github/workflows/**
 
 jobs:
   test_pydarshan:
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         platform: [ubuntu-latest,
                    macos-latest]
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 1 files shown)


tylerjereddy:
The pydarshan suite isn't running in CI b/c of one of my previous complaints: https://github.com/darshan-hpc/darshan/issues/795 maybe I'll just fix that while I'm at it here...

shanedsnyder (approved):

PR #825, opened 2022-10-03
BUG: memleak part 4

### LINKED ISSUE #824: Another memory leak in PyDarshan
I've come across another memory leak while working with PyDarshan. I'm running the current `main` branch, compiled locally on Ubuntu 20.04, and using Python3.9.14

Running this code will slowly leak memory:
```python
def leak(report_file):
    with darshan.DarshanReport(report_file, read_all=True, lookup_name_records=True) as report:
        pass

for i in range(1_000_000):
    leak("sample.darshan")
```

Running this code without `read_all=True` doesn't leak anymore (as it was fixed in #811):
```python
def no_leak(report_file):
    with darshan.DarshanReport(report_file, read_all=False, lookup_name_records=True) as report:
        for row in report.mod_records('POSIX'):
            pass
```

The issue appears to occur with any darshan file, but here's a [sample](https://github.com/darshan-hpc/darshan/files/9568265/sample.darshan.zip) that is showing the leak.

shanedsnyder:
* `dup()` returned record name and destroy the name record hash table and all associated memory in functions `darshan_log_get_name_records()` and `darshan_log_get_filtered_name_records()`
* free allocated array of record_id,record_name tuples passed back from above functions in CFFI backend
* free individual record names passed back frome above functions in CFFI backend

Fixes #824

### DIFF
--- a/darshan-util/darshan-logutils.c
+++ b/darshan-util/darshan-logutils.c
@@ -2143,7 +2143,10 @@ void darshan_log_get_name_records(darshan_fd fd,
     HASH_ITER(hlink, name_hash, curr, tmp)
     {
         (*name_records)[i].id = curr->name_record->id;
-        (*name_records)[i].name = curr->name_record->name;
+        (*name_records)[i].name = strdup(curr->name_record->name);
+        HASH_DELETE(hlink, name_hash, curr);
+        free(curr->name_record);
+        free(curr);
         i++;
     }
  
@@ -2183,7 +2186,10 @@ void darshan_log_get_filtered_name_records(darshan_fd fd,
     HASH_ITER(hlink, name_hash, curr, tmp)
     {
         (*name_records)[i].id = curr->name_record->id;
-        (*name_records)[i].name = curr->name_record->name;
+        (*name_records)[i].name = strdup(curr->name_record->name);
+        HASH_DELETE(hlink, name_hash, curr);
+        free(curr->name_record);
+        free(curr);
         i++;
     }
  
--- a/darshan-util/darshan-logutils.c
+++ b/darshan-util/darshan-logutils.c
@@ -2143,6 +2143,16 @@ void darshan_log_get_name_records(darshan_fd fd,
     HASH_ITER(hlink, name_hash, curr, tmp)
     {
         (*name_records)[i].id = curr->name_record->id;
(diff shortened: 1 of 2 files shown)


shanedsnyder:
Memory profile looks much flatter now for reproducer in #824.
image

Although, you do see some slight increase over time, looks like basically a couple of MiB.

Extending the iterations:
image

Not sure what's going on as it seems to periodically reset, then rise again. Is it some sort of artifact of garbage collection happening in CFFI or something like that?

tylerjereddy (commented):
The Python side looks good to me at least. I also confirmed a fix to the memory leak with `ior_hdf5_example.darshan` and the reproducer in the matching issue, after accounting for likely garbage collection oscillations:

image

darshan-util/darshan-logutils.c
```diff
@@ -2143,7 +2143,10 @@ void darshan_log_get_name_records(darshan_fd fd,
     HASH_ITER(hlink, name_hash, curr, tmp)
     {
         (*name_records)[i].id = curr->name_record->id;
-        (*name_records)[i].name = curr->name_record->name;
```
carns, line 2146: Good fix.  My only suggestion would be to put a comment in here explaining that the hash table was previously constructed on the fly by darshan_log_get_namehash(), so this function is responsible for releasing the hash table memory before it returns.  It doesn't look symmetric at first glance.

carns (changes requested):
Looks good, I just have a minor suggestion to add comments to help remember what's going on here in the future.  I think there are two instances to put it on.  Otherwise looks good to me.

darshan-util/darshan-logutils.c
shanedsnyder: Yeah, that's a good point. These functions might seem strange to someone at first glance, but really these are just helpers for PyDarshan that translate the name record hash Darshan would normally provide into something that can be managed by CFFI (an array of record ID, name tuples).

shanedsnyder: This should be done in [8f7d8a3](https://github.com/darshan-hpc/darshan/pull/825/commits/8f7d8a318d7041469177fa2ef5b504c1b4e5da8b).

carns: perfect, thanks

carns (approved):
lgtm

PR #827, opened 2022-10-04
BUG: Fix handling of forked processes in darshan-runtime

shanedsnyder:
Darshan's ability to instrument forked processes was broken by some optimizations to the module registration code path that were  applied as part of #703.

This PR includes fixes to each module that resets the module's `runtime_init_attempted` flag at module `cleanup()` time. If this flag isn't reset, then modules can not be re-initialized after they have shut down -- Darshan's fork handling relies on this ability to cleanup and re-initialize modules and was leading to fork children not being able to register records in the Darshan log.

Additionally, this PR includes a regression test to detect whether this functionality is broken again going forward.

[WIP tag while I fine tune the regression test]

### DIFF
--- a/darshan-test/python_runtime_tests.py
+++ b/darshan-test/python_runtime_tests.py
@@ -197,6 +197,8 @@ def test_forked_process(tmpdir):
                      "-x",
                      f"LD_PRELOAD={darshan_lib_path}",
                      "-x",
+                     "DARSHAN_ENABLE_NONMPI=1",
+                     "-x",
                      f"DARSHAN_LOGPATH={cwd}",
                      "python",
                      f"{test_script_path}"])
--- a/darshan-test/python_runtime_tests.py
+++ b/darshan-test/python_runtime_tests.py
@@ -177,7 +177,6 @@ def test_env_vars_config(tmpdir):
 
 def test_forked_process(tmpdir):
     # regression test for gh-786
-    n_ranks = 1
     root_path = os.environ.get("DARSHAN_ROOT_PATH")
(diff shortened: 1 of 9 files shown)


shanedsnyder:
Removing WIP tag, this is ready for review/merge, I think. darshan-runtime library changes are pretty non-controversial, but also up for suggestions on Python regression testing I added. The testing is ensuring that expected file records generated on parent and child processes are found -- I don't think we need to check counter values, so much as just confirm expected records were generated.

darshan-test/python_runtime_tests.py
```diff
@@ -175,3 +174,50 @@ def test_env_vars_config(tmpdir):
...
+                                    "python_mpi_scripts",
+                                    "runtime_prog_issue_786.py")
+    darshan_lib_path = os.path.join(darshan_install_path,
+                                    "lib",
+                                    "libdarshan.so")
+
+    with tmpdir.as_cwd():
+        cwd = os.getcwd()
+        subprocess.check_output([
+                     "python",
```
tylerjereddy, line 193: I needed this to be changed to `python3` to execute locally. That would be my normal expectation on Ubuntu shell call-outs with Python anyway, though CI seems happy enough.

I think rather than asking you to change that now, what I'll do is open a separate small issue suggesting we try something a bit more portable like asking Python where its executable is directly:

```python
import sys
py_exe = sys.executable
```

tylerjereddy (approved):
I checked that if checkout the testing changes on `main` the regression test does indeed fail for the expected reason--that is, both parent and child produce darshan log files, but the child log lacks the record.

tylerjereddy:
Phil already approved the C changes, so merging now.

PR #832, opened 2022-10-17
BUG: LUSTRE fcounters guard

### LINKED ISSUE #831: PyDarshan mod_records fails on LUSTRE records
`DarshanReport.mod_record` and `cffi_backend.log_get_generic_record` throw when loading a `LUSTRE` module
```python
r = darshan.DarshanReport('sample.darshan')
next(r.mod_records('LUSTRE'))
```
throws:

```
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/dist-packages/darshan-3.4.0.1-py3.9.egg/darshan/report.py", line 974, in mod_records
    rec = backend.log_get_generic_record(self.log, mod, dtype=dtype)
  File "/usr/local/lib/python3.9/dist-packages/darshan-3.4.0.1-py3.9.egg/darshan/backend/cffi_backend.py", line 354, in log_get_generic_record
    flst = np.copy(np.frombuffer(ffi.buffer(rbuf[0].fcounters), dtype=np.float64))
AttributeError: cdata 'struct darshan_lustre_record *' has no field 'fcounters'
```

I'm running PyDarshan 3.4.0.1

I've attached a [sample darshan file](https://github.com/darshan-hpc/darshan/files/9780804/sample.darshan.zip) that shows the issue.

tylerjereddy:
Fixes #831

* prevent `log_get_generic_record()` from attempting to access the `LUSTRE` `fcounters` field, which
apparently does not exist

### DIFF
--- a/darshan-util/pydarshan/darshan/backend/cffi_backend.py
+++ b/darshan-util/pydarshan/darshan/backend/cffi_backend.py
@@ -354,7 +354,8 @@ def log_get_generic_record(log, mod_name, dtype='numpy'):
         rec['file_rec_id'] = rbuf[0].file_rec_id
 
     clst = np.copy(np.frombuffer(ffi.buffer(rbuf[0].counters), dtype=np.int64))
-    flst = np.copy(np.frombuffer(ffi.buffer(rbuf[0].fcounters), dtype=np.float64))
+    if mod_name != "LUSTRE":
+        flst = np.copy(np.frombuffer(ffi.buffer(rbuf[0].fcounters), dtype=np.float64))
     libdutil.darshan_free(buf[0])
 
     c_cols = counter_names(mod_name)
@@ -362,7 +363,8 @@ def log_get_generic_record(log, mod_name, dtype='numpy'):
 
     if dtype == "numpy":
         rec['counters'] = clst
-        rec['fcounters'] = flst
+        if mod_name != "LUSTRE":
+            rec['fcounters'] = flst
 
     elif dtype == "dict":
         rec['counters'] = dict(zip(c_cols, clst))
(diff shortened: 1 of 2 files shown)


shanedsnyder:
Looking at the code, it looks like `log_get_generic_record()` isn't intended to be used on Lustre module data. See this line in `mod_read_all_records()`:

```
        unsupported =  ['DXT_POSIX', 'DXT_MPIIO', 'LUSTRE', 'APMPI', 'APXC', 'HEATMAP']
```

To be honest, I can't really wrap my heard around all the logic here, but it seems pretty clear that things need some work. My understanding is that `log_get_generic_record()` was intended for use with modules that have fixed-length records and that don't require module-specific logic for getting records, but there's a clear discrepancy in how/if modules are gated from calling this function in various places in the Report interface:

1. `mod_read_all_records()` returns an error if user calls with an unsupported module from above
2. `mod_records()` allows any module to be used, ultimately leading to error in the CFFI backend (as the original user reports)

This code needs more thorough auditing, but I think something like the following might make more sense:

1. `mod_read_all_records()` should read literally every record as its name implies, multiplexing to module-specific code rather than "generic" records if need be
2. `mod_records()` should similarly support any given module, with multiplexing to module-specific code if need be
3. `log_get_generic_records()` should probably enforce any unsupported modules internally, rather than relying on higher levels to reproduce this logic
4. `log_get_generic_records()` ought to support reading of APMPI and APXC data. They are listed in the "unsupported" list, but they are fixed length so this should be fine -- I wonder if this was just a hack around early issues in PyDarshan in terms of integration with APXC/APMPI.

PR #833, opened 2022-10-18
WIP: workaround for hdf undefined symbol error

carns:
For discussion.  There may be a cleaner solution, but this works around a potential undefined symbol error on preloaded executables when Darshan is compiled against HDF5 1.13.2.


shanedsnyder (approved):

shanedsnyder:
LGTM, thanks!

PR #834, opened 2022-10-19
LDMS implementation for runtime I/O collection

Snell1224:
This includes LDMS API code that collects I/O event data in real time.
This PR is a work in progress and will be updated with the necessary test script and documentation.


Snell1224:
Branch is now updated with a CI github test. This tests the Darshan-LDMS integrated code by running a simple mpi-io-test and checks that both Darshan and LDMS are collecting I/O events and completes successfully when LDMS is not functioning properly (i.e. env variables not set, cannot connect to daemon).

Right now, the CI test clones and installs Darshan from my repository (i.e. Snell1224/darshanConnector) to test the LDMS implemented code. 

Once this PR has been reviewed and/or approved to merge into the main branch, I will update the CI test accordingly (i.e. install Darshan from the main branch of this repository). If it's better to leave this PR as a separate branch and not merge the changes, that's fine too. I will also update the CI test accordingly (i.e. test will trigger and the Darshan will be installed from the darshanConnector branch of this repository).

darshan-runtime/lib/darshan.h
```diff
@@ -387,7 +387,7 @@ static inline int darshan_core_disabled_instrumentation(void)
 }
 
 /* retrieve absolute wtime */
-static inline double darshan_core_wtime_absolute(void)
+static inline double darshan_core_wtime_absolute(struct timespec *tspec)
```
shanedsnyder, line 390: Can you clarify why you needed to make the change to pass in timespec values to these functions? Not clear from reviewing just the changes made to Darshan's core, I haven't started reviewing the changes to specific instrumentation modules yet.

shanedsnyder: Looking further into the POSIX module, it looks like the timespecs are expected by the LDMS Darshan connector APIs. I feel like there might be a simpler way to modify all of the timer logic, but I'm not exactly sure what LDMS is expecting in terms of timestamps.

Snell1224: LDMS is collecting timeseries data of each individual operation for each module so it is expecting the absolute time in which they occur instead of the relative time (i.e. epoch instead of seconds). 

From what I've found in the Darshan code, it looks like all of the timer logic (i.e. the timespec struct) is within this darshan_core_wtime_absolute() function which returns a floating point value. In order to collect the absolute timestamp without disrupting Darshans timer logic, we figured the best way to do this was to pass the timespec values through the functions where LDMS is being called/used (i.e. in the macro functions). 

We can add a separate timespec struct inside one of the Darshan LDMS functions but this would affect Darshan's overhead and performance.

shanedsnyder: I see, thanks for the explanation.

I'm sure you noticed, but Darshan does capture time in terms of seconds/nanoseconds (using a `struct timespec`), it's just that it converts it to floating point time relative to the application start. when doing measurements. I'll think about it some more, but we might have a couple of options for simplifying this a bit:

1. Just have Darshan timing functions return timespecs and appropriately convert those elsewhere where we just want floating point elapsed time.
2. Darshan captures and stores a `struct timespec` for the start time of the job, of which the relative timers should be offset from. Maybe we could expose that value and let some logic handle the conversion when sending to LDMS (i.e., convert given relative timer to an absolute timer using the job start time).

shanedsnyder (changes requested):
Can we move this LDMS functionality to a new darshan-ldms.c source file, rather than in the DXT source here?

The DXT module currently just piggybacks off of our POSIX and MPI-IO modules to provide more detailed tracing, similar to what you want to with the LDMS connector. But, there isn't really any integration between LDMS/DXT that I can see here (all calls to LDMS are in the POSIX module, in the MPI-IO module, etc.), so I'm not sure this is the best spot for the code responsible for talking with LDMS. Pushing as much code to a standalone LDMS source file as possible and trying to limit LDMS-specific code in other source files will probably make maintenance a little easier on our end.

darshan-runtime/lib/darshan.h
Snell1224: I agree that there's probably a better way to sort this out, we were just trying to get LDMS implemented and working on Darshan without modifying or requesting significant changes to the original code. 

1. This seems to be the most straight forward but could be a bit time consuming to do because, from what I recall, there are multiple functions/macros in all modules that use the floating point elapsed time.
2.  This is a bit more complex but once working, it doesn't sound like a lot of code refactoring is required.

Right now, I can't really can't think of any better ideas. But I am fine with which ever option is the best for Darshan and don't mind making the necessary changes in this branch.

Snell1224:
The LDMS functionality has been moved to a new darshan-ldms.c source file. Please let me know if there are any issues with these changes.

darshan-runtime/lib/darshan.h
Snell1224: @shanedsnyder I spoke with one of our LDMS team members who's very experienced in the coding and development of LDMS and, based on their experience, they suggest only having timespec values returned from all timing functions would be best (i.e. option 1). LDMS already has this code structure and, as far as I know, it's been fairly quick and straightforward to implement any new timing functions, conversions, units, etc. (i.e. there hasn't been any need for major code changes).

I don't know if there were any new ideas about what to do with ```struct timespec``` or the future plans for Darshan but thought I'd let you know what we were thinking.

shanedsnyder:
Thanks for the updates! I'll do my best to make another pass over this this week and see if I have more comments related to the timer issue.

darshan-runtime/lib/darshan-core.c
```diff
@@ -394,12 +407,37 @@ void darshan_core_initialize(int argc, char **argv)
         darshan_core_fprintf(stderr, "darshan:init\t%d\t%f\n", nprocs, init_time);
     }
 
+
+#ifdef HAVE_LDMS
```
shanedsnyder, line 411: I think this entire block could just be shifted into `darshan_ldms_connector_initialize()`, with you passing whatever is needed as arguments? That would tidy up the integration a bit more.

darshan-runtime/lib/darshan-core.c
```diff
@@ -75,6 +76,11 @@ static int mnt_data_count = 0;
 extern void bgq_runtime_initialize();
 #endif
 
+#ifdef HAVE_LDMS
+extern struct darshanConnector dC;
+extern void darshan_ldms_connector_initialize();
+#endif
```
shanedsnyder, line 82: If we move all of the LDMS code into the initialize routine as I suggest below, we can remove this block. No need to define external variable `dC` in that case. Also, there shouldn't be any need to declare the initialize routine as an extern in any case, since it's defined as a prototype in the header included above.

darshan-runtime/lib/darshan-core.c
```diff
@@ -394,12 +407,37 @@ void darshan_core_initialize(int argc, char **argv)
...
+        dC.uid = getuid();
+        dC.exename = argv[0];
+
+        /* Pull executable name from darshans variable if no arguemments are given. */
+        if (argc==0)
+        {
+            char buff[DARSHAN_EXE_LEN];
+            int len = readlink("/proc/self/exe", buff, sizeof(buff)-1);
+            buff[len] = '\0';
+            dC.exename = buff;
```
shanedsnyder, line 430: Most of this info is already captured by Darshan and I think we could just reuse it instead of redoing the work here.

- `jobid`: you are already using the value captured by Darshan, so all good
- `uid`: you can just use the value in the Darshan job structure from above: `init_core->log_job_p->uid`
- `exename`: this one is a little trickier, but I think you can actually just use `init_core->log_exemnt_p`, which is a null-terminated string of the command line and it's arguments.

Additionally, I notice that there is already a codepath in the LDMS initialize routine that calls `gethostname()` (see `setup_connection()`, line 70). Could we consolidate down to one call there? Actually, looking closer, it looks like you use the environment variable `DARSHAN_LDMS_HOST` to override this in all cases, so is the call even needed at all?

darshan-runtime/lib/darshan-ldms.c
```diff
@@ -0,0 +1,249 @@
...
+                sem_post(&dC.recv_sem);
+                break;
+        case LDMS_XPRT_EVENT_SEND_COMPLETE:
+                break;
+        default:
+                printf("Received invalid event type %d\n", e->type);
+        }
+}
+
+#define SLURM_NOTIFY_TIMEOUT 5
```
shanedsnyder, line 60: This doesn't appear to be used, so we could probably drop? At first glance, I wondered if there was something Slurm specific about the code in this file but that doesn't appear to be the case, so could also avoid some confusion.

shanedsnyder (changes requested):

darshan-runtime/lib/darshan-ldms.c
```diff
@@ -0,0 +1,249 @@
...
+        return ldms_g;
+}
+
+void darshan_ldms_connector_initialize()
+{
+    if (!getenv("DARSHAN_LDMS_STREAM"))
+        dC.env_ldms_stream = "darshanConnector";
+
+    /* Set flags for various LDMS environment variables */
+    if (getenv("DXT_ENABLE_LDMS"))
```
shanedsnyder, line 112: I think we should get rid of any leftover references to DXT as far as LDMS integrations go. DXT is just a module that captures finer-grained tracing details for POSIX/MPI-IO modules, but I don't think there's any need for LDMS to be aware of it. LDMS should be able to get all the information it needs using the integrations into MPI-IO and POSIX modules directly.

darshan-runtime/lib/darshan-ldms.c
```diff
@@ -0,0 +1,249 @@
...
+        dC.stdio_enable_ldms = 0;
+    else
+        dC.stdio_enable_ldms = 1;
+
+    if (getenv("HDF5_ENABLE_LDMS"))
+        dC.hdf5_enable_ldms = 0;
+    else
+        dC.hdf5_enable_ldms = 1;
+
+    if (getenv("MDHIM_ENABLE_LDMS"))
```
shanedsnyder, line 137: I would actually recommend dropping the MDHIM support, too. This isn't a module that is used for much of anything at this point, so let's not even worry about maintaining LDMS support there. The code won't ever get used, so no point in maintaining it.

shanedsnyder (changes requested):

shanedsnyder:
I just added a bunch of review comments. Mostly just another pass to see how to tidy up the integration overall, but let me know if you have any questions or if anything seems unreasonable. 

I still need to think more on the timespec issue to see what might be the cleanest solution there, but I'm sure we can figure something out.

darshan-runtime/lib/darshan-ldms.c
Snell1224: Yep, I agree. It's not being used anywhere --- will drop it.

Snell1224:
Made the necessary changes to ``jobid``, ``uid``, and ``exename`` and cleaned up the LDMS code.

The ```DARSHAN_LDMS_HOST``` is different from ```gethostname()```. 
```DARSHAN_LDMS_HOST```: Used to connect to the ldmsd (I.e. what host the _ldms daemon_ is running on) 
```gethostname()```: Used for identifying the producer name (i.e. what host the _application_ is running on). 

However, after looking at ```setup_connection()```, you are correct that ```gethostname()``` should not be here. Not only is this the wrong function call (needs to be the host the _daemon_ is running on and not the _application_) but the ```darshan_ldms_connector_initialized()``` already checks if ``DARSHAN_LDMS_HOST`` is set before calling ```setup_connection()```.  This chunk of code has been removed

Thank you for reviewing this! Please let me know if there is any other ldms code that can be cleaned up or doesn't make sense.

File: darshan-runtime/lib/darshan.h (hunk 1 above)
Comment (carns) [reply darshan-runtime/lib/darshan.h:390]: I'm super duper late to this conversation, but we were chatting today about how to wrap up this PR now that Darshan 3.4.3 is finally out the door and I was curious about the change set :)

I have a strong preference not to alter the darshan wtime functions.  That change (to effectively make the function return timestamps in two formats) is making the diff a lot bigger than it ought to be.  For better or worse, relative floating point time stamps are deeply ingrained in Darshan at this point.

I would actually advocate for option 2).  The darshan.h header already has a global offset value (`__darshan_core_wtime_offset`) available in double format (so we can get back to an absolute value while still in floating point format without worrying about overflow) and it doesn't take many floating point operations from there to get back to a timespec.  Someone needs to sanity check this because I try to run it, but I think something like this might work in darshan.h:

```
static inline struct timespec darshan_core_abs_timespec_from_wtime(double t)
{
    struct timespec tp;

    /* add time offset back to get to absolute time */
    t += __darshan_core_time_offset;

    /* cast to integer type, deliberately truncating after decimal point to get whole number seconds */
    tp.tv_sec = (time_t)t;
    /* subtract seconds out, then multiply remainder to nsecs and cast */
    tp.tv_nsec = (time_t)((t - (double)tp.tv_sec) * 1.0e-9);
    return(tp);
}
```

If we had this, then the new ldms_send function could just use the existing available darshan wtime values from an API point of view and then call the above internally to translate to timespec before handing it to LDMS.  The code would look cleaner and I think we would have a difficult time measuring the performance impact of the conversion.

Comment (shanedsnyder) [reply darshan-runtime/lib/darshan.h:390]: Thanks, @carns for the code snippet. I think what you propose here makes a lot of sense. Let us know if this makes sense and/or if you could use some help @Snell1224

Comment (Snell1224) [reply darshan-runtime/lib/darshan.h:390]: @carns @shanedsnyder Thank you for the input and follow up. Yes, this makes sense to me and after reviewing the darshan code again, I agree that it would be much cleaner and easier to add this function than changing any of the existing darshan functions. I will go ahead and implement this code snippet, test it and update this PR with the changes. I'll reach out if I come across any issues or need any help. Thanks!

**Comment by Snell1224** [comment]
I've added the function in darshan.h to convert darshan's floating point times to absolute timestamps and removed all timestructs that were previously initialized to collect this information (i.e. passed through darshan_wtime functions). Please let me know if there's any issues or code changes that need to be made. Thanks! @shanedsnyder @carns

**Comment by shanedsnyder** [comment]
Thanks for making the timestamp changes, @Snell1224 ! That made the diff considerably smaller and a lot easier to review. Your changes all make sense on that front.

I have some smaller review comments I can give, but one immediate thing that stands out as problematic is something I noticed about how the LDMS module is using the `ldms_set_meta()` call. The `set_meta` call is setting some info (rank, filename, record ID) on what appears to be a global LDMS structure whenever a file is opened. On subsequent read/write calls, this metadata information is copied out of the global LDMS structure and sent to LDMS (along with correpsonding details on the read/write operation). The problem, I think, is that this metadata information could easily be overwritten by details related to another file since it's just a single global value? If a process opens 2 separate files before issuing read/write operations, it seems like this will cause issues on properly attributing the I/O to the proper file, for instance.

Does that make sense to you? If so, maybe you have ideas on a straightforward way to correct this. I think we are supposed to have a meeting this coming week, so we can discuss there to make sure we're on the same page.

**Comment by Snell1224** [comment]
@shanedsnyder Yes that does make sense and I see what you mean.  I'm thinking the best way to solve this is to replicate whatever Darshan does to keep track the opened files and their corresponding read/write operation details. From a glance I think this is what the record ID is for? Does Darshan use/have a hash map for mapping the filename and record ID?

If there is a way to get the file name (and path) from the record ID then I can get rid of the "meta_data" function altogether, collect the record ID and rank with "darshan_ldms_connector_send()" and script some code in "darshan_ldms_connector_send()" that gets the filename (and dataset name from hdf5) based on the record id. 

I agree we can discuss this more when we meet. Thanks for catching this though!

**Comment by Snell1224** [comment]
Thanks for the information! I've updated the branch to remove the set_meta function and am now pulling the 'filename' from record ID using the 'darshan_core_lookup_record_name'. 

It looks like there's a build issue with the Darshan-LDMS in the github-ci testing workflow. I'm going to be focusing on fixing this and will give an update when the branch is ready to be looked over again!

File: darshan-runtime/lib/darshan-posix.c [hunk 8]
```diff
@@ -1620,6 +1636,13 @@ int DARSHAN_DECL(close)(int fd)
             rec_ref->file_rec->fcounters[POSIX_F_META_TIME],
             tm1, tm2, rec_ref->last_meta_end);
         darshan_delete_record_ref(&(posix_runtime->fd_hash), &fd, sizeof(int));
+
+#ifdef HAVE_LDMS
+        /* publish close information for posix */
+        extern struct darshanConnector dC;
+        if(!dC.posix_enable_ldms)
```
shanedsnyder, line 1643: I noticed that in the close routine, you're not also checking the value of `dC.ldms_lib` like you do in open/read/write routines. This is true for all modules, not just POSIX.

You might want to add that check just to be safe. This dC struct is only partially initialized, so I think unset values are implicitly initialized to 0. Given my comments elsewhere about some of this logic being ambiguous/inverted, I think it might just be safest to always check the value is as expected?

shanedsnyder (changes requested):

shanedsnyder:
Thanks so much for addressing my previous comments @Snell1224. And sorry for the delay, I'm just now getting a chance to get caught back up on Darshan github traffic.

I did add a handful of final review comments that I think would be nice to address and/or think about. Some of this stuff is pretty minor, but other things we should probably make sure are okay.

I still need to review the build system changes and your new workflow for testing this, but wanted to focus on ironing out the source code first (which I think are really close to final). We also need to think about how we're going to document this feature since it spans both of our tools. We can sort that out later, though, just wanted to mention it in case you've thought about it.

darshan-runtime/lib/darshan-ldms.c
```diff
@@ -0,0 +1,270 @@
...
+
+    /* get the full file path from record ID */
+    filepath = darshan_core_lookup_record_name(record_id);
+
+    if (strcmp(rwo, "open") == 0)
+        dC.open_count = record_count;
+
+    /* set record count of closes to number of opens since we are closing the same file we opened.*/
+    if (strcmp(rwo, "close") == 0)
+        record_count = dC.open_count;
```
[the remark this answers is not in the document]
Snell1224: Yes I see what you mean. The goal was to keep track of the number of closes that occurred (i.e. counter) for each file. For example, if there are "x" number of opens of a specific file then, by the end of the run, there should be "x" number of closes of the same file.

I tried searching through darshan's module code but couldn't find any counters for this I/O event. Does darshan have a counter that keeps track of the number of file closes?  If not, then I was going to try using the record_id and see if it's possible to map the cached ``open_count`` value to the corresponding file.

Also, I'm assuming that all applications close any/all files that were opened before completing (which could be completely wrong). If I am wrong, then I'll just delete this block of code and set it to "-1" (i.e. N/A).

darshan-runtime/lib/darshan-ldms.c
```diff
@@ -0,0 +1,270 @@
...
+    else
+        dC.posix_enable_ldms = 1;
+
+    if (getenv("MPIIO_ENABLE_LDMS"))
+        dC.mpiio_enable_ldms = 0;
+    else
+        dC.mpiio_enable_ldms = 1;
+
+    /* Disable STDIO if verbose is enabled to avoid a recursive
+    function for darshan_ldms_connector_send() */
```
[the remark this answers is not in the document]
Snell1224: You're right, I was seeing a deadlock. Any ``printf`` statement in `darshan_ldms_connector_send()`` or any of the macros would trigger Darshan of a "write" which would call these functions again and thus, output the same ``printf`` statement which would trigger Darshan and the cycle would continue.  The only way to stop this was for the job to time out or I would have to cancel it. 

I have replaced the ``printf`` with ``darshan_core_fprintf()`` and it works! Thanks for the information!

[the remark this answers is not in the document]
shanedsnyder: Ahh, that makes sense. We've definitely seen that kind of infinite loop behavior behavior before when we're accidentally calling the wrapped functions instead of the actual implementation. As you point out, it just basically turns into an endless loop of invoking the wrapper function instead of eventually calling the real implementation.

Glad it was a simple fix!

darshan-runtime/lib/darshan-ldms.c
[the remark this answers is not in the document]
shanedsnyder: Yep, you're right, we don't have any counters keeping track of closes, which complicates a bit what you're trying to do. The reasoning for that on our end was mostly just that it's not a super interesting counter -- it will almost always match the open counter. We do keep track of how much time is spent in close operations (and attribute this to `META_TIME` timer), but that's basically it.

If you really want to capture it, probably you'll have to do something along the lines of what you mention above where you maintain your own map from file record ID to the close count in LDMS code. I was mostly just concerned about the possibility of these erroneous results confusing folks (e.g., this file was opened 1 time but it says it was closed 1000 times...what gives? oh, it's accidentally the counter from another file), so as long as that's cleared up I'm fine with whichever way you want to go here (punt on this functionality or implement it in a separate mapping).

FWIW, I would expect the vast majority of apps to be pretty good about closing each file they open. That said, not closing a file won't likely have negative side effects on most applications, so folks that don't do it (either because they forgot or because they are lazy) aren't really punished anyway. Once the application terminates, the OS will implicitly close all opened files for you. There are OS limits to the number of opened files you can have, so not closing can affect you if you are up against that, but in general forgetting to call it isn't going to result in data loss or anything, so not always that critical.

Snell1224:
@shanedsnyder Thank you for reviewing all LDMS code and making sure it's logical and correct! I've made the suggested changes and decided to include the close counts (described a bit more above). 

I agree, I think it's best to focus on hammering out any issues with the LDMS implemented code first before looking at the github-ci test so please let me know if you find any issues with these updates.

As for the documentation, we think it'd be best to have it on both Darshan and LDMS. I already have some documentation on it (see below) that explains how to build and compile Darshan with the LDMS enabled and how to collect the I/O event data during runtime.  

https://ovis-hpc.readthedocs.io/en/latest/ldms/ldms-streams.html#darshan

I was thinking about emailing the parts directly related to Darshan (i.e. definition, purpose, configuration flags and env variables) and have that added into Darshan's runtime webpage. For any additional details or help, users can be referred to LDMS's readthedocs webpage. Do you think this will be a good path forward? If so, I can go ahead and start working on this.

darshan-runtime/lib/darshan-ldms.c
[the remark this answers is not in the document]
Snell1224: This helps a lot (better understanding of how the OS works) so thank you for the information!

Our LDMS team already have some python analyses that query/use the close counts so I figured it'd be easier to continue collecting this information instead of updating the analyses. I mapped the record_id to the file name via 2 dynamic arrays and it seems to work as expected (ran a test with IOR and mpi-io-test in ``darshan-test/``).

There was a good chunk of code that I added so if you can please review this and see if there're any issues then that would be great!

shanedsnyder:
Thanks for the update @Snell1224 ! I did approve the github workflow to run and it appears it triggered some errors, if you don't mind having a look and trying to resolve them. At first glance, it appears they are related to LDMS changes and not something else that has happened on our main branch.

Snell1224:
> Thanks for the update @Snell1224 ! I did approve the github workflow to run and it appears it triggered some errors, if you don't mind having a look and trying to resolve them. At first glance, it appears they are related to LDMS changes and not something else that has happened on our main branch.

For some reason the struct ``dC`` was not initialized in the module macros. I have another branch where all the development/testing takes place (i.e. lots of commits) and usually merge the changes to this one. However for the most recent merge, github didn't add the ``extern struct darshanConnector dC; \`` lines. Don't know why but I've updated this and the github-ci test should now be working!

darshan-runtime/lib/darshan-ldms.c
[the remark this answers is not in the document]
shanedsnyder: I just looked this over. So, to summarize the current code, it seems that you are caching the number of file opens for each file record on `open` operations, and then using that value for the record count on `close` operations.

I still think that's not quite right. Correct me if I'm wrong, but I'm guessing what you really want to provide to LDMS here is the total number of observed close operations so far for the given file? Since Darshan doesn't track that, I think we would need to track that independently and not use the open count as a substitute for this value as the current code does. Assuming that all sounds right, I think the easiest way to do this is to add a value to the `*_record_ref` structure for close counts, increment that value on close operations, and then pass that value to LDMS send routine? 

E.g., for POSIX:

```
struct posix_file_record_ref
{
    ...
#ifdef HAVE_LDMS
    int64_t close_counts;
#endif
}; 
```

Then in the LDMS code block in the close wrapper, do something like:

```
#ifdef HAVE_LDMS
        rec_ref->close_counts++;
        /* publish close information for posix */
        extern struct darshanConnector dC;
        if(dC.ldms_lib)
            if(dC.posix_enable_ldms)
                darshan_ldms_connector_send(rec_ref->file_rec->base_rec.id, rec_ref->file_rec->base_rec.rank, rec_ref->close_counts, "close", -1, -1, -1, -1, -1, tm1, tm2, rec_ref->file_rec->fcounters[POSIX_F_META_TIME], "POSIX", "MOD");
#endif
```

Does that make sense or am I missing something?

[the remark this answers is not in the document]
Snell1224: Yes that is exactly what I am trying to do and this will work! It was stuck in my mind that the only solution was use ``record_id`` to map the number of opens & closes (which made it much more complicated and confusing than it needed to be) so I didn't even think about this much more simpler/easier way.

I've added these changes and, after testing, the correct number of closes are being collected. Thank you for double checking this!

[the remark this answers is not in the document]
shanedsnyder: Thanks for having a look, I think this looks good now.

darshan-runtime/lib/darshan-hdf5.c
```diff
@@ -433,6 +445,15 @@ herr_t DARSHAN_DECL(H5Fclose)(hid_t file_id)
...
             darshan_delete_record_ref(&(hdf5_file_runtime->hid_hash),
                 &file_id, sizeof(hid_t));
+
+#ifdef HAVE_LDMS
+    rec_ref->close_counts++;
+    /* publish close information for h5f */
+    extern struct darshanConnector dC;
+    if(dC.ldms_lib)
+        if(dC.hdf5_enable_ldms)
+            darshan_ldms_connector_send(rec_ref->file_rec->base_rec.id, rec_ref->file_rec->base_rec.rank, rec_ref->close_counts, "close", -1, -1, -1, -1, rec_ref->file_rec->counters[H5F_FLUSHES],rec_ref->file_rec->fcounters[H5F_F_CLOSE_START_TIMESTAMP], rec_ref->file_rec->fcounters[H5F_F_CLOSE_END_TIMESTAMP], rec_ref->file_rec->fcounters[H5F_F_META_TIME], "H5F", "MOD");
```
shanedsnyder, line 455: Okay, 1 more thing that I think is an error that I hadn't caught previously. I don't mind fixing if you can confirm.

You are calling `darshan_ldms_connector_send()` differently here than in other modules. In other modules, you always pass in `tm1` and `tm2` (which are timers for the start and end time of this specific operation), but here you are passing in the `CLOSE_START_TIMESTAMP` and `CLOSE_END_TIMESTAMP` (which measure the time of the very first close and the very last close) instead.

I'm assuming we want to modify the use of this in the HDF5 modules to always use `tm1` and `tm2` instead? Affected functions are H5Fclose, H5Dclose, H5Dwrite, and H5Dread -- H5Fopen and H5Dopen both already are using `tm1` and `tm2`.

shanedsnyder (changes requested):

darshan-runtime/lib/darshan-hdf5.c
Snell1224: Yes this is correct. It's supposed to be ``tm1`` and ``tm2``. I actually have some changes that I need to update here (just renaming env variables) so I can fix this and add it to the commit. Thanks for catching this!

shanedsnyder:
I pushed some small updates over the last day or so, and think I'm ready to sign off on things. For the source code, I mostly did some refactoring to further reduce the diff (e.g., moving the definition of the `extern dC` variable into the LDMS header so it doesn't have to be repeated everywhere), and also made some whitespace changes to make sure indentations were consistent. I also made some relatively minor updates to the autotools files to simplify and cleanup the output from `configure`.

As far as the github workflow goes, I've only briefly scanned it and the most recent run from my latest commit and noticed that it's not building the LDMS module currently. It looks like a stage of the workflow is explicitly cloning our main branch, whereas I think we want it to just build the code associated with the workflow (in this case, this PR). Maybe have a look at the existing runtime_ci.yml example to see how we are building Darshan directly there?

At any rate, it'd be great if we could get the workflow to run on an LDMS-enabled version of Darshan before we merge so I can sanity check that it's not causing issues while running examples. We should also think about how we could modify the existing workflow to trip some sort of error if the Darshan-LDMS integration isn't working properly (e.g., the workflow is passing now even without LDMS support being built in Darshan). Doing the typical inspection of Darshan logs in this case doesn't tell much other than that Darshan didn't crash and produced some output. Is there anyway to "check" LDMS to see that it has ingested Darshan data? Maybe at the very least, we could enable the verbose option you have for the LDMS module and check that you see some of that output at runtime on stdout/stderr? Any other ideas?

shanedsnyder:
Thanks for the update,  @Snell1224 ! That all sounds great. The key thing I was interested in checking was that there was a way to ensure that LDMS is capturing the Darshan data properly. Sounds like you have a couple of different files/logs you can check for which is perfect.

> Now for checking if Darshan is working, I'm not currently doing anything else other than checking that the .darshan log files exist but I'll look into the runtime_ci.yml and see if there's anything there I could refer to. If you have any other suggestions then please let me know!

I think this is sufficient, for now, probably. We have a more advanced regression testing system for darshan-runtime in the repo, but it's not currently running in any of our workflows (this is on my TODO list) so there's not an example to point to. I can always come back and modify this test to use that more thorough testing if we want, but I think a test that does the following is more than fine:

- ensure application runs fine, terminates without error
- ensure LDMS properly collects data/logs as expected
- ensure Darshan generates a log and that darshan-parser can parse it without error

The main thing I think that would be useful to include here from `runtime_ci.yml` is the code for making sure we build the current branch (rather than `main`): https://github.com/darshan-hpc/darshan/blob/c57f730c0f0d4458c21a427f2a6a64f8d260de22/.github/workflows/runtime_ci.yml#L34

(That would allow the workflow to run against this PR now, but once it's merged into `main`, subsequent workflows will target `main`, if that makes sense.)

Let me know when you have the workflow ready, and I can review and approve it to run to ensure we're getting the expected result on this branch. After that, I think we can go ahead and merge all of this into `main`.

Snell1224:
@shanedsnyder Okay the test script is updated and ready to be reviewed! The script now builds the Darshan-LDMS integrated code and is triggered whenever this is pull/push on this branch.  

Apologies for the delay, that LDMS issue took a bit longer than expected to debug but I finally found out the problem and have a temporary workaround until it's addressed by the other LDMS developers. 

I've added some additional ``ls -ltrch`` commands for the log file directory (towards the bottom of the file) so that we can see all the Darshan log files, their parsed output files, LDMS output and LDMS csv file. Please feel free to make any edits to this .yaml file and/or let me know if there's any issues or questions!

shanedsnyder (approved):

shanedsnyder:
Awesome, I just reviewed and everything looks good. I made a few smaller simplifications (like removing HDF5 and python dependencies that aren't used), but that's all. Looks like the LDMS workflow is running and passing now, which is great to confirm. There are some workflow failures elsewhere, but those appear to be related to some PyDarshan issues not introduced in this PR -- I'm working on getting a fix in for those now in a separate PR.

So, I went ahead and approved this -- I don't think I need any other specific changes from you, unless there was something you still wanted to add? I will add a note to our darshan-runtime docs about the new configure argument to enable LDMS (and will link to the docs you shared previously), but that's my last planned change before merging this PR in.

Thanks a lot for the contribution and for working with us to get the code ready for merge! This is a really neat capability that I hope will lead to some interesting analysis of I/O behavior at runtime.

Snell1224:
> So, I went ahead and approved this -- I don't think I need any other specific changes from you, unless there was something you still wanted to add? I will add a note to our darshan-runtime docs about the new configure argument to enable LDMS (and will link to the docs you shared previously), but that's my last planned change before merging this PR in.
> 
> Thanks a lot for the contribution and for working with us to get the code ready for merge! This is a really neat capability that I hope will lead to some interesting analysis of I/O behavior at runtime.

Okay sounds good! And no there's nothing else that I need to add so it can be merged whenever you're ready!

About the docs, I just emailed a version of the darshan-runtime doc with the added LDMS info so if you haven't started editing the document yet then feel free to refer to this! 

No problem! It was a great learning experience and I'm also excited to see what analyses we can produce with this. Thank you for reviewing and cleaning up the code!

Snell1224:
@shanedsnyder I actually did have one more thing to add which is a env "switch" so that users can choose to turn off/on the LDMS capability. Right now it automatically initializes a connection to the LDMS daemon (when enabled during Darshan build). I'm going to add the commit in a few to this branch and updating the darshan-runtime document that was emailed earlier.

shanedsnyder:
Looks good, only suggestion is that I think we want to use the new env variable you just added (`DARSHAN_LDMS_ENABLE`) in test case 3 for your github workflow, as in test case 1? To ensure nothing bad happens when trying to connect to an LDMS daemon that has shut down?

I'll try to push in the docs updates later today, thanks for sending those to me!

Snell1224:
> Looks good, only suggestion is that I think we want to use the new env variable you just added (`DARSHAN_LDMS_ENABLE`) in test case 3 for your github workflow, as in test case 1? To ensure nothing bad happens when trying to connect to an LDMS daemon that has shut down?

The `DARSHAN_LDMS_ENABLE` is already set for test case 3 because it's under the same step name: "Test Preparation and Run" and it is not being unset in test case 2. From what I understand, github actions keeps the exported variables until that step is complete. I did re-run the test to double check that this is correct and, for test case 3, LDMS threw a connection error (Error 111) and exited but there were no runtime issues with the program or Darshan.

Although I still exported it in test case 3 for ease of mind.

Please let me know if you have any questions about the documentation part!

shanedsnyder (approved):

shanedsnyder:
All sounds good. I just pushed in the docs changes you sent me with a couple small tweaks. I also fixed some of the unrelated CI failures in another PR, so hopefully this passes all checks. Assuming so, I'll go ahead and merge it in. Thanks again for the contribution!

PR #835, opened 2022-10-20
CI: remove shared flag

tylerjereddy:
* based on discussion with Shane, apparently we no longer need to use `--enable-shared` because shared libs are built by default in the "new" build system

* I believe Shane will clean up some docs related to this, like here:
https://www.mcs.anl.gov/research/projects/darshan/docs/darshan-util.html


shanedsnyder (approved):

shanedsnyder:
LGTM, thanks!

PR #840, opened 2022-10-30
MAINT: improve C accum error handling

tylerjereddy:
* returning `-1` from a public API C function isn't sufficient to provide useful error information when working in Python/CFFI--it only tells you something went wrong if you check for a return code in the first place (which isn't normally done in Python anyway--normally you end execution at the error point and `exit()` with an appropriate error message)

* when there are multiple conditions that can trigger the `-1` return value, the situation is even worse, one literally has to `printf` sprinkle the source to figure out what went wrong where

* as a compromise, I'll leave the `-1` approach in since that is quite common in standard `C`, but I'm going to add in prints to `stderr` so that Python can then intercept the `-1` and refer the user to `stderr`

* also, `darshan_accumulator_inject()` assumed that the `module_id` was reasonable because it was set/checked in
`darshan_accumulator_create()`, however my experience in gh-839 was that the accumulator memory location can get freed, or not properly set at CFFI boundary after calling the creation function, so I think the assumption that a previous function was called and worked perfectly is too fragile--I'm adding error handling to prevent a hard segfault on nonsense values of that structure member as a result

### DIFF
--- a/darshan-util/darshan-logutils-accumulator.c
+++ b/darshan-util/darshan-logutils-accumulator.c
@@ -10,6 +10,7 @@
 
 #include <stdlib.h>
 #include <assert.h>
+#include <stdio.h>
 
 #include "darshan-logutils.h"
 #include "uthash-1.9.2/src/uthash.h"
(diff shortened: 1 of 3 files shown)


tylerjereddy:
Unique return codes for different situations would likely be even more useful--i.e., preventing a segfault due to corruption vs. "this module doesn't support accum interface" or whatever.

shanedsnyder:
I agree the changes in the C library here are definitely helpful.

I think I'd vote for us to fix the issues in #839 first, and then rebase on that and adopt the same style code for the test in this PR just for consistency.

Issue #847, opened 2022-11-09
Query: bandwidth calculation for derived metrics


tylerjereddy:
I'm curious how the bandwith portion of this line is being calculated for logs repo file `laytonjb_test1_id28730_6-7-43012-2131301613401632697_1.darshan`:
![image](https://user-images.githubusercontent.com/7903078/200915679-082b8b40-0cef-4466-8300-ee9b5c0363d9.png)

 which seems to have only a single record/single rank. My math is (ignoring floating truncation in the prints): `151` / `0.000004` / (2 ** 20) = approx. 36 MiB / s

Here are the `counter` and `fcounter` dataframes for this 1-record/1-rank log -- can you clarify which numbers are being used here for the calculation to get `4.22 MiB/s`?

```
.counters_df:
    rank                    id  STDIO_OPENS  STDIO_FDOPENS  STDIO_READS  \
0     0  17359025305976423428            1              0            0   

   STDIO_WRITES  STDIO_SEEKS  STDIO_FLUSHES  STDIO_BYTES_WRITTEN  \
0            10            0              0                  151   

   STDIO_BYTES_READ  STDIO_MAX_BYTE_READ  STDIO_MAX_BYTE_WRITTEN  \
0                 0                    0                     150   

   STDIO_FASTEST_RANK  STDIO_FASTEST_RANK_BYTES  STDIO_SLOWEST_RANK  \
0                   0                         0                   0   

   STDIO_SLOWEST_RANK_BYTES  
0                         0  
fcounters_df:
    rank                    id  STDIO_F_META_TIME  STDIO_F_WRITE_TIME  \
0     0  17359025305976423428            0.00003            0.000004   

   STDIO_F_READ_TIME  STDIO_F_OPEN_START_TIMESTAMP  \
0                0.0                      0.025467   

   STDIO_F_CLOSE_START_TIMESTAMP  STDIO_F_WRITE_START_TIMESTAMP  \
0                       0.025565                       0.025558   

   STDIO_F_READ_START_TIMESTAMP  STDIO_F_OPEN_END_TIMESTAMP  \
0                           0.0                    0.025487   

   STDIO_F_CLOSE_END_TIMESTAMP  STDIO_F_WRITE_END_TIMESTAMP  \
0                     0.025575                     0.025565   

   STDIO_F_READ_END_TIMESTAMP  STDIO_F_FASTEST_RANK_TIME  \
0                         0.0                        0.0   

[...]

PR #848, opened 2022-11-09
WIP, DEBUG: Python derived metrics

tylerjereddy:
* this is a Python/pandas-only version of doing some simple derived metrics; I don't think we'll
actually do this, but I was exploring a bit because of the difficulties in gh-839

* this matches pretty well with the `perl` based reports for total bytes, but even simple cases can sometimes disagree on bandwidth per gh-847, so now I'm curious what is going on

* one problem with doing this is that we'd have the same algorithms implemented in two different languages; the advantages include:
  - not reading all the records in a second time, one at a time, crossing the CFFI boundary each time (because all the records are already stored in the DataFrame from the first time we did this)
  - easier to debug/maintain because bounds checking/no segfaults, etc.
  - likely easier to regex-filter directly on the pandas data structures than with interop to C/CFFI for i.e., custom file derived metrics

### DIFF
--- a/darshan-util/pydarshan/darshan/derived_metrics.py
+++ b/darshan-util/pydarshan/darshan/derived_metrics.py
@@ -7,11 +7,11 @@ def perf_estimate(report, mod_name: str):
     # that it might be more natural to use a library
     # like humanize to automatically select i.e., GiB
     # depending on magnitude
-    mod_name_adjusted = mod_name.replace("-", "")
-    total_mebibytes = (counters_df[f"{mod_name_adjusted}_BYTES_WRITTEN"].sum()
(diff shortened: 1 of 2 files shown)


tylerjereddy:
Ok, CI issues notwithstanding, the new `test_perf_estimate()` does pass for two log files here. I won't carry this any farther for now because it looks like things can get more complex with `MPI-IO` involvement and the logic is already written in C.

Still, we may do this someday for some of the reasons outlined above, especially if it is more performant (no re-reading records across langauge barriers) and more maintainable.

tylerjereddy:
We're not doing this for now, though long-term it is likely far easier to maintain via `pandas` or something high-level.

PR #849, opened 2022-11-09
CI: try to fix MacOS CI

tylerjereddy:
* switch to `DYLD_FALLBACK_LIBRARY_PATH` for `pydarshan` shared library searching on MacOS in GitHub actions

* note that we're already doing this for the M1 Mac testing in Cirrus CI, and I need it locally as well on newer versions of MacOS

* we'll see if CI agrees with my assessment (I don't know all the technical details for justification..)

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -75,6 +75,7 @@ jobs:
       - name: Test with pytest
         run: |
           export LD_LIBRARY_PATH=$PWD/darshan_install/lib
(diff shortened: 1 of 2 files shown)


tylerjereddy:
There was a second issue after the initial one--new version of `mypy` released two days ago--attempted fix for that added in here as well.

shanedsnyder (approved):

shanedsnyder:
LGTM, thanks for the quick fix!

PR #854, opened 2022-11-23
bug fix: when MPI is disabled

wkliao:
These 2 APIs contain an argument of type MPI_Datatype.
```c
   static void darshan_core_reduce_min_time(
       void* in_time_v, void* inout_time_v,
       int *len, MPI_Datatype *datatype);
   static void darshan_core_reduce_max_time(
       void* in_time_v, void* inout_time_v,
       int *len, MPI_Datatype *datatype);
```

### DIFF
--- a/darshan-runtime/lib/darshan-core.c
+++ b/darshan-runtime/lib/darshan-core.c
@@ -151,12 +151,14 @@ static int darshan_deflate_buffer(
 static void darshan_core_cleanup(
     struct darshan_core_runtime* core);
 static void darshan_core_fork_child_cb(void);
+#ifdef HAVE_MPI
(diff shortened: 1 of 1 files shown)


wkliao:
I also suggest to add a CI to test when MPI is disabled, i.e. `--without-mpi` is used.

shanedsnyder:
Thanks for catching this @wkliao. I'm working on adding a CI test for this in #855. Once, I have something setup I'll merge this fix in.

shanedsnyder:
Okay, I reproduced in CI in #855. I'll merge this in and keep fleshing out the test there. Thanks!

shanedsnyder (approved):

PR #856, opened 2022-11-30
CI: bump cibuildwheel to newest release

shanedsnyder:
bump to v2.11.2

note there's also a typo fix in current path inclusions (forgotten `*`), as well as addition of a new inclusion for updates to the build wheels workflow file

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -13,7 +13,7 @@ on:
       - main
     paths:
       - darshan-util/**
-      - include/*
+      - include/**
       - .github/workflows/build_wheels.yml
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


tylerjereddy (approved):
LGTM, I spot checked the Python `3.11` `pytest` output on both platforms and it looks solid. The CI cancellation is unrelated--that'll be fixed by gh-858.

I don't know how much the `path` corrections matter for commit-message driven triggers for wheel builds (?) but I don't mind either way as long as we don't get expensive wheel builds more often. Presumably we're still gated on the commit message.

shanedsnyder:
> I don't know how much the `path` corrections matter for commit-message driven triggers for wheel builds (?) but I don't mind either way as long as we don't get expensive wheel builds more often. Presumably we're still gated on the commit message.

Yeah, we are definitely still gated by the commit message, I did confirm that, so I think we should be fine here.

Issue #828, opened 2022-10-09
MAINT, TST: portability of test_forked_process


tylerjereddy:
This is pretty minor, but as mentioned at https://github.com/darshan-hpc/darshan/pull/827#discussion_r990857501, we may want to consider using `sys.executable` to portably retrieve the Python exe path in that test rather than plain `python`--I needed to manually change the test to use `python3` when using `act` locally.

PR #857, opened 2022-11-30
CI: add non-MPI test case

shanedsnyder:
* builds a non-MPI version of darshan-runtime
* adds MPI and non-MPI versions of existing runtime CI tests for forked processes

### DIFF
--- a/.github/workflows/runtime_ci.yml
+++ b/.github/workflows/runtime_ci.yml
@@ -71,6 +71,7 @@ jobs:
         run: |
           export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PWD/darshan_install/lib
           export DARSHAN_INSTALL_PATH=$PWD/darshan_install
+          export DARSHAN_NONMPI_INSTALL_PATH=$PWD/darshan_nonmpi_install
           export DARSHAN_ROOT_PATH=$PWD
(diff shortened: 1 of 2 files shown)


tylerjereddy (approved):
I confirmed that this feature branch fails the runtime testing locally when I use `act` for containerized testing (because it lacks gh-854) as expected with `darshan-core.c:156:15: error: unknown type name 'MPI_Datatype'` at build time. That doesn't happen in CI, because GitHub actions auto-merges into `main` before starting.

When I rebase this feature branch on latest `main`, it fixes the build error and gets to the testing stage. The tests don't all pass with `act` locally in that situation, but it is something I've seen before re: discrepancy b/w `act` and GitHub actions proper with `python` vs. `python3`:

```
| FAILED darshan-test/python_runtime_tests.py::test_forked_process_nonmpi - FileNotFoundError: [Errno 2] No such file or directory: 'python'
| FAILED darshan-test/python_runtime_tests.py::test_forked_process_mpi - FileNotFoundError: [Errno 2] No such file or directory: 'python'
```

In fact, I have an issue open for that gh-828, so we can ignore that here, though we should patch it eventually.

The changes themselves look fine to me, and the CI timeout will be solved by gh-858.

PR #858, opened 2022-11-30
BUG: fix pytest hang/lock

### LINKED ISSUE #851: CI: Python 3.10 on ubuntu-latest issue
As discussed in gh-830, there is a mysterious "hang" of the `pytest` suite with Python `3.10` and `ubuntu-latest` GitHub actions runner on that branch. Logging in to the runner via `ssh` allows execution via `pytest` as expected, so it is pretty unusual/weird. Furthermore, in https://github.com/tylerjereddy/darshan/pull/23 I checked that `pytest` `6.2.5` and Python `3.10.0` had no effect on the hang (vs. the newer versions of those libs in use for the original hang).

I'll keep this issue open as a reference, but since I couldn't reproduce even in the runner itself when working interactively, nor in `act` with Python `3.10` and `ubuntu-latest`, there seems to be no sane way to debug this. It looks like switching to the latest Ubuntu LTS release, using the `ubuntu-22.04` tag alleviates the issue so I'll probably try applying that to Shane's PR and referencing this issue in a comment.

Unless we get related user reports, I wouldn't suggest putting any more time on this one though...

tylerjereddy:
Fixes #851

* the testsuite now always uses `DarshanReport` with a context manager to avoid shenanigans with `__del__` and garbage collection/`pytest`/multiple threads; another bonus is that this avoids constructing `DarshanReport` objects in `pytest` parameters, which would explain why reports were being built up during collection stage before we even start running (which, if you add CFFI into the mix, may be asking for trouble)

* this appears to fix the problem with testsuite hangs described in gh-839 and gh-851; I pushed this commit into gh-839 recently so if the CI there stops hanging with `3.10` on top of my local confirmation, hopefully we're good to go on this annoyance

* if the fix is confirmed by the CI over there, I do suggest we encourage the use of `DarshanReport` with a context manager in our documentation--perhaps we could open an issue for doing that and maybe looking for cases in our source (beyond the tests) where we may also consider the switchover. Here, I've tried to prioritize what we really need to avoid the hang in the short-term at least.

### DIFF
--- a/darshan-util/pydarshan/darshan/tests/test_cffi_misc.py
+++ b/darshan-util/pydarshan/darshan/tests/test_cffi_misc.py
@@ -57,10 +57,10 @@ def test_file_hash_type(log_path):
 
     # additionally check that the dataframes
     # generated are of the correct types
-    report = darshan.DarshanReport(log_path, read_all=True)
-    report.mod_read_all_records("POSIX", dtype="pandas")
-    rec_counters = report.records["POSIX"][0]["counters"]
-    rec_fcounters = report.records["POSIX"][0]["fcounters"]
+    with darshan.DarshanReport(log_path, read_all=True) as report:
+        report.mod_read_all_records("POSIX", dtype="pandas")
+        rec_counters = report.records["POSIX"][0]["counters"]
+        rec_fcounters = report.records["POSIX"][0]["fcounters"]
     # verify the records returned have the correct
     # data type for the ids/hashes
     assert rec_counters["id"].dtype == np.uint64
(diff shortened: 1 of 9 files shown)


shanedsnyder:
Awesome, thanks for diagnosing that and applying the fix! I know debugging the issue was bad enough, but applying the changes to tests looks not so fun either, so thanks for the help.

I took a close look at the changes and think it all makes sense. I agree that parameterizing based on log file paths looks much cleaner (and also does explain pretty well why we were accessing logs before running tests, that was always a mystery to me), so that's nice. 

Prioritizing testing, then circling back to docs/examples/notebooks makes sense to me in terms of priorities, too. I'll go ahead and merge this in now. I'll open an issue to remind us about docs and such, too.

Issue #859, opened 2022-11-30
DOC: update PyDarshan docs/examples to use context manager for report objects


shanedsnyder:
See #851 for original issue and #858 for solution.

Long story short, it seems that using a context manager (i.e., `with darshan.DarshanReport(log_path) as report:`) for report objects helps avoid potential memory/threading issues when releasing them. This ultimately was causing some of our pytest runs to hang in various contexts.

We should update our docs and various examples to use this approach for managing report objects so users don't get potentially bitten by this.

PR #863, opened 2022-12-01
DOC: fixup pydarshan docs

shanedsnyder:
* cleanup some usage of "pydarshan" (for either "darshan" or "PyDarshan")
* cleanup docs usage of Darshan report objects to use context managers

### DIFF
--- a/darshan-util/pydarshan/README.rst
+++ b/darshan-util/pydarshan/README.rst
@@ -17,7 +17,7 @@ Features
 Usage
 -----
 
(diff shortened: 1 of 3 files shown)


shanedsnyder:
The 2nd bullet there should at least partially address #859 but I'll leave that open so we can make a pass over Jupyter notebooks and anywhere else. I just fixed what I saw in the docs files I was scanning.

tylerjereddy (approved):
LGTM

We still have non-context uses in the source of `darshan` lib itself, so makes sense to leave issue open for now.

Issue #867, opened 2022-12-06
file category count mismatch between darshan-parser and darshan-job-summary.pl


carns:
Reproduce by running `darshan-parser --show-incomplete --file` and `darshan-job-summary.pl` on the log at https://github.com/darshan-hpc/darshan-logs/blob/main/darshan_logs/imbalanced_io/imbalanced-io.darshan .

The former shows this in the POSIX summary section:
```
# <file_type> <file_count> <total_bytes> <max_byte_offset>
# total: 1026 106730099902 52938398983
# read_only: 12 852195214 575999999
# write_only: 2 84608 24293491975
# read_write: 1 105877820080 52938398983
# unique: 1023 852195214 575999999
# shared: 3 105877904688 77231890958
```
This looks to be correct according to manual inspection of the log file.  The total count is higher than the sum of the readonly, writeonly, and readwrite categories because most files are recorded as being opened but having no I/O traffic in this (incomplete) log example.

The darshan-job-summary.pl reports this, however:

![image](https://user-images.githubusercontent.com/5541937/205958935-720426a4-fddb-4b0b-a3c1-01db6cdf5d03.png)

We should understand what's different about the perl logic that produces a different result here to make sure we aren't missing something in the logutils logic.

carns:
FYI @tylerjereddy

carns:
On closer inspection, I don't think that either of these is technically wrong; we just need to be careful about wording.

The `darshan-job-summary.pl` tool calculates these category metrics in perl based on a hashtable of all of the files that it has encountered in the full parser output.  However, it skips any file record that doesn't have a non-zero open counter: https://github.com/darshan-hpc/darshan/blob/main/darshan-util/darshan-job-summary/bin/darshan-job-summary.pl.in#L1167 .  So it's deliberately only tracking files that have literally been opened.

The log in question includes file records that in fact were not opened; they were only stat'd:

```
carns-x1-7g ~/w/s/d/d/b/d/d/bin (main=)> ../../darshan-parser ~/working/src/darshan/darshan-logs/darshan_logs/imbalanced_io/imbalanced-io.darshan --show-incomplete |grep POSIX_OPENS |head -n 5
# WARNING: POSIX_OPENS counter includes both POSIX_FILENOS and POSIX_DUPS counts
POSIX	0	18041004828094502	POSIX_OPENS	0	/lus/theta-fs0/972582213	/lus/theta-fs0	lustre
[...]

PR #868, opened 2022-12-06
ENH: file count summary

tylerjereddy:
* this builds on gh-839, and that PR should
    be reviewed/merged first
    
* add the "File Count Summary" to the Python
    summary reports based on the table with the same
    name in the Perl reports; note considerable
    discussion in this branch/code comments about differences
    with the old reported values, and see also
    gh-867
    
* an additional contrast with the old Perl reports
    is that the above table is reported for each module
    rather than "just POSIX"
    
* some other considerations:
    - unlike the parent PR/branch, this branch aims to start
        moving new accumulator infrastructure/testing to a more
        appropriate location than the CFFI backend modules (since
        much of the work is higher-level than that, etc.)
    - similar to parent PR, we may want to do some checking for
        memory leaks because they are just so easy to introduce with the
        poor static analysis/tooling around the CFFI barrier
    - we can probably bikeshed on the table formatting a bit; I'll paste
        a few samples from Python summary reports below
    - we have a new dependency on `humanize` here--not sure on tradeoffs
        there, we've discussed using this lib for file size conversions
        before though
    - like the parent PR, type hints and docstrings for new function(s)
        probably could use some work

Samples from `imbalanced-io.darshan`:

image

image

image


tylerjereddy:
CI error is `Version 3.6 was not found in the local cache` -- I think it really is time to dump `3.6` support if they're starting to do brownouts, etc. That version has been completely EOL for a full year--it really is time for HPC centers to upgrade or be left behind on that one I'd say.

darshan-util/pydarshan/darshan/backend/cffi_backend.py
```diff
@@ -657,3 +658,82 @@ def _log_get_heatmap_record(log):
...
+                           f"This could mean that the {mod_name} module does not "
+                           "support derived metric calculation, or that "
+                           "another kind of error occurred. It may be possible "
+                           "to retrieve additional information from the stderr "
+                           "stream.")
+
+    buf = ffi.new("void **")
+    r = 1
+    while r >= 1:
+        r = libdutil.darshan_log_get_record(log_handle['handle'], modules[mod_name]['idx'], buf)
```
carns, line 703: This code is explicitly opening the log and iterating through the records to produce the derived metrics.  The hope with the accumulator API, though, was that it could be applied (kind of like a filter) on records during the first/main pass rather than going back and doing another pass over the log.  The calculations done by the accumulator functions are modest compared to the cost of retrieving records.

I don't know what performance impact that has in practice, though, or how awkward it would be from a code organization point of view.

tylerjereddy: The easiest thing for me would be to have the records read in to a single contiguous buffer that I could then pass around. I think we've discussed that before.

As it stands, we add the records 1-by-1 to pandas Dataframes for counters and fcounters, but the C API doesn't support/accept Python objects holding the records. Writing the accumulator logic to work directly on on our record (pandas) dataframes would allow:

1) Not reading the data in again
2) Easily filtering the dataframe (records) using regexes in Python/pandas, etc.
3) Maintaining the analyses at a higher level
4) The ability to use GPUs/concurrency on large record dataframes with API-matched pandas backends like: https://github.com/rapidsai/cudf 

I'm not sure we'll come to an agreement on the path forward though in the short term. My suggestion would be to aim for correctness now and efficiency later. Note that we probably want to review the base code in the parent PR rather than this extension for file tables.

What I don't want is to basically end up having to write a bunch of C-like code at the CFFI level just to use the API more efficiently--if we're using dataframes to store the records I think long-term we should produce single dataframes for counters and fcounters, and then do any work on them directly. Since the logic is already in C, we're avoiding the duplication for now, but with all the work the community puts into accelerated dataframe operations, I'm not so sure we gain much long-term with hand-writing filters vs. the accelerator backends popping up.

The only problem with this is that we need to dump everything in to memory/buffer, which is also something we've discussed before.

carns: Ok, got it.  That's a good synopsis of the issue that helps clarify what's going on.  Agree with the strategy to get correctness now and then circle back to it.  Do you want copy that synopsis into an issue for how to optimize the aggregate metrics?

carns (approved):
LGTM

darshan-util/pydarshan/darshan/lib/accum.py
```diff
@@ -0,0 +1,64 @@
...
+import humanize
+
+
+def log_file_count_summary_table(log_path: str,
+                                 module: str):
+    # the darshan_file_category enum is not really
+    # exposed in CFFI/Python layer, so we effectively
+    # re-export the content indices we need here
+    # so that we can properly index the C-level data
+    darshan_file_category = {"total opened":0,
```
carns, line 20: Whoops actually, the suggested change her still stands to make this "total files" rather than "total opened" since technically the count will include things that may not have literally been opened.

carns (changes requested):

darshan-util/pydarshan/darshan/backend/cffi_backend.py
shanedsnyder: FWIW, I'm fine just targeting correctness here, as well.

I manually compared time to generate reports for the E3SM-IO test cases (one with DXT data, one without)  on this branch and compared to `main`, and I don't see any obvious slowdown, though it's clearly costing more to make 2 passes over log data. Guess that means other aspects of report generation are dominating overheads.

I was thinking we could potentially add some shims to CFFI bindings / report interface to automatically add records to an accumulator as they are read in (i.e., before we fully convert them to our Python format and free the memory associated with the C-style format). That would keep us at one pass, and we could maintain the accumulators and derived metrics within the report API so other tools could use them. However, that kind of restricts the generality of accumulators at the PyDarshan level (users couldn't just create an accumulator for a subset of files they are interested in, they would only be implicitly maintained on a per-module basis). If we want to keep the accumulators more general (as intended), we'd still need to solve the general problem of converting from the Python format back to what the C library expects...

shanedsnyder: Looking closer at the code here, I think we're at 3 total passes over the Darshan log, though. One for the traditional job summary pass we've always one done, one for per-module bandwidth calc, and one for the per-module file count table. Could we just cache the derived metrics and aggregate record data returned from emit for each module rather than doing that twice?

tylerjereddy: `log_get_derived_metrics()` is the shared entrypoint if that is what you mean. We could try adding an [`@lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) decorator, as long as the hashability requirements are met. I think if you grep the Python part of the codebase you'll see a few other uses of that construct.

I think we're on the same page re: Python vs. C workflow awkwardness. If you want to make it nice and convenient to filter the records prior to accumulator/derived metrics processing at the Python level, it looks messy to me at the moment re: regex application to a C data structure instead of `pandas`, etc.

shanedsnyder:
I think I have this all synced with changes from #839. I need to review the lib/accum code and the tests closer, but everything else looks good so far.

darshan-util/pydarshan/darshan/lib/accum.py
```diff
@@ -0,0 +1,116 @@
...
+            avg_size = math.ceil(float(avg_size)) # type: ignore
+            if avg_size == 0:
+                avg_size = "0"
+            else:
+                avg_size = f"{avg_size} {binary_units}"
+        df.iloc[index] = [index, num_files, avg_size, max_size]
+    # we don't need the index column once we're done with
+    # the CFFI/C interaction
+    df.drop(columns="index", inplace=True)
+    ret = plot_common_access_table.DarshanReportTable(df,
```
shanedsnyder, line 113: Could we generalize DarshanReportTable, so that it isn't tied to the `plot_common_access_table` module (here or maybe we can open an issue if it's involved)? I guess they are the only existing tables in the report, but a "common access table" is a specific table instance and it will be awkward to have to create all tables this way.

darshan-util/pydarshan/darshan/lib/accum.py
```diff
@@ -0,0 +1,116 @@
...
+        num_files = int(cat_counters.count)
+        if num_files == 0:
+            max_size = "0"
+            avg_size = "0"
+        else:
+            max_size, binary_units = humanize.naturalsize(cat_counters.max_offset_bytes + 1,
+                                                          binary=True).split()
+            # some rounding work to match the old Perl
+            # report table style...
+            max_size = math.ceil(float(max_size)) # type: ignore
```
shanedsnyder, line 96: Instead of always rounding up to nearest integer, could we just round to a couple of decimal places all of the time? Reporting both a 1.01GiB and a 1.99GiB file as 2GiB is probably confusing, for instance.

I don't think the Perl reports were ever intentionally trying to round things, but I do see that a lot of times the reported values are integers, but not always (from `ior_hdf5_example.darshan`):
image

In any case, it's probably most clear to provide a little more precision.

darshan-util/pydarshan/darshan/lib/accum.py
```diff
@@ -0,0 +1,116 @@
...
+            if max_size == 0:
+                max_size = "0"
+            else:
+                max_size = f"{max_size} {binary_units}"
+            # NOTE: internal formula based on discussion with Phil Carns
+            avg_size, binary_units = humanize.naturalsize((cat_counters.total_max_offset_bytes + num_files) / num_files,
+                                                           binary=True).split()
+            avg_size = math.ceil(float(avg_size)) # type: ignore
```
shanedsnyder, line 104: Same for rounding here.

darshan-util/pydarshan/darshan/tests/test_lib_accum.py
```diff
@@ -0,0 +1,211 @@
...
+    # the file count summary table for all modules, for
+    # example often only showing for POSIX, so in those
+    # cases we really just verify the file count and
+    # the other columns are regression guards against
+    # what we currently have (max size may be available
+    # in a subset of these cases as well)
+    ("e3sm_io_heatmap_only.darshan",
+     "POSIX",
+     # <file count> <avg size> <max size>
+     [[3, "100 GiB", "298 GiB"],
```
shanedsnyder, line 125: Not actionable yet, but I want to have a closer look at some of these test cases to make sure the values are all expected and make sense. This case, for instance, actually has less "total files" reported here than in the Perl report, which I'm not sure I understand at all -- other examples of mismatches we've seen in the past were with the Perl report having _less_ files than reported by the parser (since it's doing some filtering), but don't think I expected to see _more_ files reported by Perl summary.

shanedsnyder (changes requested):

darshan-util/pydarshan/darshan/lib/accum.py
shanedsnyder: Fixed via [c61a024](https://github.com/darshan-hpc/darshan/pull/868/commits/c61a024b90547163477936397918689477aeeae3)

darshan-util/pydarshan/darshan/lib/accum.py
tylerjereddy: I think you're just repeating my review comment from https://github.com/darshan-hpc/darshan/pull/618#issuecomment-1018889780

> DarshanReportTable should be in a general data structure-related module somewhere most likely, I'll leave that up to you to discuss with the team.

tylerjereddy: I opened gh-892 for you

shanedsnyder: okay, thanks, i'll resolve this for now and look to review/merge gh-892 after this is merged.

darshan-util/pydarshan/darshan/tests/test_lib_accum.py
shanedsnyder: I don't think anything is actually wrong here. I forgot that the Perl report is summing these file statistics across POSIX and MPI-IO modules, so it's not surprising that if we only sum stats across the POSIX module here that there will be some mismatches.

shanedsnyder:
Any objections if I make the rounding change I mentioned (and update affected tests), and then merge this?

tylerjereddy:
> Any objections if I make the rounding change I mentioned (and update affected tests), and then merge this?

No objections from me--the diff looks fine/CI is currently passing, and I don't have an opinion on how you should round floats in the summary report, so I'd say go for it.

The performance stuff is going to be harder--we keep bumping up against the log format only releasing one record at time, which is just hitting "don't iterate in Python" performance guidance basically.

shanedsnyder (approved):

darshan-util/pydarshan/darshan/lib/accum.py
shanedsnyder: done in [aff9aaa](https://github.com/darshan-hpc/darshan/pull/868/commits/aff9aaa071e578fc129d6779bd53165abca1e3fe)

darshan-util/pydarshan/darshan/lib/accum.py
shanedsnyder: done in [aff9aaa](https://github.com/darshan-hpc/darshan/pull/868/commits/aff9aaa071e578fc129d6779bd53165abca1e3fe)

shanedsnyder (approved):

Issue #866, opened 2022-12-06
Segmentation fault when instrumenting PnetCDF when defining a scalar variable


yzanhua:
## Summary
A segmentation fault will occur when using `ncmpi_def_var` to define a scalar variable, using Darshan 3.4.1.

## How to reproduce
A test program to reproduce the problem (`test.c`) together with a simple `makefile` is attached below. A segmentation fault will occur at `ncmpi_def_var`. The output from `gdb` is also attached.

`make` to compile.
`make run` to run the program without Darshan (this step will not cause any problem).
`make run-darshan` to run the program with Darshan, a segmentation fault will occur.

<details> <summary> click here to see test.c </summary>

```c
#include <mpi.h>
#include <pnetcdf.h>
#include <stdio.h>
#include <stdlib.h>

#define CHECK_ERR                                           \
    {                                                       \
        if (err < 0) {                                      \
            printf("Error at %s:%d\n", __FILE__, __LINE__); \
        }                                                   \
    }

int main(int argc, char **argv) {
    int ncid, err = 0, varid;

    MPI_Init(&argc, &argv);

    /* create a file */
    err = ncmpi_create(MPI_COMM_WORLD, "test.nc", NC_CLOBBER | NC_64BIT_OFFSET, MPI_INFO_NULL, &ncid);
    CHECK_ERR;

    /* define a scalar variable of integer type */
    err = ncmpi_def_var(ncid, "scalar_var", NC_INT, 0, NULL, &varid);
    CHECK_ERR;

    err = ncmpi_enddef(ncid);
    CHECK_ERR;

    err = ncmpi_close(ncid);
    CHECK_ERR;

    MPI_Finalize();
    return 0;
}
```

</details>

<details> <summary> click here to see makefile </summary>

```makefile
PNETCDF_DIR=${LOCAL_HOME}/PnetCDF/1.12.3
DARSHAN_DIR=${LOCAL_HOME}/Darshan/3.4.1

CC=mpicc
RUN=mpirun

DARSHAN_SO=${DARSHAN_DIR}/lib/libdarshan.so

all:
	${CC} -g -O0 test.c -o test -I${PNETCDF_DIR}/include -L${PNETCDF_DIR}/lib -lpnetcdf

run:
	LD_LIBRARY_PATH=${PNETCDF_DIR}/lib:${LD_LIBRARY_PATH} \
	${RUN} -n 1 ./test

run-darshan:
[...]

PR #869, opened 2022-12-07
bug fix: check if a variable is a record variable

wkliao:
This is to fix issue #866

* Scalar variable is never a record variable
* When defining a scalar variable, the ndums argument is 0 and dimids argument can be NULL
* Also check whether the record dimension has been defined in the file. If not, no variable can be a record variable.


shanedsnyder:
Yikes, I'm not sure how I missed this one. I did confirm the fix, so will go ahead and merge this in.

shanedsnyder (approved):

Issue #510, opened 2021-10-01
MAINT: when Python 3.6 is dropped


tylerjereddy:
Currently, Argonne requires Python `3.6` support, which is now passing CI in gh-505. This issue is a reminder of some features we can reactivate when we can ditch Python `3.6`:

- [ ] restore usage of `from __future__ import annotations` wherever needed so that we can leverage PEP 563 postponed evaluation type hints
- [ ] remove the mocking from `darshan-util/pydarshan/darshan/experimental/plots/plot_dxt_heatmap.py` for `numpy.typing as npt`

tylerjereddy:
I'd say it is time to move on from `3.6` now--see https://github.com/darshan-hpc/darshan/pull/868#issuecomment-1340018256

We should also just error out on import/install attempt on `main` if you are using less than `3.7`. To be honest, the ecosystem is mostly at `3.8+` now, and soon `3.9+` I think.

PR #870, opened 2022-12-07
CI: drop 3.6 & ubuntu switch

tylerjereddy:
* the Github Actions infrastructure is progressively phasing in a newer base image of Ubuntu:
https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/

* that means that we are somewhat randomly going to see images that lack Python `3.6` from the build cache sometimes per:
https://github.com/actions/setup-python/issues/355#issuecomment-1335042510

* instead of pinning to an old version of Ubuntu in GHA for 3.6 support, let's just drop 3.6 from the testing matrix per:
https://github.com/darshan-hpc/darshan/issues/510#issuecomment-1340023264 (it has been EOL for 1 year)

* there's also no reason to retain the special Python `3.10` handling where we use Ubuntu `22.04` for that case, for two reasons: 

1) `22.04` is being rolled out as the new default anyway 
2) gh-851 with the Python garbage collection issues in `3.10` was resolved by using contexts, so special treatment not justified anymore

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -21,14 +21,7 @@ jobs:
       matrix:
         platform: [ubuntu-latest,
                    macos-latest]
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
-        # see gh-851
-        exclude:
-          - platform: ubuntu-latest
-            python-version: "3.10"
-        include:
-          - platform: ubuntu-22.04
-            python-version: "3.10"
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


shanedsnyder (approved):

shanedsnyder:
To fully remove on from 3.6, there are still some other tasks, correct? Off the top of my head, I guess pyproject.toml needs to be updated so cibuildwheel doesn't build wheels for it, but probably other things, too, I'm guessing.

Are those out of scope here? Or is it worth ripping the bandaid off?

tylerjereddy:
I was just going to remove the CI support to start, I don't really care if we continue here though, but you'll end up with randomly failing CI for awhile if you wait

shanedsnyder:
Maybe I'll just merge this in, and can start up a more general PR shortly, just to avoid any distractions on any other PRs in the mean time.

PR #871, opened 2022-12-08
BUG: fix bug causing misattribution of pnetcdf_var i/o

shanedsnyder:
* various put/get API calls were all using the ncid rather than the varid to retrieve variable records

### DIFF
--- a/darshan-runtime/lib/darshan-pnetcdf-api.m4
+++ b/darshan-runtime/lib/darshan-pnetcdf-api.m4
@@ -781,7 +781,6 @@ define(`PNETCDF_VAR_RECORD_OPEN',`
         else type_size = 8;
         rec_ref->var_rec->counters[PNETCDF_VAR_DATATYPE_SIZE] = type_size;
         rec_ref->var_rec->file_rec_id = file_rec_id;
-        fprintf(stderr, "adding %d to var ncid hash\n", *($6));
         darshan_add_record_ref(&(pnetcdf_var_runtime->ncid_hash), $6, sizeof(int), rec_ref);
     } while (0);
 ')dnl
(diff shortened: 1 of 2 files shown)


darshan-runtime/lib/darshan-pnetcdf-api.m4
```diff
@@ -234,7 +234,7 @@ int DARSHAN_DECL(APINAME($1,$2,$3,$4))(int ncid, int varid, ArgKind($2)BufArgs($
     if (ret == NC_NOERR) {
         PNETCDF_VAR_PRE_RECORD();
         struct pnetcdf_var_record_ref *rec_ref;
-        rec_ref = darshan_lookup_record_ref(pnetcdf_var_runtime->ncid_hash, &ncid, sizeof(int));
+        rec_ref = darshan_lookup_record_ref(pnetcdf_var_runtime->ncid_hash, &varid, sizeof(int));
```
carns, line 237: This looks odd to look up `varid` in something called `ncid_hash`.  Is this because the ncid_hash data structure is doing double duty to track both depending on whether it is instrumentation for the pnetcdf file module or the pnetcdf var module?

shanedsnyder: That's a good point. I was thinking that maybe `ncid` was a more general thing like a `hid` is in HDF5 (a `hid` could reference an open file, dataset, property list, etc.), but looking closer at the PnetCDF interface docs it's not. An `ncid` (which is just an int) refers specifically to an open file it looks like, so I see how that could get confusing. I'll change the name here to fix that up, too.

shanedsnyder: They are not in the same hash. Looking closer at the code, I think it was just copying the HDF5 module where the same exact runtime structure could be used for files and datasets since everything is an `hid`. That works fine with the fix applied previously here, but it is confusing so I did try to fix. Maybe sanity check that  8538154 looks okay?

carns (approved):
Good deal, I like this organization with each module having it's own runtime struct with differentiated field names.  Should help avoid mixups in the future.

PR #873, opened 2022-12-08
bug fix: PnetCDF vard APIs when filetype is MPI_DATATYPE_NULL

wkliao:
When argument filetype is MPI_DATATYPE_NULL in vard APIs, it indicates this request is of zero length. Thus we can skip updating the counters.

Failed test program:
https://github.com/Parallel-NetCDF/PnetCDF/blob/master/test/testcases/test_vard_multiple.c

### DIFF
--- a/darshan-runtime/lib/darshan-pnetcdf-api.m4
+++ b/darshan-runtime/lib/darshan-pnetcdf-api.m4
@@ -136,8 +136,12 @@ define(`CALC_VARN_ACCESS_INFO',
 dnl
 define(`CALC_VARD_ACCESS_INFO',
     `int mpi_size;
-            PMPI_Type_size(filetype, &mpi_size);
-            $3 = mpi_size;')dnl
+            if (filetype == MPI_DATATYPE_NULL)
+                $3 = 0;
+            else {
+                MPI_Type_size(filetype, &mpi_size);
+                $3 = mpi_size;
+            }')dnl
 dnl
 define(`CALC_ACCESS_INFO',
     `ifelse(


shanedsnyder:
We've actually run into similar issues in the past with the MPI-IO module, not sure why it didn't occur to me to error check this.

Instead of bailing out of instrumentation, I think we should instrument as normal, just marking the access size as `0`. That matches what we do in MPI-IO, and would allow us to detect that applications are doing 0-byte operations. Could we instead add logic to `CALC_VARD_ACCESS_INFO` macro to just set access size to 0 in cases where the given datatpye is `NULL`?

wkliao:
revised as suggested

shanedsnyder:
Thanks, looks good from my local testing.

Quick question I noticed when sanity checking the output log for the test case that was failing: What would we expect the total number of bytes to be transferred from this portion of the test to be:

```
    /* bufcount is ignored when buftype is MPI_DATATYPE_NULL */
    err = ncmpi_put_vard_all(ncid, varid[0], MPI_INT, buf[0], -1,
                             MPI_DATATYPE_NULL); CHECK_ERR
```

0 bytes, as well, since the memory datatype is `MPI_DATATYPE_NULL`?

wkliao:
In this case, the write amount is 4 bytes, i.e. size of 3rd argument, filetype.
The buftype being MPI_DATATYPE_NULL means the write buffer is contiguous and its amount is equal to the size of filetype.

shanedsnyder:
Perfect, that matches what Darshan is instrumenting so all looks good.

Thanks, I'll go ahead and merge this.

shanedsnyder (approved):

PR #874, opened 2022-12-08
CI: test all PnetCDF test programs

wkliao:
Download PnetCDF release and run its `make check` and `make ptest` when Darshan library is pre-loaded.

### DIFF
--- a/.github/workflows/pnetcdf.yml
+++ b/.github/workflows/pnetcdf.yml
@@ -1,0 +1,113 @@
+name: PnetCDF
+
+on:
+  push:
+    branches: main
+  pull_request:
+    branches: main
+
+env:
+  PNETCDF_VERSION: 1.12.3
+
+jobs:
+    build:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v3
+        - name: Set up dependencies
+          run: |
+            sudo apt-get update
+            sudo apt-get install automake autoconf libtool libtool-bin m4
+            # mpi
+            sudo apt-get install mpich
+            # zlib
+            sudo apt-get install zlib1g-dev
+        - name: Install PnetCDF
+          run: |
+            echo "---- Install PnetCDF library into ${GITHUB_WORKSPACE}/PnetCDF"
+            cd ${GITHUB_WORKSPACE}
+            rm -rf PnetCDF
+            mkdir PnetCDF
+            cd PnetCDF
+            wget -cq https://parallel-netcdf.github.io/Release/pnetcdf-${PNETCDF_VERSION}.tar.gz
+            tar -zxf pnetcdf-${PNETCDF_VERSION}.tar.gz
+            cd pnetcdf-${PNETCDF_VERSION}
+            ./configure --prefix=${GITHUB_WORKSPACE}/PnetCDF \
+                        --silent \
+                        --enable-shared \
+                        MPICC=mpicc \
+                        MPICXX=mpicxx \
+                        MPIF77=mpifort \
+                        MPIF90=mpifort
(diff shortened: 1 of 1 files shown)


wkliao:
Note this CI failed due to a bug in #873
It should pass once the bug fix is merged.

.github/workflows/pnetcdf.yml
```diff
@@ -0,0 +1,113 @@
...
+            make ptest
+            unset LD_PRELOAD
+            make -s distclean >> qout 2>&1
+            rm -f qout
+        - name: Run Darshan make distcheck
+          if: ${{ success() }}
+          run: |
+            echo "make distcheck"
+            cd ${GITHUB_WORKSPACE}/build
+            make distcheck DISTCHECK_CONFIGURE_FLAGS="--with-log-path-by-env=DARSHAN_LOGPATH --with-jobid-env=NONE --enable-hdf5-mod --with-hdf5=${GITHUB_WORKSPACE}/HDF5 --enable-pnetcdf-mod --with-pnetcdf=${GITHUB_WORKSPACE}/PnetCDF RUNTIME_CC=mpicc"
```
shanedsnyder, line 107: I'm guessing earlier CI failures are masking this, but this configure line is specifying to enable HDF5 module (along with an HDF5 install path), but I don't see any HDF5 installation as part of this workflow. Should we just remove this?

.github/workflows/pnetcdf.yml
```diff
@@ -0,0 +1,113 @@
...
+            make -j 8
+            echo "-------- make check --------"
+            make check
+            echo "-------- make install --------"
+            make install
+        - name: Run PnetCDF make check
+          run: |
+            cd ${GITHUB_WORKSPACE}/PnetCDF/pnetcdf-${PNETCDF_VERSION}
+            make -s LIBTOOLFLAGS=--silent V=1 -j 8 tests > qout 2>&1
+            export LD_PRELOAD=${GITHUB_WORKSPACE}/darshan_install/lib/libdarshan.so
```
shanedsnyder, line 85: Can we add simple logic here to test for existence of expected number of Darshan logs for these tests or something? You can see in the workflow output that no log files are created because we never set the DARSHAN_LOGPATH env variable. I think the least we could do is assert that the expected number of logs was generated (and that the log path is correctly set so Darshan can write them).

.github/workflows/pnetcdf.yml
```diff
@@ -0,0 +1,113 @@
...
+          if: ${{ failure() }}
+          run: |
+            cd ${GITHUB_WORKSPACE}/PnetCDF/pnetcdf-${PNETCDF_VERSION}
+            cat src/utils/ncvalidator/*.log
+            cat test/*/*.log
+        - name: Run PnetCDF make ptest
+          run: |
+            cd ${GITHUB_WORKSPACE}/PnetCDF/pnetcdf-${PNETCDF_VERSION}
+            export LD_PRELOAD=${GITHUB_WORKSPACE}/darshan_install/lib/libdarshan.so
+            make ptest
```
shanedsnyder, line 98: Check log files are correctly generated here, too? Later on, we could add a script that checks more in-depth details as part of the workflow, but something simple is fine for now.

shanedsnyder (changes requested):

.github/workflows/pnetcdf.yml
wkliao: Setting DARSHAN_LOGPATH has been added.

How does Darshah check the number of log files?
Note the number of PnetCDF test programs can increase in new releases.

.github/workflows/pnetcdf.yml
wkliao: How do one check whether a Darshan log file is correct or not?

PR #877, opened 2022-12-16
MAINT: more Python 3.6 cleanups

### LINKED ISSUE #510: MAINT: when Python 3.6 is dropped

tylerjereddy:
Fixes #510

* in the absence of Python `3.6` support, we can safely use [PEP563](https://peps.python.org/pep-0563/) delayed type evaluations and use newer NumPy type hinting machinery; disable the old mocking shims related to this

* remove an `importlib_resources` requirement for Python `3.6`; note that for testing with Python < `3.9` you still want this third-party package I think, but clearly this conditional no longer serves any purpose since we don't support `3.6`

* remove Python `3.6` from the supported list in project metadata, and include `3.10` and `3.11`, both of which are regularly tested in our CI and have wheels available on PyPI

* adjust `cibuildwheel` settings to exclude Python `3.6` from wheel builds moving forward

* adjust `setup.py`/`pyproject.toml` to error out when attempting to build with Python < `3.7` (fail fast, early, and clearly..)

* we should no longer need `importlib_resources` in `summary.py` with 3.7+ guaranteed, so remove that import shim (we may also want to remove the import alias/remapping, but for now that should be harmless)

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -5,10 +5,7 @@ import base64
 import argparse
 import datetime
 from collections import OrderedDict
-if sys.version_info >= (3, 7):
-    import importlib.resources as importlib_resources
-else:
-    import importlib_resources
+import importlib.resources as importlib_resources
 
 from typing import Any, Union, Callable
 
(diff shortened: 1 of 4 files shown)


shanedsnyder (approved):

shanedsnyder:
Looks good!

PR #879, opened 2022-12-16
CI, MAINT: Cirrus bump/naming

tylerjereddy:
* currently our Cirrus CI job is called nothing more than `main`--let's see if we can name it something more helpful/informative for reviewers

* while I'm at it, let's see if I can use the latest MacOS M1 image from their docs:
https://cirrus-ci.org/guide/macOS/

* also, let's double check I can temporarily disable GitHub actions CI with a commit message tag in this project when we clearly don't need it like here

[skip actions]


shanedsnyder (approved):

shanedsnyder:
LGTM

will try to remember the "[skip actions]" thing, wasn't aware of that

PR #880, opened 2022-12-16
TST: portable forked process test

### LINKED ISSUE #828: MAINT, TST: portability of test_forked_process

tylerjereddy:
Fixes #828

* the forked process tests were failing locally with `act -j test_runtime`

* use a more portable approach to retrieving the absolute path to the system Python
executable (compared with assuming that `python` will do)

### DIFF
--- a/darshan-test/python_runtime_tests.py
+++ b/darshan-test/python_runtime_tests.py
@@ -1,3 +1,4 @@
+import sys
 import glob
 import subprocess
 import os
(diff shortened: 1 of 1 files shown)


shanedsnyder:
Notice that the other 2 test cases (`test_h5oopen_h5py_roundtrip` and `test_runtime_heatmap_div_by_zero`) reference `python` executable in the same way. Presumably those aren't tripping any errors because `mpirun` is invoking them, though I'm not sure I understand why?

Should those 2 be fixed as well?

tylerjereddy:
Probably better to change those other tests as well, but I haven't seen executable evidence of a problem. I can circle back to it, probably low priority for the other two.

shanedsnyder (approved):

shanedsnyder:
Sounds good. Probably not a big deal if they work in CI as well as with `act`, that covers all the use cases. I'll go ahead and approve, and leave to you to either update or go ahead and merge.

PR #882, opened 2022-12-17
propagate version string update

wkliao:
This simplifies the future version string update, by just editing file ./darshan.version.


tylerjereddy (dismissed):
The Python CI failures are real. FWIW, these changes make me a bit nervous from the Python build system standpoint anyway. I'd like to keep the Python build and packaging system pretty much uncoupled from what the C side is using, and indeed it would be much easier for some tooling, like benchmarking with `asv`, if the Python side could use a "Python ecosystem" solution (maybe [`meson`](https://github.com/mesonbuild/meson) at some point) to not only build the Python code but also the C code, and just link it in.

I don't think we want two separate build systems, but then we already kind of have two separate packages anyway (they're already two separate `spack` packages for example). That said, I guess I'm just suggesting that, if anything, we weaken the interdependencies between the two builds rather than making them more coupled since Python ecosystem folks definitely won't want to be messing around with autoconf/autotools ecosystem, etc.

tylerjereddy:
To be clear, a Python *developer* should be able to iterate on the Python code, and depend on its static version string, without caring about any of the autotools/autoconf stuff.

wkliao:
I can remove the 2nd commit which involves pydarshan.

From my experience on netcdf, I think making separate packages, i.e.
darshan runtime, darshan util, and pydarshan is a mistake, because
it allows users to mix different versions of the packages, i.e. use
darshan-runtime version 3.4.0 together with darshan-util 3.3.0.
There may be unexpected results.

configure.ac
```diff
@@ -5,7 +5,8 @@ dnl Process this file with autoconf to produce a configure script.
 dnl You may need to use autoheader as well if changing any DEFINEs
 
 AC_PREREQ([2.69])
-AC_INIT([Darshan],[3.4.2],[darshan-users@lists.mcs.anl.gov],
+AC_INIT([Darshan], m4_esyscmd_s([cat ./darshan.version]),
```
eli-schwartz, line 8:
```suggestion
AC_INIT([Darshan], m4_normalize(m4_include([./darshan.version])),
```

Avoids running /usr/bin/cat.

wkliao: Updated to use this better solution. Thanks.

shanedsnyder:
Changes look good to me, this will simplify releases a bit.

I agree with limiting any sort of dependencies between PyDarshan and the traditional Darshan build system -- we've made this a point in the past as it seems to give us needed flexibility to accomplish what we want on both ends. Ultimately it would be nice if we could automate how PyDarshan captures these versions (as well as C headers in `include`), but I'm not sure we need to do it by adopting Autotools in PyDarshan -- it doesn't seem to be used generally in Python projects.

shanedsnyder (approved):

tylerjereddy:
Ok, the Python stuff was removed, Shane approved, and you got a code review from an experienced build system developer (Eli) as well, so I merged it.

Incidentally, the CI only ran for Python on MacOS M1 native, I'm not actually sure what CI we'd want to flush for these changes normally.

shanedsnyder:
Re: CI tests, I added the `[skip actions]` key word on my commit since it just modified a doc. I guess that annotation doesn't affect the Cirrus stuff?

tylerjereddy:
Ah understood, yeah the services have different flags. Another project I work on summarizes some of them here, though I believe for some they added some custom handlers we could copy if we wanted
https://scipy.github.io/devdocs/dev/contributor/continuous_integration.html#skipping

PR #883, opened 2022-12-17
MAINT: I/O cost label spacing

tylerjereddy:
Fixes #881

* rotate the x axis tick labels by 90 degrees counterclockwise in the I/O cost plot to avoid
label overlap

* add regression test for x axis tick label positions/rotations/content, including for the
exact log in the matching issue and for one other log

Samples of adjusted plots:

- `shane_ior-PNETCDF_id438100-438100_11-9-41525-10280033558448664385_1.darshan`
image

- `imbalanced-io.darshan`
image

PR #884, opened 2022-12-17
ENH: WAIT_TIME in Python summary

### LINKED ISSUE #860: ENH: add `WAIT_TIME` to PnetCDF I/O cost plots
The PNETCDF_FILE module is unique compared to other modules currently supported by our I/O cost plots in PyDarshan, in that it has a new I/O cost category called `WAIT_TIME`. In PnetCDF, wait time represents the amount of time applications spent waiting on the completion of asynchronous I/O operations -- it's important to reflect this cost in total I/O cost for completeness. We should add the ability to include this wait time in the generated I/O cost plots for PnetCDF.

tylerjereddy:
Fixes #860

* include `PNETCDF_FILE_F_WAIT_TIME` in the I/O cost plot in the Python summary reports, along with regression testing and caption changes

* I would have felt more comfortable if the issue pointed me to a log file with prominent non-zero wait times, because all-zero values will also be the default for the `Wait` category when it isn't valid (even when `PNETCDF` is not used); I mostly decided not to exclude the category when `PNETCDF` is absent because it was easier to code + accounted for in the caption adjustment anyway

* I'll provide some samples of the new I/O cost plots--note that the x label squishing is handled separately in gh-883.

- `shane_ior-PNETCDF_id438100-438100_11-9-41525-10280033558448664385_1.darshan`
image

- `imbalanced-io.darshan`
image

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
@@ -30,7 +30,7 @@ def get_by_avg_series(df: Any, mod_key: str, nprocs: int) -> Any:
     Returns
     -------
     by_avg_series: a ``pd.Series`` containing the
-    average read, write, and meta times.
+    average read, write, meta, and wait times.
 
     """
     # filter out all except the following columns
@@ -39,11 +39,16 @@ def get_by_avg_series(df: Any, mod_key: str, nprocs: int) -> Any:
         f"{mod_key}_F_WRITE_TIME",
         f"{mod_key}_F_META_TIME",
     ]
+    if "PNETCDF_FILE" in mod_key:
+        cols.append("PNETCDF_FILE_F_WAIT_TIME")
+    else:
+        cols.append("Wait")
     by_avg_series = df.filter(cols, axis=1).sum(axis=0) / nprocs
     # reindex to ensure 3 rows are always created
     by_avg_series = by_avg_series.reindex(cols, fill_value=0.0)
     # rename the columns so the labels are automatically generated when plotting
     name_dict = {cols[0]: "Read", cols[1]: "Write", cols[2]: "Meta"}
+    name_dict[cols[3]] = "Wait"
     by_avg_series.rename(index=name_dict, inplace=True)
     return by_avg_series
 
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -405,7 +405,8 @@ class ReportData:
             "Average (across all ranks) amount of run time that each process "
             "spent performing I/O, broken down by access type. See the right "
             "edge bar graph on heat maps in preceding section to indicate if "
-            "I/O activity was balanced across processes."
+            "I/O activity was balanced across processes. The 'Wait' category "
+            "is only meaningful for PNETCDF asynchronous I/O operations."
         )
         io_cost_params = {
             "section_title": "Cross-Module Comparisons",
--- a/darshan-util/pydarshan/darshan/tests/test_plot_io_cost.py
+++ b/darshan-util/pydarshan/darshan/tests/test_plot_io_cost.py
@@ -21,37 +21,37 @@ from darshan.experimental.plots.plot_io_cost import (
             "ior_hdf5_example.darshan",
             pd.DataFrame(
                 np.array([
-                    [0.0196126699, 0.1342029571533203, 0.0074423551],
-                    [0.0196372866, 0.13425052165985107, 0.0475],
-                    [0.016869, 0.086689, 0.097160],
-                    [0.0, 2.5570392608642578e-05, 0.0],
+                    [0.0196126699, 0.1342029571533203, 0.0074423551, 0.0],
+                    [0.0196372866, 0.13425052165985107, 0.0475, 0.0],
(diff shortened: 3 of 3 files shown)


darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
```diff
@@ -39,11 +39,16 @@ def get_by_avg_series(df: Any, mod_key: str, nprocs: int) -> Any:
         f"{mod_key}_F_WRITE_TIME",
         f"{mod_key}_F_META_TIME",
     ]
+    if "PNETCDF_FILE" in mod_key:
+        cols.append("PNETCDF_FILE_F_WAIT_TIME")
+    else:
+        cols.append("Wait")
```
shanedsnyder, line 45: Is the else statement here just needed to ensure `cols` has the right size or something? Just double checking that's as expected here, since the column name is not actually a counter.

tylerjereddy: It just makes a column of all-zero values yeah. What worried me a bit, as suggested above, is that the test file had zeros there anyway, so programmer paranoia about testing the genuine code path.

shanedsnyder:
Looks good, thanks!

We can stick with displaying the `WAIT` component in all cases for now, even though it's only used for PnetCDF module. I started to push the other direction, but we don't suppress any of the other components based on contents of the log (i.e, we don't suppress WRITE component for logs that are read-only). I won't make it harder here. It's possible we might more generally tweak how the I/O cost plots are presented now that they cover more use cases, but that's for another day I think.

As for not having any logs that demonstrate the new counter, that's true. The E3SM-IO benchmark is a good example of which I have a larger log example we could use. We already have a couple of older E3SM-IO examples in the darshan-logs repo, but those predate the new module. We might consider just replacing with a newer version if it's not too disruptive -- I'm not sure we want to add a ton of variations of the same workloads unless it's needed for specific regressions, just to keep darshan-logs from exploding in terms of log count (there's probably a longer term maintenance cost without keeping an eye on that, but obviously a short term maintenance cost to update existing tests).

At any rate, here's I/O cost for a newer E3SM-IO test case generated locally for me on this branch:
image

darshan-util/pydarshan/darshan/experimental/plots/plot_io_cost.py
shanedsnyder: Sounds good, just wanted to check.

shanedsnyder (approved):

shanedsnyder:
Feel free to merge unless anything else you want to discuss in the mean time.

wkliao:
Just curious, how come there is no POSIX time in the chart?

shanedsnyder:
> Just curious, how come there is no POSIX time in the chart?

I think what's happening is an artifact of collective I/O algorithm and the fact that this I/O cost is calculated in terms of _average_ cost per process. Only the aggregator processes (1 of every 32 processes is an aggregator here, probably one per Crusher compute node) are doing most of the I/O for this workload, so many processes are left idle which causes the average I/O cost to be calculated as really low. We have plans to address this at some point to add new counters that allow us to calculate a worst-case I/O cost metric that is more useful for quantifying the cost of collective I/O measured at the POSIX level.

wkliao:
That makes sense.

As E3SM's I/O pattern consists of lots of small, noncontiguous requests from each process, it is expected that the communication cost of the two-phase I/O dominates the MPI-IO time. I wonder what kind of plot can reveal that information. Max or Avg, or both.

One question about your run of E3SM-IO. Which case was this run? The number of processes, and Lustre striping info?

shanedsnyder:
Here's the details of the job from darshan-parser's header:
image

So, 1K processes running that larger scale F-case file you shared with me. Note that this is targeting GPFS -- I'm not sure if there's any Lustre scratch on Crusher. Our project space is on a GPFS volume.

> As E3SM's I/O pattern consists of lots of small, noncontiguous requests from each process, it is expected that the communication cost of the two-phase I/O dominates the MPI-IO time. I wonder what kind of plot can reveal that information. Max or Avg, or both.

I'm not sure. It's possible the AutoPerf APMPI module could help quantify the cost of communication a bit, but it captures information across the entire communicator, it can't characterize communication related specifically to collective I/O for a particular file.

Off the top of my head, it seems that comparing the average MPI-IO timing with the worst-case POSIX timing could give you some sort of indication of communication costs for collective I/O. The average for MPI-IO gives you the total cost of the collective I/O algorithm, while the worst-case POSIX cost gives you an idea of how much was spent doing I/O (or idling on non-aggregators). The difference is probably a fair estimate of the comm. costs.

wkliao:
For collective I/O, the slowest process represents the cost. So, probably Max for both MPI-IO and POSIX? (At least for E3SM-IO's PnetCDF case, as it only  call collectives.)

Direct measurement of MPI communication cost is hard, as there are many nonblocking communication calls in ROMIO. We usually subtract the max of collective I/O call from the max POSIX I/O call to present the communication.

shanedsnyder:
Yeah, that all makes sense. It's probably easier to use max for MPI-IO too like you mention. Given the synchronization typical in collectives, I assume there isn't much deviation at the MPI-IO level in terms of timing.

tylerjereddy:
There's an approval to merge and CI is passing so I'll put it in. The discussion above seems focused on understanding an example, but doesn't seem to present a request for additional changes.

PR #885, opened 2022-12-21
add darshan.version to distribution

wkliao:
Forgot to add this new file in #882


wkliao:
It turns out autoconf automatically includes that file as a configure dependency.
So, there is no need to add this file in EXTRA_DIST explicitly.

PR #886, opened 2023-01-05
WIP, ENH: add df->rec converter

tylerjereddy:
* add a per-record converter from DataFrame to darshan-util C record format that passes tests for single records (rows) with `counters` and `fcounters` (but not `rank` & `id` yet)

* add a full df->multi-record converter; this one still leads to segfaults with the C inject func with stuff like (probably corrupted data..):
```
../darshan-logutils-accumulator.c:145: darshan_accumulator_inject: Assertion `rank < acc->job_nprocs' failed.`
```

* one of the two added tests passes, the other xfails, feel free to push in fixes

* as discussed in meeting earlier today (Rob R. seems to agree), this likely shouldn't be a long-term solution--just temporary until we can hoist all analysis infrastructure to the Python level

* some of the accumulator API prototypes are copied in here from other PRs for testing purposes only.. since the main purpose of doing something like this would be to allow filtering DataFrames and then passing the raw records to the C accum interface.. (well, you also avoid re-reading the log file records)

* I'll leave backwards compatibility and support beyond STDIO/POSIX drafted here to Argonne folks most likely

### DIFF
--- a/darshan-util/pydarshan/darshan/backend/cffi_backend.py
+++ b/darshan-util/pydarshan/darshan/backend/cffi_backend.py
@@ -679,10 +679,7 @@ def _df_to_rec(rec_dict, mod_name, rec_index_of_interest=None):
 
     Returns
     -------
-    buf: CFFI data object containing a buffer of record(s) or a single
-          record; when multiple records are present this will effectively
-          be a raw char array; with a single record, it should be cast
-          to a struct of the appropriate type
+    buf: Raw char array containing a buffer of record(s) or a single record.
     """
     counters_df = rec_dict["counters"]
     fcounters_df = rec_dict["fcounters"]
--- a/darshan-util/pydarshan/darshan/backend/cffi_backend.py
+++ b/darshan-util/pydarshan/darshan/backend/cffi_backend.py
@@ -684,7 +684,6 @@ def _df_to_rec(rec_dict, mod_name, rec_index_of_interest=None):
           be a raw char array; with a single record, it should be cast
           to a struct of the appropriate type
     """
-    mod_type = _structdefs[mod_name]
     counters_df = rec_dict["counters"]
     fcounters_df = rec_dict["fcounters"]
     counters_n_cols = counters_df.shape[1]
--- a/darshan-util/pydarshan/darshan/backend/cffi_backend.py
+++ b/darshan-util/pydarshan/darshan/backend/cffi_backend.py
@@ -657,3 +657,64 @@ def _log_get_heatmap_record(log):
     libdutil.darshan_free(buf[0])
     
     return rec
+
+
+def _df_to_rec(rec_dict, mod_name, rec_index_of_interest=None):
+    mod_type = _structdefs[mod_name]
+    counters_n_cols = rec_dict["counters"].shape[1]
+    fcounters_n_cols = rec_dict["fcounters"].shape[1]
+    size = (counters_n_cols +
+            fcounters_n_cols -
+            # id and rank columns are duplicated
+            # in counters and fcounters
+            2)
+    rbuf_orig = np.empty(shape=(size,), dtype=np.float64)
+    rbuf = ffi.cast(mod_type, ffi.from_buffer(rbuf_orig))
+    rbuf[0] = ffi.from_buffer(rbuf_orig)
+    for counter_name, dtype, cols in zip(["counters", "fcounters"],
+                                         [np.int64, np.float64],
+                                         [counters_n_cols, fcounters_n_cols]):
+        df = rec_dict[counter_name]
+        record = df.iloc[rec_index_of_interest, ...]
+        data = np.ascontiguousarray(record[:cols - 2].to_numpy(), dtype=dtype)
+        if counter_name == "fcounters":
+            rbuf[0].fcounters = data.tolist()
+        else:
+            rbuf[0].counters = data.tolist()
+    #rbuf[0].base_rec.rank = np.int64(record.loc["rank"])
+    #rbuf[0].base_rec.id = np.uint64(record.id)
+    return rbuf
+
+
+def _df_to_rec_array(rec_dict, mod_name):
+    # same _df_to_rec, but first allocate a contiguous
+    # buffer into which we can add ALL the records
+    # in the pandas data structures
+    # TODO: looping like this in Python/pandas is
+    # inefficient for many reasons--we really should
+    # be doing the data analysis on the DataFrame
+    # long term rather than moving data backwards
+    # like this to the C layer
+    counters_df = rec_dict["counters"]
+    fcounters_df = rec_dict["fcounters"]
+    num_recs_counters = counters_df.shape[0]
+    num_recs_fcounters = fcounters_df.shape[0]
+    size = (counters_df.shape[1] +
+            fcounters_df.shape[1] -
+            # id and rank columns are duplicated
+            # in counters and fcounters
+            2)
+    if num_recs_counters != num_recs_fcounters:
+        raise ValueError(f"{num_recs_counters} counter records "
+                         f"but {num_recs_fcounters} fcounter records")
+    record_array = np.empty(shape=size * num_recs_counters, dtype=np.float64)
+    # NOTE: we may be able to slurp the data in as a single
+    # buffer rather than iterating, but moving data from Python back to C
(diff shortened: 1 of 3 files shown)


darshan-util/pydarshan/darshan/backend/cffi_backend.py
```diff
@@ -657,3 +657,64 @@ def _log_get_heatmap_record(log):
...
+                                         [counters_n_cols, fcounters_n_cols]):
+        df = rec_dict[counter_name]
+        record = df.iloc[rec_index_of_interest, ...]
+        data = np.ascontiguousarray(record[:cols - 2].to_numpy(), dtype=dtype)
+        if counter_name == "fcounters":
+            rbuf[0].fcounters = data.tolist()
+        else:
+            rbuf[0].counters = data.tolist()
+    #rbuf[0].base_rec.rank = np.int64(record.loc["rank"])
+    #rbuf[0].base_rec.id = np.uint64(record.id)
```
tylerjereddy, line 685: a careful review of the buffer memory layouts used here/elsewhere in this branch is probably the main thing needed to move this forward

tylerjereddy: one thing we could do in testing is probably check the raw bytes--something like `np.frombuffer(ffi.buffer(rbuf[0])).tobytes()` vs. the original read-in bytes for a given record.

tylerjereddy:
This is much closer now--`test_reverse_record_array` even shows the C accumulator machinery operating on a pre-filtered dataframe that has been converted into a buffer of records, and retrieves somewhat-reasonable file count values locally from the derived metrics. If I don't pre-filter I do see the same total files as described in gh-867 (`1026` total files), so that's a good sign...

There's a memory access/`free`-related issue caused by something in `test_reverse_record_array` -- feel free to give me a hand with that.. otherwise pretty close I think...

darshan-util/pydarshan/darshan/tests/test_cffi_misc.py
```diff
@@ -159,3 +160,107 @@ def test_log_get_generic_record(dtype):
...
+    expected_ro_files = 13 # see gh-867
+    expected_wo_files = 12 # see gh-867
+    actual_total_files = darshan_derived_metrics.category_counters[0].count
+    actual_ro_files = darshan_derived_metrics.category_counters[1].count
+    actual_wo_files = darshan_derived_metrics.category_counters[2].count
+    libdutil.darshan_free(buf[0])
+    backend.log_close(log_handle)
+    assert actual_total_files == expected_total_files
+    assert actual_ro_files == expected_ro_files
+    assert actual_wo_files == expected_wo_files
```
tylerjereddy, line 266: To give an idea of how close we are, the test passes locally with no segfault issues with this diff to match `darshan-parser --show-incomplete --file` from gh-867 (but I wanted to demo the filtering to remove files that were only stat'd). Something is still sensitive when I run the entire `darshan` suite though.. probably memory management.

```diff
--- a/darshan-util/pydarshan/darshan/tests/test_cffi_misc.py
+++ b/darshan-util/pydarshan/darshan/tests/test_cffi_misc.py
@@ -220,8 +220,8 @@ def test_reverse_record_array():
     # here at Python layer, then feed back to C accum stuff
     fcounters_df = fcounters_df[counters_df["POSIX_OPENS"] > 0]
     counters_df = counters_df[counters_df["POSIX_OPENS"] > 0]
-    rec_dict["counters"] = counters_df
-    rec_dict["fcounters"] = fcounters_df
+    #rec_dict["counters"] = counters_df
+    #rec_dict["fcounters"] = fcounters_df
     record_array = backend._df_to_rec(rec_dict, "POSIX")
     num_recs = rec_dict["fcounters"].shape[0]
     mod_type = backend._structdefs["POSIX"]
@@ -253,9 +253,9 @@ def test_reverse_record_array():
     # 1 = RO
     # 2 = WO
     # 3 = R/W
-    expected_total_files = 28 # see gh-867
-    expected_ro_files = 13 # see gh-867
-    expected_wo_files = 12 # see gh-867
+    expected_total_files = 1026 # see gh-867
+    expected_ro_files = 12 # see gh-867
+    expected_wo_files = 2 # see gh-867
     actual_total_files = darshan_derived_metrics.category_counters[0].count
     actual_ro_files = darshan_derived_metrics.category_counters[1].count
     actual_wo_files = darshan_derived_metrics.category_counters[2].count
```

tylerjereddy:
Another thing I thought of--it may be possible to simplify the complexity in `test_reverse_record_array` if we could replace stuff like `darshan_accumulator_create()` with an equivalent struct initialization on the CFFI/Python/NumPy recarray side. Whether it is worth it depends on how complex the logic is, and how annoying it is to do portably in Python across the different modules.

shanedsnyder:
I still need to review the tests, but the conversion code looks good to me. I don't have any suggestions yet, but maybe I will after I work through some of the issues you mention for reviewers. I'm not sure it's a big deal, but conceptually I would think you could always return a buffer of the underlying record type, as opposed to a `char[]` as you do in the multi-record case -- I wonder if that has something to do with the warning you mention.

> determine if there any modules that are supported by the accumulator machinery that wouldn't automatically work with the infrastructure that is working with STDIO and POSIX here--for example, LUSTRE may have variable-length fields that require additional shims, but I don't think LUSTRE is supported for accumulation anyway. Related to this--more relevant test cases you'd like to see.

Luckily, all of the modules that would ultimately support accumulation follow the same structure as what you already have implemented for POSIX/STDIO with little tweaks. Lustre is one that won't support that. The only issue I can think of is that the `H5D` and `PNETCDF_VAR` modules have an additional field in the structure that will need to be packed, but that's probably not a big hassle.

shanedsnyder:
> help me figure out if we should be worried about the mismatch in accumulator file count values for the stat-filter scenario (see the one xfailed test case in this PR and file category count mismatch between darshan-parser and darshan-job-summary.pl #867 discussion); that will take some digging to figure out I suspect, but since we match darshan-parser perfectly when we don't filter the stat'd files, I'm not so sure we have any problem on the Python side here; still, it would be good to get an explanation for the filtered case

I just pushed a fix for this and a couple of other smaller things for this test. The short story is that these derived metrics presented in the old Perl report are actually summed across both POSIX and STDIO modules, so we needed to account for that difference in this test since it's POSIX only. I modified the expected values to reflect the POSIX module only values -- I started to add STDIO support for the test, but it was getting messy and I'm not sure there's much value in us testing the expected values for combining modules.

For a little more background, in the past we have often combined POSIX/STDIO data when doing some analyses, since they are both found at the lowest layer of the I/O stack (that we can instrument) and they don't layer on top of each other (we don't have to worry about double counting statistics if the APIs don't layer on top of each other). If you recall, we had a similar issue with the "file access by category" plots, where we needed to update them to additionally include STDIO data. That said, I'm fine with our current approach of providing these stats on a per-module basis in the new PyDarshan job summary tool, so don't think there's much to follow up on here.

shanedsnyder:
I just pushed another commit that, among other things, addresses the warning you mentioned related to an implicit cast.

I think it simplifies `_df_to_rec()` a little bit as an added bonus, which now always returns a bytes object -- previously it was returning different things depending on whether you gave it one or multiple records.

The tests have been updated a bit, too, and are passing for me locally:
- `test_df_to_rec()` now casts the returned bytes object into appropriate record pointer type it can use to sanity check specific fields
- `test_reverse_record_array()` was not handling one of the arguments to `accumulator_emit()` routine properly, so I fixed that, though we aren't using this value yet

tylerjereddy:
I checked all 4 of your commits and they look like clear improvements to me. I suppose you could also update the return value docstring for `_df_to_rec` now that it is simply bytes, but it is a `_private` function anyway.

I guess some decision will have to be made on if you proceed with this PR first, then modify some of the older aggregation ones to use this machinery instead of the Python hand-feeding records to the accumulator approach from before.

tylerjereddy:
`mypy 1.0.0` was released two days ago--maybe just pin to `mypy<1.0.0` to silence CI for now (probably easy to do the proper fix as well.. but not a priority).

shanedsnyder (approved):

shanedsnyder:
I think all looks good here. Not an ideal solution, but our hands are tied until we can move more stuff into Python. In the mean time, this will let us use the existing C aggregation stuff while giving us flexibility (i.e., for filtering) using pandas.

PR #890, opened 2023-01-23
BLD: fix editable installs

### LINKED ISSUE #887: BLD, TST: editable installs broken
Only really affects developers, but this used to work for us. It may be that i.e., PEP standard adoption in build systems/`pip` has changed/updated a bit and we may need to move into conformance with that.

To reproduce on latest `main` (may need recent versions of `pip`, etc.):

- build/install `darshan-util`
- do an editable install of `pydarshan` with: `python -m pip install -e .`
- try running the tests from another path like `/tmp`: `python -m pytest --pyargs darshan`
  - won't work: `ERROR: module or package not found: darshan (missing __init__.py?)`

This would be useful to have working again because it allows rapid iteration when no compiled code changes are made.

tylerjereddy:
Fixes #887

* this allows `pip install -e .` to produce a working editable install locally

* the basis for the `setuptools` pin here is based on reading through https://github.com/pypa/setuptools/issues/3548; in short, looks like a behavior change in `setuptools` so let's just pin it for now

* since this only affects Python **developers** for our project I've not added any "regression test"/"CI test" for it, but feel free to add one editable job to the CI if you want

### DIFF
--- a/darshan-util/pydarshan/pyproject.toml
+++ b/darshan-util/pydarshan/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
     "wheel",
-    "setuptools",
+    "setuptools<64.0.0",
 ]
 
 [project]


shanedsnyder:
LGTM. I guess we'll circle back at some point and either address whatever's needed with our setup to get this working again, or maybe it'll be automagically fixed in future setuptools releases.

shanedsnyder (approved):

Issue #895, opened 2023-02-02
CI: failure in runtime job with h5py


tylerjereddy:
Example CI failure in gh-894 for the runtime job to build `h5py` from source with MPI support: https://github.com/darshan-hpc/darshan/actions/runs/4078783404/jobs/7029407901

Snippet of the error is below. Since `h5py` release went up on PyPI on January 23/2023, it may be worth a temporary pin to circumvent for now...

```
Collecting h5py
  Downloading h5py-3.8.0.tar.gz (400 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 400.8/400.8 kB 6.5 MB/s eta 0:00:00
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Installing backend dependencies: started
  Installing backend dependencies: finished with status 'error'
  error: subprocess-exited-with-error
  
  × pip subprocess to install backend dependencies did not run successfully.
  │ exit code: 1
  ╰─> [26 lines of output]
      DEPRECATION: --no-binary currently disables reading from the cache of locally built wheels. In the future --no-binary will not influence the wheel cache. pip 23.1 will enforce this behaviour change. A possible replacement is to use the --no-cache-dir option. You can use the flag --use-feature=no-binary-enable-wheel-cache to test the upcoming behaviour. Discussion can be found at https://github.com/pypa/pip/issues/11453
      Collecting wheel
        Using cached wheel-0.38.4-py3-none-any.whl (36 kB)
      Collecting mpi4py==3.1.0
        Downloading mpi4py-3.1.0.tar.gz (2.4 MB)
           ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.4/2.4 MB 25.0 MB/s eta 0:00:00
        Installing build dependencies: started
        Installing build dependencies: finished with status 'done'
        Getting requirements to build wheel: started
        Getting requirements to build wheel: finished with status 'error'
        error: subprocess-exited-with-error
      
        × Getting requirements to build wheel did not run successfully.
        │ exit code: 1
        ╰─> [1 lines of output]
            error in mpi4py setup command: 'python_requires' must be a string containing valid version specifiers; Invalid specifier: '!=3.4.*''
            [end of output]
```

tylerjereddy:
[...]

PR #896, opened 2023-02-02
CI: setuptools pin for runtime job

tylerjereddy:
* try to circumvent gh-895 for now

### DIFF
--- a/.github/workflows/runtime_ci.yml
+++ b/.github/workflows/runtime_ci.yml
@@ -29,7 +29,8 @@ jobs:
           python -m pip install --upgrade pytest mpi4py
           # we need to build h5py with the system HDF5 lib backend
           export HDF5_MPI="ON"
-          CC=mpicc python -m pip install --no-binary=h5py h5py
+          # h5py is pinned to avoid gh-895
+          CC=mpicc python -m pip install --no-binary=h5py "h5py==3.7.0"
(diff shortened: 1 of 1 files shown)


tylerjereddy:
Revised to instead pin `pip` version for the runtime job--it looks likely that enforcement of new build system rules was deployed that `mpi4py` is not yet compliant with.

tylerjereddy:
Ah, that doesn't work either, so it is perhaps the latest version of `setuptools` (Feb 1 new release) that needs to be pinned, though that's trickier to deal with as a dependency that may be specified indirectly through other deps.

tylerjereddy:
I think the `setuptools` pin/fix here is more appropriate given the upstream discussion, and seems to work better locally with `act`, but now GHA is complaining about something unrelated: `E: Failed to fetch https://packages.microsoft.com/ubuntu/22.04/prod/dists/jammy/main/binary-amd64/Packages.gz  Hash Sum mismatch.`

Maybe best to try that later or tomorrow then...

tylerjereddy:
Alright, the CI is passing now. It isn't exactly pretty, but you can see other projects adopting a similar solution in the linked upstream issue.

shanedsnyder (approved):

shanedsnyder:
Sounds good, thanks for tracking that down!

PR #897, opened 2023-02-08
CI: pin to mpy<1.0.0

shanedsnyder:
Address new CI failures popping up in #886 related to a recent release of mypy by pinning for now.


tylerjereddy (approved):
LGTM

PR #898, opened 2023-02-10
updated derived metrics interface

shanedsnyder:
Mostly for discussion for now, see commits for more details.

One thing I'd like to cleanup is add some sort of helper logic for mapping between module names and their C library module identifiers (integers). It's annoying to maintain that in Python, but otherwise we don't really have anyway to use the accumulator API unless we have an open Darshan log we can use to find the underlying module identifier. For now, the Python accumulator interface is ugly since it requires you to give the module name and ID.


darshan-util/pydarshan/darshan/backend/cffi_backend.py
```diff
@@ -701,10 +703,60 @@ def _df_to_rec(rec_dict, mod_name, rec_index_of_interest=None):
...
     else:
-        rec_arr.id = counters_df.iloc[rec_index_of_interest, 0]
-        rec_arr.rank = counters_df.iloc[rec_index_of_interest, 1]
+        rec_arr.id = counters_df.iloc[rec_index_of_interest, id_col]
+        rec_arr.rank = counters_df.iloc[rec_index_of_interest, rank_col]
     buf = rec_arr.tobytes()
     return buf
+
+
+def log_get_derived_metrics(rec_dict, mod_name, mod_idx, nprocs):
```
tylerjereddy, line 715: I suppose in the future we could allow an optional kwarg for custom filtering operations on the DataFrames, which then gets passed to a similar optional kwarg for `_df_to_rec()`.

For now, I think we're focused on just getting the standard accum elements into the summary report, but I'm just thinking ahead to make sure we don't lock ourselves into a design that is harder to apply custom DataFrame filters to.

I suppose, however, that for supporting arbitrary slicing and dicing of DataFrames we may want to just allow folks to manually operate on the two DataFames in `rec_dict` and then pass it in here, since a pass-through API that can achieve *any* kind of filtering op may be more work than it is worth.

Perhaps some pre-canned filtering ops will be built-in for us, like an argument for a regex to filter on filenames or something.

darshan-util/pydarshan/darshan/backend/cffi_backend.py
```diff
@@ -701,10 +703,60 @@ def _df_to_rec(rec_dict, mod_name, rec_index_of_interest=None):
...
+                           "darshan_accumulator_inject() at the C level. "
+                           "It may be possible "
+                           "to retrieve additional information from the stderr "
+                           "stream.")
+    derived_metrics = ffi.new("struct darshan_derived_metrics *")
+    total_record = ffi.new(_structdefs[mod_name].replace("**", "*"))
+    r = libdutil.darshan_accumulator_emit(darshan_accumulator[0],
+                                          derived_metrics,
+                                          total_record)
+    libdutil.darshan_accumulator_destroy(darshan_accumulator[0])
```
tylerjereddy, line 755: I suppose we could check for a `0` retcode here, though the function appears to have no error handling anyway, not even in the form of non-zero exit code.

darshan-util/pydarshan/darshan/backend/cffi_backend.py
```diff
@@ -701,10 +703,60 @@ def _df_to_rec(rec_dict, mod_name, rec_index_of_interest=None):
...
+    record_array = _df_to_rec(rec_dict, mod_name)
+
+    r_i = libdutil.darshan_accumulator_inject(darshan_accumulator[0], record_array, num_recs)
+    if r_i != 0:
+        raise RuntimeError("A nonzero exit code was received from "
+                           "darshan_accumulator_inject() at the C level. "
+                           "It may be possible "
+                           "to retrieve additional information from the stderr "
+                           "stream.")
+    derived_metrics = ffi.new("struct darshan_derived_metrics *")
```
tylerjereddy, line 750: The CFFI stuff is of course a bit awkward, especially given how many different entrypoints to the C layer we need to interact with just to get a single final struct back, but I'd probably be inclined to just live with it if we just operate on DataFrames or Arrow in-memory stuff in a few years anyway.

It occurs to me that for some things, like `struct darshan_derived_metrics`, we could probably avoid CFFI traversal by using the bytes from a NumPy `recarray`, but this would be more verbose with minimal gain for now I suspect.

tylerjereddy (commented):
I think I'm fine with most of it, but then again a lot of this is effectively improved versions of my original rough code so I may be biased.

The blue text bandwidth strings are still showing up in the Python summary reports for me on this branch, which is a good sign.

darshan-util/pydarshan/darshan/backend/cffi_backend.py
shanedsnyder: I did think about this briefly and just assumed we'd expect users to do the filtering directly on the dataframes before having PyDarshan operate on them, similar to your approach in the `test_reverse_record_array` test case. I'd definitely be open to more directly supporting in our APIs if we think it'll avoid some complexity and simplify things for users. I guess users have to filter counters/fcounters separately, so avoiding that and instead passing in some kwargs to handle internally could be useful if it's not a pain to get working for different use cases.

darshan-util/pydarshan/darshan/backend/cffi_backend.py
shanedsnyder: Yeah, I just punted on error checking that for a couple of reasons:
- Internal knowledge that the `destroy()` call doesn't actually have an error path despite having a return code -- it's basically just freeing some stuff, so if there's an error it's likely going to be a crash.
- We want to call `destroy()` before potentially raising an exception related to the `emit()` call returning a failure just to make sure we clean up C memory. It just becomes a little complicated/verbose to unwind the errors from back-to-back calls like that, so I just opted for simplicity with the internal knowledge from above.

darshan-util/pydarshan/darshan/tests/test_lib_accum.py
```diff
@@ -0,0 +1,96 @@
...
+        derived_metrics = log_get_derived_metrics(rec_dict, mod_name, mod_idx, nprocs)
+        actual_str = log_get_bytes_bandwidth(derived_metrics=derived_metrics,
+                                             mod_name=mod_name)
+        assert actual_str == expected_str
+
+#    elif expected_str == "ValueError":
+#        with pytest.raises(ValueError,
+#                           match=f"{mod_name} is not in the available log"):
+#            log_get_bytes_bandwidth(derived_metrics=derived_metrics,
+#                                    mod_name=mod_name)
```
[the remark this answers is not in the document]
shanedsnyder: Thanks, I add this patch with [5d9e362](https://github.com/darshan-hpc/darshan/pull/898/commits/5d9e36283b191eee3049767fa2a68bda92f5c4d5) and looks to be working fine.

shanedsnyder:
I just pushed one more commit that adds some helper for mapping between module name to index and uses this logic internally in `log_get_derived_metrics()` so that callers don't have to ensure they hold onto both module names and log format indices. Really, we should probably not require users to know what a module index is and have them interact entirely with module names for simplicity.

I think I'm satisfied with the state of things here and am willing to merge things in assuming there are no other critical concerns to address? I'm not sure if it makes any difference for attribution purposes, but I could cherry-pick these commits back onto your original PR @tylerjereddy if it matters any to you? Otherwise, I just requested a review here and we can merge this in directly and close #839

tylerjereddy:
The last two commits look fine to me.

I suppose we could debate about using a dictionary to retrieve the `mod_idx` to get `O(1)` vs. the `O(n)` search from `list.index()`, but the sequence is so small, and unlikely to balloon, that it isn't worth the effort either way.

Feel free to merge this PR if that is easiest, and if you want to squash the commits or whatever I don't mind.

PR #900, opened 2023-02-14
MAINT: unpin mypy

tylerjereddy:
* `mypy` `1.0.0` was released on Feb. 6/2023, and we recently pinned it down to older versions while working on more important things because we had a few new type hinting issues pop up

* fix the type hinting issues, which were actually just ignore directives on `try/except` blocks that are no longer needed, so that's a clear win I think

* I doubt it matters much for us, but `mypy 1.0.0` can be ~40 % faster on large codebases:
https://mypy-lang.blogspot.com/2023/02/mypy-10-released.html

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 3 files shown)


tylerjereddy:
Ah, confusingly the issue does still exist on `mypy 1.0.0`: https://github.com/python/mypy/issues/13914#issuecomment-1420406716

It just didn't show up when I tested with Python `3.10` locally, which uses a different code path.

Issue #906, opened 2023-02-25
Incorrect Timing of File Close when Using NetCDF4


yzanhua:
# Summary
NetCDF4 can perform parallel IO using parallel HDF5. When using Darshan to capture a NetCDF4 application's I/O behavior, I observe that the actual file close is delayed from `nc_close` call to `MPI_Finalize`. The incorrect timing of file close will affect the correctness of [Log VOL](https://github.com/DataLib-ECP/vol-log-based) who needs to use/release HDF5 resources at file close time, some of which are not available at `MPI_Finalize` (e.g. H5T_STD_B8LE).

### Test program
`test.c` is a simple NetCDF4 programs that open a NetCDF4 file and close directly. It also prints a string `application: nc_close start` and `application: nc_close end` before and after `nc_close`.

<details> <summary>Click here to see test.c</summary>

```c
#include <stdio.h>
#include <stdlib.h>
#include <mpi.h>
#include <netcdf.h>
#include <netcdf_par.h>

#define FATAL_ERR {if(err!=NC_NOERR) {printf("Error at line=%d: %s Aborting ...\n", __LINE__, nc_strerror(err)); goto fn_exit;}}
#define ERR {if(err!=NC_NOERR)printf("Error at line=%d: %s\n", __LINE__, nc_strerror(err));}

int main(int argc, char** argv)
{
    const char* filename="testfile";
    int err;
    int ncid, cmode;

    MPI_Init(&argc, &argv);

    /* create a new file for writing ----------------------------------------*/
    cmode = NC_NETCDF4 | NC_CLOBBER | NC_MPIIO;
    err = nc_create_par(filename, cmode, MPI_COMM_WORLD, MPI_INFO_NULL, &ncid); FATAL_ERR
    
    /* exit define mode */
    err = nc_enddef(ncid); ERR

    /* close the file */
    printf("========= application: nc_close start\n");
    err = nc_close(ncid); ERR
    printf("========= application: nc_close end\n");

fn_exit:
    MPI_Finalize();
    return 0;
}

```
</details>

### Compile and Run
[...]

PR #908, opened 2023-02-27
Fix HDF5 reference issue.

yzanhua:
According to [HDF5 Doc](https://docs.hdfgroup.org/hdf5/develop/group___h5_i.html#title5), the identifier returned by `H5Iget_file_id` needs to be released using `H5Fclose`. 

This fix darshan-hpc/darshan#906

### DIFF
--- a/darshan-runtime/lib/darshan-hdf5.c
+++ b/darshan-runtime/lib/darshan-hdf5.c
@@ -364,6 +364,9 @@ herr_t DARSHAN_DECL(H5Fflush)(hid_t object_id, H5F_scope_t scope)
                     tm1, tm2, rec_ref->last_meta_end);
             }
             H5F_POST_RECORD();
+
+            MAP_OR_FAIL(H5Fclose);
+            __real_H5Fclose(file_id);
         }
     }
 


darshan-runtime/lib/darshan-hdf5.c
```diff
@@ -364,6 +364,9 @@ herr_t DARSHAN_DECL(H5Fflush)(hid_t object_id, H5F_scope_t scope)
                     tm1, tm2, rec_ref->last_meta_end);
             }
             H5F_POST_RECORD();
+
+            MAP_OR_FAIL(H5Fclose);
```
shanedsnyder, line 368: Should we move this inside the `if` block above just to be safe? If `H5Iget_file_id` returns an error (i.e., `H5I_INVALID_HID`), that could lead to some noisy warnings/errors passing that value in to `HF5close`.

shanedsnyder (changes requested):

shanedsnyder:
Thanks a lot for investigating and fixing this @yzanhua !

I just had one small suggestion in the review, but your proposed fix makes sense. We were clearly forgetting to free up some HDF5 resources so will be good to fix that.

darshan-runtime/lib/darshan-hdf5.c
wkliao: Isn't that already inside of `if(file_id > 0)` block?

shanedsnyder: Oh wow, you're right. Not sure what I was thinking, probably just moving too fast. Thanks for the correction.

shanedsnyder:
Never mind on the requested change, that was a misunderstanding on my part. I'll go ahead and merge this in. Thanks again!

Issue #910, opened 2023-02-28
PyDarshan job summary cleanup


shanedsnyder:
From an in-person review of the current Darshan job summary reports at Argonne, it would be nice to try to address as much of this as we can before doing a release.

First some high-level things up for discussion:
- [x] drop blue I/O performance estimate text box, and instead add a per-module table detailing key details of the module (including total files/datasets(H5D)/variables(PNETCDF_VAR) accessed, total bytes read/written, and I/O performance estimate)
  - this way, each module starts off with a high-level summary table that describes key details of I/O, and gets rid of awkward text object we are using now
- [x] disable DXT heatmap plot generation by default, print warning if DXT is available but HEATMAP isn't, and add CLI option to force DXT usage
  - for some logs, it's probably just too expensive to generate DXT heatmaps by default, so this protects against that
  - this would require us to either push a little on CLI stuff in #809 or use something simple temporarily?
- [x] after dust settles, potentially reorganize plots/tables in per-module sections?
  - nothing concrete to suggest just yet (Tyler -0.5 on this)

Smaller things:
- [x] Captions are hard to read (light colored text, also italicized)
![image](https://user-images.githubusercontent.com/24571836/221916109-98e51b3f-8476-46b4-9f5c-903809fdd152.png)
- [x] Add "% runtime" y-axis label to right-side of I/O cost plot
![image](https://user-images.githubusercontent.com/24571836/221916672-29cfb0ba-f4ef-4e7d-967a-200cac298ed6.png)
- [x] Offset autolabeling text at 45 degrees in access size histogram plot and POSIX access pattern plot as is done in I/O op count graph
- [x] Don't autolabel 0 values to further help with cluttering of plots
- [x] If not too complicated, can we add more buffer to top of these plots to avoid autolabels overlapping with plot boundaries?
- [x] We should use a ylim of 0 for access histogram and op count plots to ensure no negative values are plotted.
![image](https://user-images.githubusercontent.com/24571836/221917799-2c6a0fe4-390a-4f8d-871c-1edbbd76e7f1.png)
- [x] Erroneous "type" row in File Count Summary table
![image](https://user-images.githubusercontent.com/24571836/221918843-c533dbd9-3097-4930-81e6-f66dd133a96c.png)
- [x] drop scientific notation entirely (partially addressed in #812?), and rely on humanize to generate sensible human readable byte strings?
- [x] File access by category table font really small
[...]

PR #911, opened 2023-02-28
PyDarshan job summary cleanup fixes

shanedsnyder:
See #910 for more details.

This PR currently has fixes for:

- Captions are hard to read (light colored text, also italicized)
-  Add "% runtime" y-axis label to right-side of I/O cost plot
-  Offset autolabeling text at 45 degrees in access size histogram plot and POSIX access pattern plot as is done in I/O op count graph
-  Don't autolabel 0 values to further help with cluttering of plots
- add a forgotten space to our warning string when a log has no data

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/style.css
+++ b/darshan-util/pydarshan/darshan/cli/style.css
@@ -346,9 +346,8 @@ figure img {
 }
 
 figcaption {
-  font-style: italic;
   font-size: 0.75em;
-  font-weight: 200;
+  font-weight: 300;
   margin: 0;
 }
 
--- a/darshan-util/pydarshan/darshan/cli/summary.py
+++ b/darshan-util/pydarshan/darshan/cli/summary.py
@@ -335,7 +335,7 @@ class ReportData:
         if not self.report.modules:
             # no data in report to summarize, print warning and that's it
             no_data_message = (
-                "This Darshan log file has no instrumentation records,"
+                "This Darshan log file has no instrumentation records, "
                 "there is no data to plot. Did this app do any I/O?"
             )
             fig = ReportFigure(
--- a/darshan-util/pydarshan/darshan/experimental/plots/data_access_by_filesystem.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/data_access_by_filesystem.py
@@ -13,7 +13,7 @@ import darshan
 import matplotlib
 matplotlib.use('Agg')
 import matplotlib.pyplot as plt
-
+import humanize
 
 def process_byte_counts(df_reads, df_writes):
     """
@@ -505,17 +505,18 @@ def plot_data(fig: Any,
     num_cats: an integer representing the number of categories
     to plot; default ``None`` plots all categories
     """
+    fontsize = 18
     list_byte_axes: list = []
     list_count_axes: list = []
     # use log10 scale if range exceeds
     # two orders of magnitude in a column
     use_log = [False, False]
-    for idx, series_pair in enumerate([[bytes_rd_series, bytes_wr_series, 1048576],
-                                       [file_rd_series, file_wr_series, 1]]):
+    for idx, series_pair in enumerate([[bytes_rd_series, bytes_wr_series],
+                                       [file_rd_series, file_wr_series]]):
         maxval = max(series_pair[0].max(), series_pair[1].max())
         minval = max(min(series_pair[0].min(), series_pair[1].min()), 1)
         # adjust ratio to MiB when needed
-        ratio = ((maxval / series_pair[2]) / (minval / series_pair[2]))
+        ratio = (maxval / minval)
         if ratio > 100:
             use_log[idx] = True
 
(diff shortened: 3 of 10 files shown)


tylerjereddy:
I'll try to take quick look tomorrow--I may try to draft a small testing class to enforce the changes that are easy to roundtrip-check from the produced html, and just label them as based on feedback from Argonne team.

tylerjereddy (commented):
Seems sensible enough--here's a suggestion for some matching tests and a fix for the index name Argonne doesn't like. I avoided `read_html` because of the optional `lxml` test dep, although that is a more robust way to do round-trip parsing on tables in the report.

```diff
diff --git a/darshan-util/pydarshan/darshan/lib/accum.py b/darshan-util/pydarshan/darshan/lib/accum.py
index 924e69ee..d60fdecd 100644
--- a/darshan-util/pydarshan/darshan/lib/accum.py
+++ b/darshan-util/pydarshan/darshan/lib/accum.py
@@ -100,5 +100,6 @@ def log_file_count_summary_table(derived_metrics,
     df.drop(columns="index", inplace=True)
     ret = plot_common_access_table.DarshanReportTable(df,
                                                       col_space=200,
-                                                      justify="center")
+                                                      justify="center",
+                                                      index_names=False)
     return ret
diff --git a/darshan-util/pydarshan/darshan/tests/test_plot_exp_common.py b/darshan-util/pydarshan/darshan/tests/test_plot_exp_common.py
index 27b56aae..8e29828d 100644
--- a/darshan-util/pydarshan/darshan/tests/test_plot_exp_common.py
+++ b/darshan-util/pydarshan/darshan/tests/test_plot_exp_common.py
@@ -160,6 +160,12 @@ def test_xticks_and_labels(log_path, func, expected_xticklabels, mod):
 
     assert_array_equal(actual_xticks, expected_xticks)
     assert_array_equal(actual_xticklabels, expected_xticklabels)
+    # see Argonne formatting reqs in gh-910
+    spines = ax.spines
+    assert not spines["top"].get_visible()
+    assert not spines["right"].get_visible()
+    assert spines["bottom"].get_visible()
+    assert spines["left"].get_visible()
 
 
 @pytest.mark.parametrize(
diff --git a/darshan-util/pydarshan/darshan/tests/test_summary.py b/darshan-util/pydarshan/darshan/tests/test_summary.py
index 65832066..0dc9d2e3 100644
--- a/darshan-util/pydarshan/darshan/tests/test_summary.py
+++ b/darshan-util/pydarshan/darshan/tests/test_summary.py
@@ -21,6 +21,22 @@ except ImportError:
     has_lxml = False
 
 
+def _enforce_html_report_aesthetics(report_str):
+    # formatting requirements, partly from Argonne
+    # feedback in gh-910
+    prog = re.compile(r"figcaption {\n.*\n.*\n.*\n}")
+    m = prog.search(report_str)
+    if m:
+        result = m.group(0)
+    assert "font-weight: 300" in result
+    assert "font-size: 0.75em" in result
+    assert not "italic" in result
+
+    # Argonne team doesn't like index label
+    # for File Count Summary table
+    assert not "<th>type</th>" in report_str
+
+
 @pytest.mark.parametrize(
     "argv", [
         ["./tests/input/sample.darshan"],
@@ -187,6 +203,7 @@ def test_main_all_logs_repo_files(tmpdir, log_filepath):
             with darshan.DarshanReport(log_filepath, read_all=False) as report:
                 with open(expected_save_path) as html_report:
                     report_str = html_report.read()
+                    _enforce_html_report_aesthetics(report_str=report_str)
                     if "DXT" in "\t".join(report.modules):
                         for dxt_mod in ["DXT_POSIX", "DXT_MPIIO"]:
                             if dxt_mod in report.modules:
@@ -215,7 +232,7 @@ def test_main_all_logs_repo_files(tmpdir, log_filepath):
                             assert "Consider enabling the runtime heatmap module" in report_str
                         else:
                             # check empty log warning and return
-                            assert "This Darshan log file has no instrumentation records" in report_str
+                            assert "This Darshan log file has no instrumentation records, " in report_str
                             return
 
                     # check if I/O cost figure is present
```

carns:
I like it.

The only other possibility would be to wrap the path to multiple lines (splitting on arbitrary characters) on the path name; we aren't lacking for vertical space.  I don't know if that's possible or not.   What you have is a good compromise.

shanedsnyder:
> > The only other possibility would be to wrap the path to multiple lines (splitting on arbitrary characters) on the path name; we aren't lacking for vertical space. I don't know if that's possible or not. What you have is a good compromise.
> 
> I didn't try this, but after browsing around briefly it may be possible using matplotlib [`text`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.text.html) (`wrap` option). We are using `annotate` for this currently, but I think we could switch to `text` without issue (we are actually using it already for the byte and file count strings in this plot).
> 
> I'll tinker with it and see how it goes.

Eh, it's a little more complicated. This wrap option seems to only work on text with whitespace, which you won't get in an FS mount point. Not sure we'll have a lot of options for wrapping given that -- I don't see any example solutions looking around.

I think I'm just going to go with what we have for now.

That said, we may want to revisit in the future. I'm imagining what we'd put there for something like DAOS, which has quite long identifiers for pools/containers. It might be easiest to just: a) put the category name text in a dedicated, fixed-size subplot to the left of the 2 data subplots; b) be more aggressive about shortening category strings so they convey information and aren't arbitrarily chopped (e.g., "DAOS cont=...09zb", just showing the last 4 digits of the ID or something like that). We can worry more about that when it becomes an issue, I think.

shanedsnyder:
I just committed some testing changes (including suggestions above from Tyler).

I think I'd like to stop any work on this PR and merge it in, and address remaining changes from #910 elsewhere (they are a little more involved and might warrant some discussion).

Note there are now some weird testing failures here when simply upgrading `pip`...

tylerjereddy:
The M1 failure probably persists here b/c Cirrus perhaps doesn't auto-merge into `main` before running the tests. I think we just ignore that then. If the rest of the CI passes I'll just do a quick scan through.

tylerjereddy:
Ok, I read through the diff one more time, spot checked the HTML report produced with `e3sm_io_heatmap_only.darshan`, and didn't see any issues. It sounds like the Argonne folks are in favor of most of these aesthetic adjustments so in it goes.

tylerjereddy (approved):

PR #915, opened 2023-03-08
BUG: set DXT heatmap limit

### LINKED ISSUE #914: strange behavior with DXT MPI-IO heatmap
I am digging into an application that reads five HDF5 files.  When I turned on DXT, the python report showed an unusual graph for the DXT heatmap.   Here's a screen shot of that report:

image

heatmap behaves as I would expect:  a bunch of HDF5 reads, followed by a long period of computation
and posix HEATMAP vs DXT looks nominal.
just eyeballing, it looks like DXT_MPIIO plotted everything in the HEATMAP case but left out the non-io parts.
See that big white patch in the heat map case? That's where the app is doing computation.  There's no computation in the dxt case.  The plot still knows there's 200-some seconds of runtime, but in the dxt case the io calls are scaled weird...to fill up the whole plot.

I've attached a zip file with the darshan log and the generated html.  

[dxt-mpiio-heatmap-from-hepps.zip](https://github.com/darshan-hpc/darshan/files/10922186/dxt-mpiio-heatmap-from-hepps.zip)

tylerjereddy:
* Fixes #914

* use a new `max_time` argument to compensate for the fact that DXT records are not guaranteed to encompass the wallclock duration of an instrumented runtime

* I could probably use help developing a suitable regression test, that seems a bit tricky without i.e., adding the connected log + reconstituting the plot binary data from the HTML and then parsing that somehow (certainly doable, but annoying)

* could also add a test for usage of `get_heatmap_df()` with/without the new argument for a case where it matters, though that doesn't directly guard against the regression in question

Pertinent part of updated report:

image

### DIFF
--- a/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
+++ b/darshan-util/pydarshan/darshan/experimental/plots/heatmap_handling.py
@@ -2,7 +2,7 @@
 Module of data pre-processing functions for constructing the heatmap figure.
 """
 
-from typing import Dict, Any, Tuple, Sequence, TYPE_CHECKING
+from typing import Dict, Any, Tuple, Sequence, TYPE_CHECKING, Optional
 
 import sys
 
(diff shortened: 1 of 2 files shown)


shanedsnyder:
Changes look seem reasonable enough to me.

Testing does seem like a pain in this case, but I'm not sure I'm much help honestly. The pandas code for turning sparse DXT data into heatmap data is pretty involved, never mind it's probably a little quirky comparing the 2 data structures (sparse DXT vs dense pandas representation) to see if they don't match. I'd be fine ignoring that now unless you have time to spend on it.

tylerjereddy:
ok I self merged then

Issue #383, opened 2021-05-13
investigate whether PyDarshan installs still work properly just using the darshan-util pkgconfig file


shanedsnyder:
In some testing on Cori using Anaconda, the pkg-config method for detecting the path to the darshan-util shared lib was the only method that did not work for me -- LD_LIBRARY_PATH and using PATH + darshan-parser both worked fine. We should confirm this isn't broken or isn't otherwise unportable.

To reproduce, build PyDarshan from source using `pip install /path/to/pydarshan`, then set PKG_CONFIG_PATH=/path/to/darshan/install/prefix/lib/pkgconfig and confirm importing the Darshan module fails due to not finding any shared library.

PR #917, opened 2023-03-10
MAINT: darshan-util library discovery changes for pydarshan

shanedsnyder:
- prefer library discovery method based on wheels, since these are shipped with their own libraries that should be used over other Darshan installs (e.g., Darshan installs provided by a facility)
- fix pkg-config based library discovery, originally mentioned in #383 
  - `--path` does not seem to be a valid `pkg-config` option, but we can extract the `prefix` variable from the file as I have done here

### DIFF
--- a/darshan-util/pydarshan/darshan/discover_darshan.py
+++ b/darshan-util/pydarshan/darshan/discover_darshan.py
@@ -86,15 +86,15 @@ def discover_darshan_pkgconfig():
 
     import subprocess
 
-    args = ['pkg-config', '--path', 'darshan-util']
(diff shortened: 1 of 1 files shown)


jakobluettgau (approved):
This looks good to me.

On a separate note, I'll create new issue for us to discuss if we want to introduce some environment variables to overwrite behaviors or switch on the verbose discovery.

tylerjereddy:
Ok, this has an approval and CI is passing so I'll merge.

I don't see any major issue either if it helps some HPC folks, though I would say this shared lib handling feels awkward to me.

PR #921, opened 2023-04-13
CI: deal with codecov yank

tylerjereddy:
* see drama here: https://about.codecov.io/blog/message-regarding-the-pypi-package/

* I think we should just be able to delete the `codecov` PyPI install and use their GHA plugin at this point, but I'll let CI confirm...


tylerjereddy:
The Cirrus M1 job is a separate issue, but should also be an easy fix.

shanedsnyder (approved):

shanedsnyder:
Ah, I see. Thanks for tracking those down and for the quick fixes.

PR #922, opened 2023-04-13
BUG: fix potential segfault with apps that `fork()` and use Darshan app exclusions

shanedsnyder:
The Darshan library is currently setting fork handlers near the beginning of its initialization routine -- notably, before we detect whether the app should be excluded from Darshan instrumentation (via new APP_EXCLUDE setting). These fork handlers expect Darshan library state to be initialized properly, but that clearly does not happen if the app is excluded (there is no Darshan library state in this case). This is currently leading to crashes for this scenario (excluded app that calls fork).

To fix, don't set Darshan fork handlers until it's known that library initialization succeeded.


carns (approved):
good catch; this looks great

PR #925, opened 2023-04-14
ENH: add darshan-runtime configure option for specifying a username env var

shanedsnyder:
This should address a longstanding issue with Darshan not always being able to determine the username associated with a job on Cray systems that use Slurm. Specifically, when Slurm users use the `--export` option to `srun` to set additional environment variables, this can wipe out env vars like `LOGNAME`, which Darshan uses to determine the associated username with a job.

This new configure option allows users to specify additional env vars that can be considered for determining a job's username (e.g., `SLURM_JOB_USER`). If specified, Darshan will consider this provided env var _first_.

In my testing on NERSC Cori, this is all working as expected:
- if this new config option is not specified, Darshan can not determine the username if the user uses the `--export` option to `srun`
- if this new config option is set to `SLURM_JOB_USER`, then Darshan can properly determine the username irrespective of whether the user uses `--export`

### DIFF
--- a/darshan-runtime/doc/darshan-runtime.txt
+++ b/darshan-runtime/doc/darshan-runtime.txt
@@ -90,6 +90,7 @@ make install
 * `--with-username-env=NAME`: this specifies the environment variable that Darshan
   should check to determine the username for a job. If not specified, Darshan
   will use internal mechanisms to try to determine the username.
+** NOTE: Darshan relies on the `LOGNAME` environment variable to determine a username, but this method isn't always reliable (e.g., on Slurm systems, `LOGNAME` can be wiped when specifying additional environment variables using the `--export` option to `srun`). This configure option allows specification of an additional environment variable to extract a username from (e.g., `SLURM_JOB_USER`).
 * `--with-log-path=DIR` (this, or `--with-log-path-by-env`, is mandatory): This
   specifies the parent directory for the directory tree where Darshan logs will
   be placed.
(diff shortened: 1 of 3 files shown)


darshan-runtime/doc/darshan-runtime.txt
```diff
@@ -87,6 +87,9 @@ make install
   scheduler does not advertise the job ID) then you can specify `NONE` here.
   Darshan will fall back to using the pid of the rank 0 process if the
   specified environment variable is not set.
+* `--with-username-env=NAME`: this specifies the environment variable that Darshan
```
carns, line 90: Maybe we should have either a) a recipe (or recipe variation) for Slurm systems or b) some rationale for when/why you might want this?  This is a good fix, but I'm worried that people installing on systems with Slurm schedulers won't find it.

carns (commented):
Code changes look great, just thinking about how to make sure people find it.

darshan-runtime/doc/darshan-runtime.txt
shanedsnyder: This fix is primarily intended for Slurm systems which we also disable the use of `cuserid()`, which is just a subset of Cray/HPE system at this point. If `cuserid()` isn't disabled, this fix isn't likely to be useful.

I just scanned darshan-runtime docs and there really isn't any mention of the cuserid issue either. Maybe I can just update our existing Cray recipe to mention these issues and give an example there? That might be more helpful than burying in a note up in the long list of configure options.

We should also update the darshan-runtime package in Spack for next release to take advantage of this option when using the `+slurm` variant.

carns: Ah right.  I wonder if the problematic cuserid systems have about aged out at this point?

At any rate, that seems reasonable to put a mention of this in the existing Cray recipe.  Also agree with leveraging the +slurm spack variant.

shanedsnyder: > Ah right. I wonder if the problematic cuserid systems have about aged out at this point?

I had hoped this was the case, but we actually ran into this problem again recently on NERSC Cori (see #699), so I feel like we should discourage usage of cuserid on Cray's indefinitely. It only affects statically-linked executables, which seem to increasingly be aged out themselves, but there will probably will always be apps forcing static linking for some reason -- we'll probably need to continue to be extra careful to avoid crashing apps in this context.

shanedsnyder: Alright, I think this should be addressed via [f891648](https://github.com/darshan-hpc/darshan/pull/925/commits/f8916489cedc3f88eba3cbb84f38dc670f98488d)

PR #937, opened 2023-06-05
TST: updated regression testing environments for HPC systems

shanedsnyder:
* add entirely new regression test environments for OLCF Frontier and ALCF Polaris
  * Polaris slightly involved as had to develop new scripts based on PBS job scheduler
* remove NERSC Cori regression test environment as the system has been decommissioned
* renamed old cray-module-alcf -> cray-module-alcf-theta to clarify which system the environment is intended for
* modified NERSC Perlmutter and OLCF Crusher tests to use given Darshan install prefix rather than just using the default Darshan module provided by the system
* consolidate old README.CRAY.ALCF.txt and README.CRAY.NERSC.txt into a more general README.CRAY.txt file that describes how to run regression tests on Cray systems
* updated regression testing README to list newly added Cray regression testing environments


carns (approved):
looks good, sorry for the review delay

shanedsnyder:
I eventually figured out the Polaris issue. We need to ensure we are using the GNU compilers when building Darshan there, as the default Nvidia compilers don't properly support "alias" attributes -- this prevents us from being able to intercept `PMPI_` symbols which can be problematic for instrumenting some implementations of MPI fortran bindings.

Merging.

PR #938, opened 2023-06-06
3.4.3 release prep

shanedsnyder:
Updated Darshan/PyDarshan Changelogs and version numbers ahead of the 3.4.3 release.

This should also kick off build of our 3.4.3.0 PyDarshan wheels, which we can then upload to PyPI.


carns (approved):
Great!  Go for launch :)

PR #942, opened 2023-06-16
BUG: never disable "shared reductions" for heatmap mod

### LINKED ISSUE #941: BUG: PyDarshan can't parse some logs (report from Slack)
Reported by Nafise Moti on Slack, though just a few of us have the reproducing log file for now.

I was able to reproduce the traceback below the fold for a number of versions of darshan, rebuilding from scratch (including C libs) each time.

<details>

```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/treddy/github_projects/darshan/darshan-util/pydarshan/darshan/__main__.py", line 3, in <module>
    main()
  File "/Users/treddy/github_projects/darshan/darshan-util/pydarshan/darshan/cli/__init__.py", line 164, in main
    mod.main(args)
  File "/Users/treddy/github_projects/darshan/darshan-util/pydarshan/darshan/cli/summary.py", line 569, in main
    report_data = ReportData(log_path=log_path)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/treddy/github_projects/darshan/darshan-util/pydarshan/darshan/cli/summary.py", line 127, in __init__
    self.report = darshan.DarshanReport(log_path, read_all=True)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/treddy/github_projects/darshan/darshan-util/pydarshan/darshan/report.py", line 374, in __init__
    self.open(filename, read_all=read_all)    
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/treddy/github_projects/darshan/darshan-util/pydarshan/darshan/report.py", line 431, in open
    self.read_all()
  File "/Users/treddy/github_projects/darshan/darshan-util/pydarshan/darshan/report.py", line 547, in read_all
    self.read_all_heatmap_records()
  File "/Users/treddy/github_projects/darshan/darshan-util/pydarshan/darshan/report.py", line 627, in read_all_heatmap_records
    heatmaps[mod].add_record(rec)
  File "/Users/treddy/github_projects/darshan/darshan-util/pydarshan/darshan/datatypes/heatmap.py", line 78, in add_record
    raise ValueError("Record nbins is not consistent with current heatmap.")
ValueError: Record nbins is not consistent with current heatmap.
```
</details>

Observed at these branches/tags/hashes:

- `main`
- `darshan-3.4.2`
- `darshan-3.4.1`
- `darshan-3.4.0`
- 1bfd1f26
- 592f2279
- 548d9aed6428 (big merge of pydarshan dev branch to `main` I think)

If I go back to `darshan-3.3.1`, I end up with `LD_LIBRARY_PATH` issues, probably the code base has diverged too far from PyDarshan modernization at that point anyway.

Observation, in case it helps the C-level folks: it seems that the number of (`STDIO` HEATMAP) bins is correct for a large number of HEATMAP records before the mismatch happens during retrieval from the binary log.

Using `darshan-parser` with i.e.,  `grep -E HEATMAP_READ_BIN | wc -l` type analysis for the problem log:

read bins: `22460` / `104` bins per block (based on inspection/debug prints) -> `215.96` (not divisible)
write bins: `22460` (same situation obviously)

If I try to similarly dissect the "working" `e3sm_io_heatmap_only.darshan` log:

`175104` write bins / `114` bins per block = 1536 "blocks" == `512` ranks * 3 types of heatmaps
`175104` read bins (same thing)

Ok, so then I got really curious because manual inspection of the `darshan-parser` output for the bad log seemed reasonable enough when spot checking. Need to programmatically analyze the bin number increments.

`./darshan-parser /Users/treddy/rough_work/darshan/bad_log/darshan.log_202212161428 | grep -E HEATMAP_WRITE_BIN > log.txt`

Use the Python code below the fold to parse the bin digits out and plot them for bad and good logs, respectively:

<details>

```python
import re

import numpy as np
import matplotlib
matplotlib.use("Agg")

prog = re.compile(r".*HEATMAP_WRITE_BIN_(\d+).*")

bin_numbers = []

with open("log.txt") as infile:
    for line in infile:
        m = prog.match(line).group(1)
        bin_numbers.append(int(m))

arr = np.array(bin_numbers)
bincounts = np.bincount(arr)

fig, ax = plt.subplots()
ax.plot(np.arange(arr.max() + 1), bincounts, marker="o")
ax.set_xlabel("HEATMAP BIN Number")
ax.set_ylabel("frequency")
fig.savefig("fig.png", dpi=300)
```

</details>

image

`./darshan-parser /Users/treddy/github_projects/darshan-logs/darshan_logs/e3sm_io_heatmaps_and_dxt/e3sm_io_heatmap_only.darshan | grep -E HEATMAP_WRITE_BIN > log.txt`

image

So, indeed, it seems that PyDarshan and darshan-parser agree that the bad log is missing a few of the max bin number entries for a few ranks. At this point, I suspect we're looking at either a bindings or log-specific issue I should hand over to the Argonne folks?

shanedsnyder:
The shared reduction callback for the heatmap module is used to reach a consistent heatmap format across all ranks at shutdown time, which our analysis tools expect.

Furthermore, a "shared reduction" doesn't really apply to heatmap data, as it's already captured on every process (as opposed to a shared record on rank 0) -- ignoring the disable shared reduction flag is not an issue for the heatmap module, making this an easy fix.

Fixes #941

### DIFF
--- a/darshan-runtime/lib/darshan-core.c
+++ b/darshan-runtime/lib/darshan-core.c
@@ -637,11 +637,19 @@ void darshan_core_shutdown(int write_log)
                 }
 
                 /* allow the module an opportunity to reduce shared files */
-                if(this_mod->mod_funcs.mod_redux_func && (mod_shared_rec_cnt > 0) &&
-                   !final_core->config.disable_shared_redux_flag)
+                if(this_mod->mod_funcs.mod_redux_func && (mod_shared_rec_cnt > 0))
                 {
-                    this_mod->mod_funcs.mod_redux_func(mod_buf, final_core->mpi_comm,
-                        mod_shared_recs, mod_shared_rec_cnt);
+                    /* run reductions as long as they aren't disabled */
+                    /* NOTE: shared reductions should never be disabled for the
+                     *       HEATMAP module, as the shared reduction step is used
+                     *       to produce a consistent heatmap format across ranks
+                     */
+                    if(!final_core->config.disable_shared_redux_flag ||
+                       (i == DARSHAN_HEATMAP_MOD))
+                    {
+                        this_mod->mod_funcs.mod_redux_func(mod_buf, final_core->mpi_comm,
+                            mod_shared_recs, mod_shared_rec_cnt);
+                    }
                 }
             }
 #endif


shanedsnyder:
I didn't add any testing yet, but this might warrant a regression test on our CI before merging.

shanedsnyder:
We should also look at just resolving the inconsistent heatmaps this bug produces in the darshan-util code. It may be a simple fix and provide a way to resolve the problem for any affected logs.

carns (approved):
lgtm

shanedsnyder:
I'm actually going to punt on a regression test for this. We now have a runtime library fix for this bug and a util library fix to correct older logs that exhibit the bug -- this error condition won't be easy to trip and with 2 layers of fixes, I'll just call it a day for now.

I'll go ahead and merge this.

PR #951, opened 2023-07-25
CI: pin Cython in runtime job

### LINKED ISSUE #950: CI: runtime job failure
This is showing up in unrelated PRs lately from two summer students, for example here: https://github.com/darshan-hpc/darshan/actions/runs/5624859270/job/15307599528?pr=949

Looks like an `h5py` source build issue with Cythonization, and since Cython `3.0` was recently released (and it was a huge change), I bet we could just pin it in this line: https://github.com/darshan-hpc/darshan/blob/main/.github/workflows/runtime_ci.yml#L29

@yariseidenbenz or @Yanlilyu -- do you want to take a stab at fixing up that CI issue? If not, I'll try to open a PR with that patch tomorrow. I think the syntax would just be `cython<3.0` for now.

tylerjereddy:
* Fixes #950

* looks like `h5py` from-source build is struggling with the recently-released Cython `3.0.0`, so let's see if pinning to older Cython series helps get CI green for now

### DIFF
--- a/.github/workflows/runtime_ci.yml
+++ b/.github/workflows/runtime_ci.yml
@@ -26,7 +26,7 @@ jobs:
           sudo apt-get update -y
           sudo apt-get install -y hdf5-tools libhdf5-openmpi-dev openmpi-bin
           python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


shanedsnyder (approved):

shanedsnyder:
LGTM, thanks!

PR #959, opened 2023-09-26
MAINT: 3.12 support

tylerjereddy:
* a few adjustments to `pyproject.toml` to allow for `pip install -v .` followed by `python -m pytest --pyargs darshan -n 32` to pass the full suite locally on x86_64 Linux with Python `3.12.0rc3`, given that the final release of Python `3.12.0` is about a week away I think

* changing the version of `setuptools` may break the developer experience with "editable" installs, since I recall pinning it a while back for that reason, but I think supporting isolated wheel builds/install with `3.12.0` should likely take priority

* some of the metadata updates to `pyproject.toml` are apparently newly-required fields that must be present (`pip install -v .` fails otherwise)

* perhaps we can delay adding `3.12` to the CI testing until the final release, though GHA does have the RC available if we want it

[skip cirrus]

### DIFF
--- a/darshan-util/pydarshan/pyproject.toml
+++ b/darshan-util/pydarshan/pyproject.toml
@@ -1,11 +1,13 @@
 [build-system]
 requires = [
     "wheel",
-    "setuptools<64.0.0",
+    "setuptools>=65.0.0",
 ]
 
 [project]
+name = "darshan"
 requires-python = ">=3.7"
+version = "3.4.4.0"
 
 [tool.cibuildwheel]
 environment = "PYDARSHAN_BUILD_EXT=1"
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         platform: [ubuntu-latest,
                    macos-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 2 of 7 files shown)


darshan-util/pydarshan/pyproject.toml
```diff
@@ -1,11 +1,13 @@
...
 requires = [
     "wheel",
-    "setuptools<64.0.0",
+    "setuptools>=65.0.0",
 ]
 
 [project]
+name = "darshan"
 requires-python = ">=3.7"
+version = "3.4.4.0"
```
tylerjereddy, line 10: many projects will tend to use version numbers that have development indicators like `3.4.4.0.dev0` to avoid confusion with i.e., the final release or previous release with that same number, but I thought it best to just follow our current convention and paste this in from `setup.py` for now.

There may not be a compelling reason to change it anyway, unless for example we became so popular that we needed nightly or weekly wheels, where the dev tag would have a more compelling reason to exist to distinguish from stable releases.

tylerjereddy:
Needed to add a few shims for `FutureWarning`s from newer `pandas` release as well.

tylerjereddy:
There's still a type hinting error (`mypy`) to deal with: `darshan/experimental/plots/plot_io_cost.py:180: error: "_AxesBase" has no attribute "legend"  [attr-defined]`

Maybe because of a newer `matplotlib`, I won't get to that today though.

tylerjereddy:
Latest revisions pushed in are summarized below, but we'll have to wait and see if CI is happy:

    * `plot_io_cost` now sets legend properties via the
    raw axis object instead of via the twinned object (fixes
    a `mypy` complaint)
    
    * add Python `3.12` to the main CI matrix based on
    reviewer feedback
    
    * update release checklist to reflect version
    number now maintained in `pyproject.toml` as well

shanedsnyder (approved):

PR #961, opened 2023-10-23
BUG: fixes for HDF5 module method for determining whether MPIO VFD is used

### LINKED ISSUE #960: H5Pset_fapl_mpio fails with darshan
Hi,

I was trying to trace some HDF5 applications and stumbled upon this issue.

HDF5: 1.8.20 (parallel hdf5 enabled)
Darshan version: darshan-3.4.4

Darshan was configured using:
`./configure --prefix=/xxx/darshan-3.4.4/install --with-log-path=/xxx/darshan-logs --with-jobid-env=SLURM_JOB_ID --with-hdf5=$HDF5_1_8_HOME --enable-hdf5-mod CC=mpicc`

The issue can be reproduced using the following code:
```C
int main(int argc, char* argv[]) {
    MPI_Init(&argc, &argv);

    int res;
    hid_t plist_id = H5Pcreate(H5P_FILE_ACCESS);
    if (plist_id == H5I_INVALID_HID)
        printf("H5Pcreate failed\n");

    res = H5Pset_fapl_mpio(plist_id, MPI_COMM_WORLD, MPI_INFO_NULL);
    if (res < 0)
        printf("H5Pset_fapl_mpio failed\n");

    H5Pclose(plist_id);
    MPI_Finalize();
}
```

Compile & run:
```bash
mpicc test_phdf5.c -o test_phdf5 -I$HDF5_1_8_HOME/include -L$HDF5_1_8_HOME/lib -lhdf5
srun -n1 --overlap --export=ALL,LD_PRELOAD=$libdarshan ./test_phdf5
```

Without darshan, the code finishes without any error. With darshan, `H5Pset_fapl_mpio` call fails.

shanedsnyder:
* use `H5Pget_fapl_mpio` instead of `H5Pget_driver`
* this code temporarily disables HDF5 error reporting to avoid noisy error messages in the case the associated VFD is not MPIO
* remove old hack to declare `H5FD_mpio_init` as a weak symbol

Fixes #960

For some background, we added the weak symbol hack mentioned above in our last release to work around a weird issue when LD_PRELOADing Darshan with HDF5 support leading to loader issues like this:
```
symbol lookup error: /home/shane/software/darshan/darshan-dev/install/lib/libdarshan.so: undefined symbol: H5FD_mpio_init
```

While this hack works as intended for non-HDF5 apps, it does not work as intended for HDF5 apps -- the weak dummy function is being linked in rather than the real implementation of  `H5FD_mpio_init`, causing errors as seen in #960.

Rather than go down the rabbit hole to understand why our usage of `H5FD_MPIO` is causing the HDF5 library to need to be linked (this wasn't the case in HDF5 versions prior to 1.13), I've just moved away from using that macro entirely. `H5Pget_fapl_mpio` can provide the same functionality, with the caveat that we have to temporarily disable HDF5 error printing when probing, as it will print an error message if called on a file access property list that does not have the MPIO VFD enabled.

### DIFF
--- a/darshan-runtime/lib/darshan-hdf5.c
+++ b/darshan-runtime/lib/darshan-hdf5.c
@@ -29,14 +29,6 @@
 
 #include <hdf5.h>
 
-/* NOTE: this is a dummy definition of H5FD_mpio_init() with a weak alias.
- * It prevents "libdarshan.so: undefined symbol: H5FD_mpio_init" errors if
- * Darshan is enabled via LD_PRELOAD for non-HDF executables.  This problem
- * may occur if Darshan itself was compiled against HDF5 1.13.2.
- */
-hid_t dummy_H5FD_mpio_init(void) { return(0); }
-hid_t H5FD_mpio_init (void) __attribute__ ((weak, alias("dummy_H5FD_mpio_init")));
-
 /* H5F prototypes */
 DARSHAN_FORWARD_DECL(H5Fcreate, hid_t, (const char *filename, unsigned flags, hid_t create_plist, hid_t access_plist));
 DARSHAN_FORWARD_DECL(H5Fopen, hid_t, (const char *filename, unsigned flags, hid_t access_plist));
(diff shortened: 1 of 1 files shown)


shanedsnyder:
IIRC, there's not a way to force the the currently failing CI test (which is using the Cirrus service to test Apple M1) to re-run without jumping through extra hoops.

The warning messages that led to the error there seem to have been resolved in PR #959 which is now merged. In any case, the changes here are obviously unrelated to the Python warnings for our log analysis code. Given that, I'll go ahead and force the merge through despite the failed test.

PR #968, opened 2023-12-15
Fix cibuildwheel process failing due to undeclared dynamic pyproject.toml metadata and missing test dependencies 

jakobluettgau:
cibuildwheel builds fail complaining about a missing test dependencies (pytest) and dynamic properties in pyproject.toml not declared explicitly raising the following error:

`/tmp/pip-build-env-9ywkt6zf/overlay/lib/python3.8/site-packages/setuptools/config/_apply_pyprojecttoml.py:75: _MissingDynamic: `description` defined outside of `pyproject.toml` is ignored`

This pull request updates pyproject.toml and marks the affected fields as dynamic and adds missing test dependencies to resolve the issues.

### DIFF
--- a/darshan-util/pydarshan/pyproject.toml
+++ b/darshan-util/pydarshan/pyproject.toml
@@ -8,6 +8,7 @@ requires = [
 name = "darshan"
 requires-python = ">=3.7"
 version = "3.4.4.0"
+dynamic = ["description", "readme", "keywords", "classifiers"]
 
 [tool.cibuildwheel]
 environment = "PYDARSHAN_BUILD_EXT=1"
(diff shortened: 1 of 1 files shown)


shanedsnyder:
Looks like still some failures in CI here. I've spent some time looking at this today and think we should either move completely to `pyproject.toml` or just pin `setuptools` to a prior version for the time being -- have a look at #969.

Assuming we get that working with all the quirks in PyDarshan, I'd just prefer that rather than maintaining multiple approaches.

jakobluettgau:
Yeah I am all for consolidating and moving to pyproject.toml (there is also some old build pipeline artifacts we can drop then), this restored local builds with cibuildwheels for me, but I think the github workflow was setup differently anyway and just ran into the similar errors. I tested #969 and it works for me locally and seems to pass CI in github too, so this one can be closed and we move forward with #969 instead.

PR #971, opened 2024-01-12
CI: initial setup of end-to-end regression CI

shanedsnyder:
We've long had some end-to-end (generating logs and analyzing log results using `darshan-parser`) regression tests in `darshan-test/regression` in our repo. These tests were traditionally run on a Jenkins CI instance at the ALCF. This PR attempts to fold those into a Github workflow that runs on PRs to `main`.

### DIFF
--- a/.github/workflows/end_to_end_pytest.yml
+++ b/.github/workflows/end_to_end_pytest.yml
@@ -1,4 +1,4 @@
-name: Runtime Testing
+name: End-to-end Testing (pytest)
 
 on:
   push:
@@ -9,7 +9,7 @@ on:
       - main
 
 jobs:
-  test_runtime:
+  end_to_end_pytest:
     strategy:
       matrix:
         platform: [ubuntu-latest]
--- a/.github/workflows/end_to_end_regression.yml
+++ b/.github/workflows/end_to_end_regression.yml
@@ -1,0 +1,74 @@
+name: End-to-end Testing (regression)
+
+on:
+  push:
+    branches:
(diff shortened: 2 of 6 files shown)


shanedsnyder:
Okay, there's one outstanding test failure here, that I think corresponds to an actual bug that we'll need to resolve in Darshan. I'll attempt that in a separate PR so it's not buried in this one.

shanedsnyder:
So, we should now have what used to be our "nightly" test running on PRs to `main`. I reduced the number of test cases down to the following:
- `LD_PRELOAD` based tests
- MPICH profiling configuration hooks with dynamic linking
- MPICH profiling configuration hooks with static linking (this one's failing for now)

Notably, I got rid of the tests of our compiler wrappers. We haven't really been updating those and they only work(ed) with MPICH, which supports the profiling configuration stuff already, so not sure it's something we should be concerned about going forward.

shanedsnyder:
A fix for the remaining test failure is now in #975.

carns:
> 
> Notably, I got rid of the tests of our compiler wrappers. We haven't really been updating those and they only work(ed) with MPICH, which supports the profiling configuration stuff already, so not sure it's something we should be concerned about going forward.

Unrelated to this PR, but I think at this point we should mark that method deprecated in the documentation (or maybe simply remove it outright).  It was fragile even when it was actively maintained, and at this point I don't think it offers any advantage over the profile configuration method.

carns (approved):
Looks great.

Issue #972, opened 2024-01-13
CI, BENCH: asv/seaborn conflict in CI


tylerjereddy:
Noticed this in gh-971; log is: https://github.com/darshan-hpc/darshan/actions/runs/7508377198/job/20443721367?pr=971

Upstream report/query is here: https://github.com/airspeed-velocity/asv/issues/1363

A temporary pin that worked for me locally to fix it is:

```diff
--- a/darshan-util/pydarshan/benchmarks/asv.conf.json
+++ b/darshan-util/pydarshan/benchmarks/asv.conf.json
@@ -40,6 +40,7 @@
         "numpy": [],
         "cffi": [],
         "pytest": [],
+        "seaborn": ["0.13.0"],
     },
```

Perhaps either `seaborn` or `asv` needs to change their module naming.

shanedsnyder:
Ah, good catch. I was sidetracked by other issues in my original PR that I hadn't noticed this. I will try to apply this workaround for the time being to get CI back on the right track. Thanks!

PR #974, opened 2024-01-23
CI, BENCH: pin seaborn to avoid namespace issues with asv

shanedsnyder:
Temporary workaround for issue mentioned by @tylerjereddy in #972.

### DIFF
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -92,6 +92,7 @@ jobs:
           export LD_LIBRARY_PATH=$PWD/darshan_install/lib
           export DYLD_FALLBACK_LIBRARY_PATH=$PWD/darshan_install/lib
           cd darshan-util/pydarshan/benchmarks
+          python -m pip install "seaborn==0.13.0"
           python -m asv check -E existing
       - name: codecov check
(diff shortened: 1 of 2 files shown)


tylerjereddy (commented):
Oh, I see why it is still failing, we use a flag for `asv check` in CI that leverages the "existing" Python interpreter/environment for all deps instead of using an isolated venv (to save on CI time).

So, I suppose we may want to additionally do this here:

```diff
--- a/.github/workflows/main_ci.yml
+++ b/.github/workflows/main_ci.yml
@@ -92,6 +92,7 @@ jobs:
           export LD_LIBRARY_PATH=$PWD/darshan_install/lib
           export DYLD_FALLBACK_LIBRARY_PATH=$PWD/darshan_install/lib
           cd darshan-util/pydarshan/benchmarks
+          python -m pip install "seaborn==0.13.0"
           python -m asv check -E existing
       - name: codecov check
         uses: codecov/codecov-action@v3
```

shanedsnyder:
Thanks, I made a small tweak to your suggested changes to get things working on Python 3.7, where seaborn-13 is apparently not available. Looks like things are passing now across the board.

tylerjereddy (approved):

PR #975, opened 2024-01-25
BUG: fix handling of `fscanf`/`__isoc99_fscanf`

shanedsnyder:
A long time ago we added a wrapper and an autoconf test for the `__isoc99_fscanf` function that can be found in newer glibc implementations. It doesn't appear as though this old autoconf test appropriately tests whether we should include a wrapper for the `__isoc99_fscanf` routine. Locally, this autoconf test was failing even though I have `__isoc99_fscanf`, leading to missed instrumentation of these calls.

We can't simply offer wrappers for both `fscanf` and `__isoc99_fscanf`, as this leads to symbol redefinition errors when building Darshan's shared library.

This PR tries to address the issue by detecting the presence of `__isoc99_fscanf`, and including a wrapper for that call if found. If not found, we include our traditional `fscanf` wrapper. The key is to ensure that only one of these wrappers is available in Darshan, and that we prefer `__isoc99_fscanf` if available.

### DIFF
--- a/darshan-runtime/configure.ac
+++ b/darshan-runtime/configure.ac
@@ -740,19 +740,15 @@ if test "x$enable_darshan_runtime" = xyes ; then
    # End of MPI-only checks
    #
 
-   AC_MSG_CHECKING(for fscanf redirect)
-   AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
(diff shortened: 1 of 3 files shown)


carns (approved):
Looks great, thanks!

PR #976, opened 2024-02-07
CI: new workflow for triggering ALCF GitLab CI on Polaris

shanedsnyder:
Uses: https://github.com/marketplace/actions/trigger-gitlab-ci-through-webhooks

### DIFF
--- a/.github/workflows/end_to_end_regression_polaris.yml
+++ b/.github/workflows/end_to_end_regression_polaris.yml
@@ -1,0 +1,19 @@
+name: End-to-end Testing (regression) Polaris
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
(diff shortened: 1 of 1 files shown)


shanedsnyder:
Okay, all done here (well, there are some smaller unrelated Python issues, but haven't had a close look there yet).

This new workflow will kickoff a series of regression tests on our ALCF GitLab CI instance which builds Darshan and runs test on the Polaris system. Right now, the workflow is configured to run on _push_ events to main and via manual triggers via Github actions tab. Allowing the workflow to be run on every commit to a PR leads to potentially piling up jobs on Gitlab CI (which can take quite some time as its serially running 8 different batch jobs on compute nodes). I also further restricted the workflow to only run on changes to darshan-runtime (and to the workflow itself) -- testing darshan-util stuff on a supercomputer is probably unnecessary.

carns (approved):

Issue #872, opened 2022-12-08
Unable to produce darshan trace for pytorch data loader when multiple threading is enabled


zhenghh04:
As described in the title, when pytorch load the data using multi threading (num_workers > 0 in the data loader), darshan was not able to produce the tracing data for read. 

It does produce log when num_workers=0 in which case the main thread does the data loader.  

A small program to reproduce this problem is: [test_darshan.tgz](https://github.com/darshan-hpc/darshan/files/10187983/test_darshan.tgz)

./test_darshan.py - python example that uses pytorch data loader to load np array 
./test_darshan.sh - the script for running the example and processing the darshan log

Output for num_workers=0
```
...
# DXT, file_id: 9786240366124566250, file_name: /home/huihuo.zheng/workspace/dlio/tests/data/15.npz
# DXT, rank: 0, hostname: ebafcdd5a20c
# DXT, write_count: 3, read_count: 6
# DXT, mnt_pt: /home/huihuo.zheng/workspace/dlio, fs_type: fuse.grpcfuse
# Module    Rank  Wt/Rd  Segment          Offset       Length    Start(s)      End(s)
 X_POSIX       0  write        0               0             267      0.5358      0.5361
 X_POSIX       0  write        1               0              59      0.5361      0.5363
 X_POSIX       0  write        2             267              77      0.5363      0.5366
 X_POSIX       0   read        0               0             344      0.5529      0.5533
 X_POSIX       0   read        1             322              22      0.5538      0.5538
 X_POSIX       0   read        2             344               0      0.5538      0.5538
 X_POSIX       0   read        3             302              42      0.5538      0.5538
 X_POSIX       0   read        4             267              77      0.5538      0.5538
 X_POSIX       0   read        5               0             344      0.5541      0.5542
...
```

Output for num_workers > 0 (4): (modifying L10 in test_darshan.py)
```
...
# DXT, file_id: 9786240366124566250, file_name: /home/huihuo.zheng/workspace/dlio/tests/data/15.npz
# DXT, rank: 0, hostname: ebafcdd5a20c
# DXT, write_count: 3, read_count: 0
# DXT, mnt_pt: /home/huihuo.zheng/workspace/dlio, fs_type: fuse.grpcfuse
# Module    Rank  Wt/Rd  Segment          Offset       Length    Start(s)      End(s)
 X_POSIX       0  write        0               0             267      0.5410      0.5413
 X_POSIX       0  write        1               0              59      0.5413      0.5416
[...]

PR #978, opened 2024-03-13
ENH: wrap _exit in Darshan's shared library

shanedsnyder:
Python apps using the `multiprocessing` package can invoke the `_exit()` call directly. E.g., see [this code](https://hg.python.org/cpython/file/e04cd497aa41/Lib/multiprocessing/popen_fork.py#l76).

Invoking `_exit()` directly bypasses Darshan's shared library destructor, which is what is used to write out its log file. So, these processes spawned by `multiprocessing` are properly instrumenting I/O calls, but never get an opportunity to write the instrumentation data to log file. 

This PR adds a function wrapper for `_exit()`, which Darshan uses to intercept `_exit()` calls, invoking the Darshan shutdown procedure before calling the real _exit implementation. This modification to Darshan seems to fix some issues reported by users related to this issue (e.g., #872). Let's see if it can get through our CI without causing other issues.

Note that this code is only included in our shared library, since it's intended for Darshan's non-MPI mode (which only works when dynamically linking).

Keeping this PR as a WIP at least until I add some autoconf support to disable this functionality entirely, in case it becomes problematic in some scenarios.

### DIFF
--- a/darshan-runtime/configure.ac
+++ b/darshan-runtime/configure.ac
@@ -754,6 +754,18 @@ if test "x$enable_darshan_runtime" = xyes ; then
    # look for glibc-specific functions
    AC_CHECK_FUNCS([pwritev preadv pwritev2 preadv2])
 
+   # allow users to opt out of wrapping of _exit as a shutdown hook in
+   # Darshan's non-MPI mode, in case this functionality is problematic
+   AC_ARG_ENABLE([exit-wrapper],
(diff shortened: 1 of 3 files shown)


carns (dismissed):
Great!  I know this is still a WIP but I wanted to go ahead and review.  It seems like a straightforward fix.

shanedsnyder:
I'm going to go ahead and merge this then -- the only failing test is an LDMS thing that's unrelated to changes here.

PR #979, opened 2024-03-14
WIP: Add "extra info" field to the records

roblatham00:
dusting off the "augment DXT traces" work from a while back.


wkliao:
Hi, @roblatham00 
Is this PR now a part of #1066?

PR #980, opened 2024-03-18
CI: allow LDMS CI test workflow dispatch from github

shanedsnyder:


shanedsnyder:
@Snell1224 -- are there any known issues with LDMS currently? CI tests are eventually failing here, seemingly because something is hanging when starting the LDMS daemon.

Snell1224:
> @Snell1224 -- are there any known issues with LDMS currently? CI tests are eventually failing here, seemingly because something is hanging when starting the LDMS daemon.

Hmm I don’t know of any current issues with LDMS and have never seen this behavior before so I’ll need to do some debugging and see if any of the LDMS developers know about it. 

I’ll let you know when it’s fixed and thank you for catching this!

PR #981, opened 2024-03-21
Fix for LDMS CI Workflow

Snell1224:
@shanedsnyder Turns out this was a known issue that has since been fixed. The ``ldmsd`` command was modified to run as a background process when executed so a ``&`` needed to be added at the end of this command.

Also, I removed a line of workaround code as it's no longer needed. :)


shanedsnyder:
Awesome, thanks for the quick fix @Snell1224 !

shanedsnyder (approved):

PR #993, opened 2024-06-27
fix STDIO counter names in documentation

jeanbez:
The documentation seems to refer to some STDIO counters with a slightly different name than the ones shown in the parsers.


shanedsnyder (approved):

shanedsnyder:
Good catch, @jeanbez, thanks! Merging this in.

PR #1002, opened 2024-08-06
add workaround for potential zlib-ng problem

### LINKED ISSUE #995: Darshan-util fails with poorly formatted mount table
Hello,

darhsan-util clis fail with the following message, starting from `darshan-parser`

Error: poorly formatted mount table in darshan log file.

I am instrumenting an MPI application using intel-oneapi-mpi/2021.10.0  with darshan-runtime/3.4.5. The logfile is generated without error and the darshan-parser works on logfiles generated by instrumenting the same application using openmpi.

Thank you for your help,

Laura

carns:
In some cases if zlib-ng is used as the gzip library for Darshan, it will indicate that the correct number of bytes were decompressed, but will also modify some bytes in the buffer beyond that point.  In the Darshan this specifically can produce a corrupted mount table by appending erroneous characters onto the string data.

Fixes #995

### DIFF
--- a/darshan-util/darshan-logutils.c
+++ b/darshan-util/darshan-logutils.c
@@ -1579,6 +1579,21 @@ static int darshan_log_dzread(darshan_fd fd, int region_id, void *buf, int len)
             return(-1);
     }
 
+    /* The following is a workaround for a zlib-ng bug observed in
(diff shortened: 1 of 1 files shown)


shanedsnyder (approved):

shanedsnyder:
Looks good to me! I confirmed log reading issues with the original log and that this branch resolves those issues.

PR #1003, opened 2024-08-16
write trailing null byte in mnt table

carns:
- this should make the log format slightly more robust with respect to how the data is decompressed, but does not require any file format change
- protects against problems like #1002, but would only be effective for newly written logs


shanedsnyder (approved):

shanedsnyder:
Looks good to me, thanks!

PR #1005, opened 2024-08-19
Darshan-LDMS Workflow Patch

Snell1224:
This fixes the  libjansson error that was occurring when installing the LDMS from top of tree:
```checking for libjansson... no
configure: error: libjansson or jansson.h not found
```


shanedsnyder (approved):

shanedsnyder:
Looks good to me, thanks @Snell1224 !

PR #1007, opened 2024-09-04
[WIP] MAINT: remove usage of `--no-as-needed` ld flag

shanedsnyder:
Theoretically, this shouldn't be needed if we aren't relying on weak symobls in the Darshan library.


carns (approved):
looks good to me

shanedsnyder:
Eh, nevermind. We have a test case for a scenario that shows this flag is probably still needed. If an application doesn't make calls to functions Darshan wraps, but instead links to a library that does so, then Darshan is not linked in if this flag isn't used. I'll keep the flag and open additional PRs to update documentation to explain this and to add a test case to make sure we don't break this going forward.

PR #1009, opened 2024-09-25
Patch for ldms streams function 

Snell1224:
This patch adds a check for the ``ldms_xprt_new_with_auth`` function in LDMS Streams API. There are currently two versions of this function (one takes three arguments and the other takes four) depending on the LDMS library version. The purpose of this patch is to prevent any build errors caused by version incompatibilities for users wanting to enable LDMS. 

Please let me know if there are issues with these changes.


shanedsnyder (approved):

shanedsnyder:
Looks good to me, thanks for the fix!

PR #1010, opened 2024-10-15
MAINT: stop using deprecated autoconf functions

shanedsnyder:
We've gotten reports of strange errors when configuring `darshan-util` with APMPI support with newer autoconf versions (e.g., 2.72). I could not make sense of the errors, but there were lots of warnings about usage of deprecated functions, so I went ahead and fixed all of those here. Not exactly sure how, but it does seem to have resolved the weird configure error we're getting. Tested using autoconf 2.69 and 2.72 and things all seems good, but let's see how CI goes.


carns (approved):
Looks great; confirmed fix.

PR #1013, opened 2024-10-30
LDMS: fixing lib include

Snell1224:
Wrong header file included in the patch from PR #1009.


shanedsnyder:
Thanks!

Can you also take a look at the current failure in the LDMS workflow? Looks like it's looking for a git tag that no longer exists.

Snell1224:
Oh yes, the git repo url was changed to ldms. Thanks for catching this!

shanedsnyder:
All looks good, thanks for the quick fix!

PR #1019, opened 2025-03-27
ENH: various darshan-convert obfuscate fixes

shanedsnyder:
- independent obfuscation flags to allow more control over what is obfuscated (e.g., to not obfuscate job ids)
- don't obfuscate common names like standard streams (e.g., <STDOUT>) or heatmaps

### DIFF
--- a/darshan-util/darshan-convert.c
+++ b/darshan-util/darshan-convert.c
@@ -30,6 +30,7 @@ int usage (char *exename)
     fprintf(stderr, "       Converts darshan log from infile to outfile.\n");
     fprintf(stderr, "       rewrites the log file into the newest format.\n");
     fprintf(stderr, "       --bzip2 Use bzip2 compression instead of zlib.\n");
+    fprintf(stderr, "       --obfuscate Obfuscate all items in the log.\n");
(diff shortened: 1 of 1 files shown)


darshan-util/darshan-convert.c
```diff
@@ -30,7 +30,10 @@ int usage (char *exename)
     fprintf(stderr, "       Converts darshan log from infile to outfile.\n");
     fprintf(stderr, "       rewrites the log file into the newest format.\n");
     fprintf(stderr, "       --bzip2 Use bzip2 compression instead of zlib.\n");
-    fprintf(stderr, "       --obfuscate Obfuscate items in the log.\n");
```
carns, line 33: Would it be worth keeping --obfuscate for compatibility?  (setting it would implicitly set each of the individual flags, so the comment/usage in that case would be something like "Enable all obfuscate options (jobid, uid, exe, and names)"

carns (changes requested):
Minor suggestion for how to handle the options.  Otherwise looks good.

darshan-util/darshan-convert.c
shanedsnyder: good call, i fixed this by adding back the original option that sets all flags

carns (approved):
Perfect, thanks.

PR #1024, opened 2025-05-05
CI: update upload-artifact version in build-wheels workflow

shanedsnyder:
The `upload-artifact` version we had been using is now deprecated. 

Note that there are changes in the new version that don't allow multiple workflow steps to combine their output into a single artifact archive. Changes have been made to direct output from each step into unique archives, along with a new phase of the workflow that merges all of these together into a single artifact.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -57,6 +57,6 @@ jobs:
         with:
           package-dir: ./darshan-util/pydarshan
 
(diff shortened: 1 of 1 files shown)


shanedsnyder:
@Snell1224 -- could you look into the LDMS workflow error reported here? It looks like maybe LDMS APIs changed?

Thanks!

Snell1224:
@shanedsnyder PR #1025 fixes this issue

PR #1030, opened 2025-05-14
DOC: darshan-runtime user guide in markdown format

wkliao:
To move away from using mcs web server to host Darshan's documents, one
way is to use markdown format. Luckily, both Darshan runtime and util guides
are currently in txt format. It does not take too much effort to convert.

An md file can be directly viewed from the github repo or saved in its [wiki page](https://github.com/wkliao/darshan/wiki).
See it in [darshan-runtime.md](https://github.com/wkliao/darshan/blob/user_guides/darshan-runtime/doc/darshan-runtime.md)

This PR is the first attempt on darshan-runtime guide.
If it looks fine, we can move on to other user guides.

### DIFF
--- a/darshan-runtime/doc/darshan-runtime.md
+++ b/darshan-runtime/doc/darshan-runtime.md
@@ -1,0 +1,871 @@
+# Darshan-runtime installation and usage
+
+Table of Contents
+
+0. [Introduction](#sec_intro)
+1. [Requirements](#sec_req)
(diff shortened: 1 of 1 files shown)


carns:
This looks great to me @wkliao , thanks for finding this option and trying it out.  Lets follow up a little offline.

wkliao:
Close this PR, as we have decided to go with the option of readthedocs (#1034)

PR #1035, opened 2025-05-28
clang20 cleanups

roblatham00:
When I build with clang-20 I get a few warnings.   The "uninitialized variable" and 'don't use K&R syntax' changes are  straightforward enough.  

I'm really puzzled by the `newpath != path` part -- that's one of those "how did it ever work in the first place?" things, I'd like a second set of eyes because there's got to be a reason we were doing it that way since January 2019 (db253374a4304)


wkliao:
I had a quick look at line 231
```
225          __newpath = darshan_clean_file_path(__path); \

231 -      if(__newpath != (char*)__path) free(__newpath); \
231 +      if(strcmp(__newpath, (char*)__path) != 0 ) free(__newpath); \
```
My reading to it is if `__newpath` is newly allocated string by a call
to `darshan_clean_file_path()` at line 225, then free its allocated space.

darshan-runtime/lib/darshan-stdio.c
```diff
@@ -228,11 +228,11 @@ extern int __real_fileno(FILE *stream);
     __rec_ref = darshan_lookup_record_ref(stdio_runtime->rec_id_hash, &__rec_id, sizeof(darshan_record_id)); \
     if(!__rec_ref) __rec_ref = stdio_track_new_file_record(__rec_id, __newpath); \
     if(!__rec_ref) { \
-        if(__newpath != (char*)__path) free(__newpath); \
```
carns, line 231: Oh, now that Wei-keng pointed out, I agree that the original logic was intentional (__newpath may point to __path or be a freshly allocated buffer depending on preceding logic).  There is probably another fix for the clang warning?

carns (changes requested):

wkliao:
Another way is to modify subroutine `darshan_clean_file_path` to make it always to
allocate a new string so it can be freed later without checking. The max length of the
string is `__DARSHAN_PATH_MAX` which is 4096 defined in darshan.h.
If this `__newpath` is only used once, then this approach will not take too much
memory space.

roblatham00:
casting to void didn't fool clang.

some rando on reddit suggests "try to write to the string and catch the segv" -- uh no thank you i'll keep the warning!

carns:
That's not so bad.  @wkliao what do you think?

carns:
Yeah, the h5py problem is unrelated to this PR, I'll open up a separate issue about that.

wkliao (approved):

PR #1036, opened 2025-06-02
Added I/O performance metrics to file_stats

safaad:
- Added performance metrics to file_stats, including: I/O Bandwidth by slowest, I/O time by slowest, and total bytes. 
- Added sorting by the new metrics
- Reformatted the code of file_stats.py for better readability.

### DIFF
--- a/darshan-util/pydarshan/darshan/cli/file_stats.py
+++ b/darshan-util/pydarshan/darshan/cli/file_stats.py
@@ -299,7 +299,7 @@ def setup_parser(parser: argparse.ArgumentParser):
         "--order_by",
         "-o",
         nargs="?",
-        default="bytes_read",
+        default="total_bytes",
         choices=[
             "perf_by_slowest",
             "time_by_slowest",
--- a/darshan-util/pydarshan/darshan/cli/file_stats.py
+++ b/darshan-util/pydarshan/darshan/cli/file_stats.py
@@ -233,6 +233,11 @@ def rich_print(df, mod, order_by):
         "time_by_slowest", f"[u i]{all_time_by_slowest:.2f} s", **default_kwargs
     )
     table.add_column(
+        "total_bytes",
+        f"[u i]{naturalsize(all_total_bytes, binary=True, format='%.2f')}",
+        **default_kwargs,
+    )
+    table.add_column(
         "bytes_read",
         f"[u i]{naturalsize(all_bytes_read, binary=True, format='%.2f')}",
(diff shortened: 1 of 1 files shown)


darshan-util/pydarshan/darshan/cli/file_stats.py
```diff
@@ -141,45 +197,77 @@ def first_n_recs(df, n):
...
-                      f"{row['total_jobs']}")
+        table.add_row(
+            row["file"],
+            f"{naturalsize(row['perf_by_slowest'], binary=True, format='%.2f')}/s",
+            f"{row['time_by_slowest']:.2f} s",
+            f"{naturalsize(row['bytes_read'], binary=True, format='%.2f')}",
+            f"{naturalsize(row['bytes_written'], binary=True, format='%.2f')}",
+            f"{row['reads']}",
+            f"{row['writes']}",
+            f"{naturalsize(row['total_bytes'], binary=True, format='%.2f')}",
```
carns, line 265: Can you update the table order so that the `total_bytes` appears just before the `bytes_read` and `bytes_written`?  I think they make sense to be grouped together since the latter two are subsets of the first.

carns (changes requested):
Thank you for the contribution @safaad !  This is a nice addition; it makes sense to see the I/O performance and I/O time as part of this output.  The other minor cleanups and clarifications are appreciated as well.

I just have one minor refinement to suggest for the table order.

darshan-util/pydarshan/darshan/cli/file_stats.py
safaad: Thank you for the suggestion, @carns! I updated the order

carns:
Thank you!  I apologize for not noting this sooner, but now that I see the output with your new `total_size` column, I wonder if that shouldn't be the default sort_by column.  Testing your PR with this example I noticed that the most interesting files are lower in the list because of the `bytes_written` contribution (see attached screenshot).

What do you think?

Screenshot From 2025-06-04 13-37-27

safaad:
I agree with you — defaulting the sort to ```bytes_read``` might not be the most effective way to represent the data, since the most relevant files are often the ones being written. I used ```bytes_read``` because that was the default in the previous version, but it might make more sense to use ```bytes_written``` or ```total_bytes``` instead. What do you think? What would be better to set as the default?

carns:
Apologies for the delay.  This is perfect, thank you!

PR #1038, opened 2025-06-07
TYP: Disable NumPy's deprecated mypy plugin

tylerjereddy:
* Mirror the upstream change to disable this plugin. For details, see:
https://github.com/scipy/scipy/pull/23123

* I don't think we're running `mypy` in CI anyway, but might as well clean it up in case the team
turns it on at some point (a well-maintained type stub library now exists upstream: https://github.com/numpy/numtype).


orcunyildiz (approved):

orcunyildiz:
Thanks, Tyler!

PR #1033, opened 2025-05-23
DOC: migrate documents to readthedocs

wkliao:
Readthedocs has a nice integration with Github CI. A commit pushed to a branch
of Darshan's repo can trigger a rebuild of all Darshan's documents, which include
* darshan-runtime
* darshan-util
* darshan-modularization
* PyDarshan

A short developer guide is provided in file ./readthedocs/DEVELOPER_NOTES.md

PR #1039, opened 2025-06-09
deprecate wrapper generators

carns:
The Darshan documentation still advocates for generating custom MPI compiler wrapper scripts with Darshan instrumentation.  This method was developed for MPICH 2 and early versions of MPICH 3 and isn't really appropriate for any recent versions of MPICH.   The profile method is far less fragile.

- remove the generators themselves
- refactor the runtime documentation a bit (remove discussion of wrapper generator scripts, reorganize remaining options into subsections in priority of Cray option, MPICH option, then manual option)

Note that this PR conflicts with #1033 probably.  If that one lands first, I can revise this to update the rtd version of the documentation instead.


wkliao:
I can see the only conflict is the changes to file `darshan-runtime.txt`.
Once #1033 is merged, I can push a commit to this PR to resolve the conflict.

carns:
I think this is up to date now.  I'll open a separate PR to remove the old .txt documentation files; they aren't needed now that we have fully migrated to RTD.

carns:
> File `darshan-runtime/Makefile.am` contains the followings lines that need to be removed. They cause Github CI actions to fail.
> 
> ```
>               darshan-gen-cc.pl \
>               darshan-gen-cxx.pl \
>               darshan-gen-fortran.pl \
> ```

Thanks!  Fixed now.  I also rebased to account for the removal of the .txt documents in the other PR.

PR #1043, opened 2025-06-10
revise landing page

carns:
- reorder toc
- add toc references in top bullet list
- remove stale facility documentation and replace with a more general sugestion to refer to facility documentation first


wkliao (approved):
Look fine with me.

Issue #1044, opened 2025-06-11
runtime configuration table is hard to read in readthedocs


carns:
See https://darshan.readthedocs.io/en/latest/darshan-runtime/doc/darshan-runtime.html#darshan-library-config-settings

I don't know if this is browser-specific, but in Firefox on Linux I see the following:

![Image](https://github.com/user-attachments/assets/87c5487b-c72d-40cc-8332-ca9d30672f8e)

This happens regardless of page zoom level (the margin scales with the page and the column remains truncated).

carns:
Oh I see, there is a scroll bar for the table itself.  Is there any easy way to improve the default visibility?  The layout is only using a fraction of my browser window width :)

wkliao:
fixed in #1046

PR #1046, opened 2025-06-12
experimenting with modifying css for table

carns:
Trying to see if a custom css can be used to help with #1044

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -129,3 +129,8 @@ intersphinx_mapping = {
   "matplotlib": ("https://matplotlib.org", None),
 }
 
+html_static_path = ['readthedocs/_static']
+html_css_files = [
+    'css/custom.css',
+]
+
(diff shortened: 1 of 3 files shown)


wkliao:
I think we can use the css approach.
My current implementation adds an empty line to achieve the effect of line break.
It is ugly. See codes below
https://github.com/darshan-hpc/darshan/blob/277c6d3e23f2adde3cbf39324c2a345c1122b5f7/darshan-runtime/doc/darshan-runtime.rst?plain=1#L615-L617

I think using css should do this better. I will make use of this PR to do more testings.

carns:
Ok!  If you find a better solution we can close this PR and merge something cleaner.

Along the way, I think we can remove the old *.txt and *.css files from the doc directories now that we have switched over?

carns (commented):
This looks great to me.  It avoids wasting all of that whitespace when you are viewing the page on a computer screen, and it still adds horizontal scroll bars to the table if your browser is not wide enough.

wkliao:
Please squash and merge. Thanks.

Issue #1052, opened 2025-07-14
Processes being abruptly killed in the window of time between the start of shutdown and the shutdown completing


mikail-g:
I'm running darshan with an application launched by deepspeed, which launches subprocesses to carry out some checkpointing I/O.

When the application completes, darshan only gets partway through core_shutdown before it gets killed, thus losing the desired logs in the process.

As a test we edited darshan_core_shutdown in darshan-core to immediately return instead of running any of it's code, after which we can see the desired logs in /tmp and they are in a valid state to be parsed with darshan-parser. 

We are using darshan 3.4.7

I include some screenshots below of some outputs after execution with these two methods:
### Unedited darshan-core-shutdown:
```
`[DEEPSPEED DEBUG] Darshan is loaded: **/tmp/mgossman_python_id1382732_mmap-log-7779381447136745040-0.darshan**
[DEEPSPEED DEBUG] Darshan is loaded: /home/mgossman/restart_perf/software/installs/darshan/lib/libdarshan.so.0.0.0
[DEEPSPEED DEBUG] Darshan is loaded: /home/mgossman/restart_perf/software/installs/darshan/lib/libdarshan.so.0.0.0
[DEEPSPEED DEBUG] Darshan is loaded: /home/mgossman/restart_perf/software/installs/darshan/lib/libdarshan.so.0.0.0
[2025-07-14 19:27:22,111] [INFO] [torch_checkpoint_engine.py:49:save] [Torch] Saving /grand/VeloC/mikailg/DeepSpeed-restart-perf/modelsize1_tp1_pp1_dp4/global_step1/bf16_zero_pp_rank_0_mp_rank_00_optim_states.pt...
[2025-07-14 19:27:45,907] [INFO] [torch_checkpoint_engine.py:51:save] [Torch] Saved /grand/VeloC/mikailg/DeepSpeed-restart-perf/modelsize1_tp1_pp1_dp4/global_step1/bf16_zero_pp_rank_0_mp_rank_00_optim_states.pt.
[2025-07-14 19:27:45,911] [INFO] [engine.py:3701:_save_zero_checkpoint] zero checkpoint saved /grand/VeloC/mikailg/DeepSpeed-restart-perf/modelsize1_tp1_pp1_dp4/global_step1/bf16_zero_pp_rank_0_mp_rank_00_optim_states.pt
[...]

PR #1053, opened 2025-07-15
Duplicate mmap log file at the beginning of darshan_core_shutdown

wkliao:
This is in case an interrupt occurs while processing the log file, resulting the log file corrupted.
See issue #1052.

### DIFF
--- a/darshan-runtime/lib/darshan-core.c
+++ b/darshan-runtime/lib/darshan-core.c
@@ -433,6 +433,10 @@ void darshan_core_shutdown(int write_log)
     darshan_record_id *mod_shared_recs = NULL;
     int shared_rec_cnt = 0;
 #endif
+#ifdef __DARSHAN_ENABLE_MMAP_LOGS
+    char dup_log_fame[__DARSHAN_PATH_MAX + 5];
+    dup_log_fame[0] = '\0';
+#endif
 
     /* disable darhan-core while we shutdown */
     __DARSHAN_CORE_LOCK();
@@ -470,13 +474,41 @@ void darshan_core_shutdown(int write_log)
     internal_timing_flag = final_core->config.internal_timing_flag;
 
 #ifdef __DARSHAN_ENABLE_MMAP_LOGS
-    /* remove the temporary mmap log files */
-    /* NOTE: this unlink is not immediate as it must wait for the mapping
-     * to no longer be referenced, which in our case happens when the
-     * executable exits. If the application terminates mid-shutdown, then
-     * there will be no mmap files and no final log file.
+    /* Flush memory-mapped data to the underlying file and then duplicate the
+     * mmap log file, in case an interrupt happens before the completion of
+     * this subroutine, leaving the log file corrupted. See github issue #1052.
      */
-    unlink(final_core->mmap_log_name);
+    int sys_page_size = sysconf(_SC_PAGESIZE);
+    size_t mmap_size = sizeof(struct darshan_header) + DARSHAN_JOB_RECORD_SIZE +
+        + core->config.name_mem + core->config.mod_mem;
+    if (mmap_size % sys_page_size)
+        mmap_size = ((mmap_size / sys_page_size) + 1) * sys_page_size;
+
+    msync(final_core->log_hdr_p, mmap_size, MS_SYNC);
+
+    /* duplicate the log file */
+    int mmap_fd;
+    mmap_fd = open(final_core->mmap_log_name, O_RDONLY, 0644);
+    if (mmap_fd != -1) {
+        void *buf;
+        off_t fileSize = lseek(fd, 0, SEEK_END);
+        if (fileSize >= 0) {
+            buf = (void*) malloc(fileSize);
+            lseek(fd, 0, SEEK_SET);
+            read(mmap_fd, fileSize, buf);
+            close(mmap_fd);
+            snprintf(dup_log_fame, "%s.dup", strlen(final_core->mmap_log_name),
+                     final_core->mmap_log_name);
+            mmap_fd = open(dup_log_fame, O_CREAT | O_WRONLY, 0644);
+            if (mmap_fd != -1) {
+                write(mmap_fd, fileSize, buf);
+                close(mmap_fd);
+            }
+            free(buf);
+        }
+        else
+            close(mmap_fd);
+    }
 #endif
 
     final_core->comp_buf = malloc(final_core->config.mod_mem);
@@ -770,7 +802,15 @@ cleanup:
     for(i = 0; i < DARSHAN_KNOWN_MODULE_COUNT; i++)
         if(final_core->mod_array[i])
             final_core->mod_array[i]->mod_funcs.mod_cleanup_func();
+
+#ifdef __DARSHAN_ENABLE_MMAP_LOGS
+    /* remove the temporary mmap log files */
+    unlink(final_core->mmap_log_name);
+    if (dup_log_fame[0] != '\0') unlink(dup_log_fame);
+#endif
+
     darshan_core_cleanup(final_core);
+
 #ifdef HAVE_MPI
     if(using_mpi)
     {
@@ -778,6 +818,7 @@ cleanup:
         free(mod_shared_recs);
     }
 #endif
+
     free(logfile_name);
 
     return;


darshan-runtime/lib/darshan-core.c
```diff
@@ -470,13 +474,41 @@ void darshan_core_shutdown(int write_log)
...
-    /* NOTE: this unlink is not immediate as it must wait for the mapping
-     * to no longer be referenced, which in our case happens when the
-     * executable exits. If the application terminates mid-shutdown, then
-     * there will be no mmap files and no final log file.
+    /* Flush memory-mapped data to the underlying file and then duplicate the
+     * mmap log file, in case an interrupt happens before the completion of
+     * this subroutine, leaving the log file corrupted. See github issue #1052.
      */
-    unlink(final_core->mmap_log_name);
+    int sys_page_size = sysconf(_SC_PAGESIZE);
```
carns, line 481: This code block uses two methods to calculate the size, one based on Darshan parameters and sizes, and one based on the file size (via lseek).  It seems like we could just do one or the other and reuse it for both purposes?  Of the two I would probably vote for just using the file size (lseek end position) so that nothing has to be updated here if the Darshan file format changes.

darshan-runtime/lib/darshan-core.c
```diff
@@ -470,13 +474,41 @@ void darshan_core_shutdown(int write_log)
...
+        if (fileSize >= 0) {
+            buf = (void*) malloc(fileSize);
+            lseek(fd, 0, SEEK_SET);
+            read(mmap_fd, fileSize, buf);
+            close(mmap_fd);
+            snprintf(dup_log_fame, "%s.dup", strlen(final_core->mmap_log_name),
+                     final_core->mmap_log_name);
+            mmap_fd = open(dup_log_fame, O_CREAT | O_WRONLY, 0644);
+            if (mmap_fd != -1) {
+                write(mmap_fd, fileSize, buf);
```
carns, line 504: we probably need to check the return code of the write here as well.  If it fails we just do a best effort cleanup (unlink etc.).

darshan-runtime/lib/darshan-core.c
```diff
@@ -470,13 +474,41 @@ void darshan_core_shutdown(int write_log)
...
+                close(mmap_fd);
+            }
```
carns, line 506: if the open fails then we should probably clear the dup file name since it will be used later to decide if we need to unlink.

darshan-runtime/lib/darshan-core.c
```diff
@@ -770,14 +802,23 @@ void darshan_core_shutdown(int write_log)
     for(i = 0; i < DARSHAN_KNOWN_MODULE_COUNT; i++)
         if(final_core->mod_array[i])
             final_core->mod_array[i]->mod_funcs.mod_cleanup_func();
+
```
carns, line 805: I don't know if it makes much difference, but if we wanted to we could move this a smidge earlier.  I think it would be safe immediately after darshan_log_finalize().

carns (changes requested):
This looks good, just a few minor suggestions

darshan-runtime/lib/darshan-core.c
wkliao: Not sure which smidge you meant. Lines 805 to 810, unlink calls?
Do you mean moving them to after line 725?

carns: Oh yeah, I meant moving the final unlink (for the "dup" file) from it's current position to instead happen right after darshan_log_finalize().  The things in between that happen in between there and the current position don't have any impact on log output, so we can slightly reduce the window of time were we could leave a stray tmp file.

darshan-runtime/lib/darshan-core.c
carns: ah ok.  that's fine then.

wkliao:
I also notice that no CI is testing the configure option `--enable-mmap-logs`.
Maybe we should add one.

darshan-runtime/lib/darshan-core.c
wkliao: It is done.

darshan-runtime/lib/darshan-core.c
wkliao: Calling `msync()` before `lseek()` is to ensure all pending writes are flushed to
the mmap file, so we can then safely call `lseek()` to obtain the file size.

To avoid calculation of `mmap_size` in multiple places, a solution can be to save
`mmap_size` as a member of `struct darshan_core_runtime`.

wkliao:
The solution proposed in this PR only works when `--enable-mmap-logs` is set at configure time.
When this is not the case, we may need another solution.

carns:
> I also notice that no CI is testing the configure option `--enable-mmap-logs`. Maybe we should add one.

Yeah we really should.

darshan-runtime/lib/darshan-core.c
wkliao: So, removing the call to `lseek` and just use `mmap_size` to read and write?

carns: I'm advocating for removing the mmap_size calculation.  I would be fine if we *just* used the size reported by lseek (both to determine how much to flush, and to determine how much to copy), or else added a field to the runtime structure to keep up with the size.

carns (dismissed):
This looks great.  Let's see if Mikaila can test it.  If it works for her then this can be merged.

darshan-runtime/lib/darshan-core.c
```diff
@@ -470,13 +474,34 @@ void darshan_core_shutdown(int write_log)
...
+     * of duplicated file will become useless, but such problem will be even
+     * more serious than being unable to duplicate the file.
+     */
+    int mmap_fd = open(final_core->mmap_log_name, O_RDONLY, 0644);
+    if (mmap_fd != -1) {
+        int dup_mmap_fd;
+        void *buf = (void*) malloc(final_core->mmap_size);
+        read(mmap_fd, buf, final_core->mmap_size);
+        close(mmap_fd);
+        snprintf(dup_log_fame, strlen(final_core->mmap_log_name), "%s.dup",
```
carns, line 496: This snprintf truncates the files name to something like `/tmp/carns_python3_id1445957_mmap-log-3779006568372861245-0.darsha` because the 2nd argument to is calling `strlen()` on the original `mmap_log_name`.  Maybe this size limit on the snprintf should be based on the size of the `dup_log_fname` instead?

darshan-runtime/lib/darshan-core.c
```diff
@@ -470,13 +474,34 @@ void darshan_core_shutdown(int write_log)
...
+     * not affect the remaining tasks of this subroutine, i.e. log data file
+     * processing. Note if any of these system calls failed, then the contents
+     * of duplicated file will become useless, but such problem will be even
+     * more serious than being unable to duplicate the file.
+     */
+    int mmap_fd = open(final_core->mmap_log_name, O_RDONLY, 0644);
+    if (mmap_fd != -1) {
+        int dup_mmap_fd;
+        void *buf = (void*) malloc(final_core->mmap_size);
+        read(mmap_fd, buf, final_core->mmap_size);
```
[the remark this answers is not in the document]
wkliao: I ignored them on purpose, as these codes are for file duplication, wouldn't
want any error stop the duplication.

Anyway, I added checks in 14a755c.
Strangely, I also run gcc 14.2.1, but did not get those warnings.
Please see if they are silenced.

darshan-runtime/lib/darshan-core.c
wkliao: This should be fixed in 6a5c60f.

wkliao:
I have a doubt this issue may be related to the data sieving bug in OpenMPI in #1064.
Is it possible to get @mikail-g to confirm whether or not she is using OpenMPI, etc.?

darshan-runtime/lib/darshan-core.c
carns: Actually I have one other request here.  Sorry this didn't occur to me sooner.  I like that the duplicate file has "dup" in the file name, but it may be better if it weren't used as the file extension.  I find myself doing things like `ls /tmp/*.darshan` to look for intermediate files, when with this patch it would really need to be `ls /tmp/*.darshan /tmp/*.darshan.dup` to find them all.

Can we change the naming convention so that even the duplicate file names end with `.darshan`?

wkliao: Updated. Please give it a try.

PR #1057, opened 2025-08-01
Update CHANGELOG for pydarshan 3.4.7.1 release

orcunyildiz:


carns:
> Version 3.4.7.1 has been taken. Check the list of all Darshan [tags](https://github.com/darshan-hpc/darshan/tags).
> 
> I suggest to avoid using the fourth digit, as it is not a standard practice. See the guide line for setting the version number in https://semver.org Even if Darshan did not follow it in the past, we can start doing so now.

Ok.  Let's pause and discuss more today.  I also don't like the idea of PyDarshan running ahead to 3.4.8 without the other portions of Darshan, since we would want that version number for the next release of the runtime and tools.  We either need to find an acceptable way to differentiate pydarshan or else release them all together.

PR #1059, opened 2025-08-04
Add 4 missing MPI-IO APIs

wkliao:
* MPI_File_iread_all
* MPI_File_iread_at_all
* MPI_File_iwrite_all
* MPI_File_iwrite_at_all

### DIFF
--- a/darshan-runtime/lib/darshan-mpiio.c
+++ b/darshan-runtime/lib/darshan-mpiio.c
@@ -753,7 +753,7 @@ int DARSHAN_DECL(MPI_File_read_all_begin)(MPI_File fh, void * buf, int count, MP
 
     MAP_OR_FAIL(PMPI_File_read_all_begin);
 
-    MPI_File_get_position_shared(fh, &offset);
+    MPI_File_get_position(fh, &offset);
     tm1 = MPIIO_WTIME();
     ret = __real_PMPI_File_read_all_begin(fh, buf, count, datatype);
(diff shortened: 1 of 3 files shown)


darshan-runtime/lib/darshan-mpiio.c
```diff
@@ -926,6 +930,28 @@ int DARSHAN_DECL(MPI_File_iread)(MPI_File fh, void * buf, int count, MPI_Datatyp
...
 
+int DARSHAN_DECL(MPI_File_iread_all)(MPI_File fh, void * buf, int count, MPI_Datatype datatype, __D_MPI_REQUEST * request)
+{
+    int ret;
+    double tm1, tm2;
+    MPI_Offset offset;
+
+    MAP_OR_FAIL(PMPI_File_iread_all);
+
+    MPI_File_get_position_shared(fh, &offset);
```
carns, line 941: Should we be using the regular (non-shared) `MPI_File_get_position()` here?  It look like this function operates on an individual file pointer rather than a shared file pointer per https://www.mpich.org/static/docs/v3.2/www3/MPI_File_iread_all.html .

If my interpretation is right, then we probably need to check all of the wrappers that use those functions (not just the new ones) to make sure we are using the regular version or the shared version in the right places.

carns (changes requested):

carns (approved):
thanks!

Issue #1020, opened 2025-04-07
support new MPI "large count" routines


roblatham00:
From slack:  

```
On using the latest IOR version with MPICH(4.3.2), the actual MPIIO calls don’t occur in the DXT log.
MPIIO calls appear in the darshan log but things like MPIIO_BYTES_READ are zero
We think the issue might be the series of _C MPI(introduced with MPI 4.0) (eg. MPI_File_write_at_c) functions used by IOR; darshan may not be wrapping them.
(We didn’t have this issue with OpenMPI; it doesn’t have the _c functions, IOR just wraps the original set of MPI functions with _c suffix) (edited) 
```

recently MPI added "large count" support, but they didn't change the old routines that take 32 bit int.   They added a bunch of new _c routines that take MPI_Count, a 64 bit (or maybe 128 bits on some platforms... something big)  value.  

These routines behave in every way like the traditional routines but are new symbols and Darshan will have to wrap them.

carns:
Just to clarify, is IOR explicitly calling the `_C` variants?

wkliao:
This may take some time.
Here is a list of MPI-IO large-count APIs, i.e. name ended with "_c".
```
int MPI_File_read_c(MPI_File fh, void *buf, MPI_Count count, MPI_Datatype datatype, MPI_Status *status)
int MPI_File_read_all_c(MPI_File fh, void *buf, MPI_Count count, MPI_Datatype datatype, MPI_Status *status)
int MPI_File_read_all_begin_c(MPI_File fh, void *buf, MPI_Count count, MPI_Datatype datatype)
int MPI_File_read_at_c(MPI_File fh, MPI_Offset offset, void *buf, MPI_Count count, MPI_Datatype datatype,
int MPI_File_read_at_all_c(MPI_File fh, MPI_Offset offset, void * buf, MPI_Count count,
int MPI_File_read_at_all_begin_c(MPI_File fh, MPI_Offset offset, void *buf, MPI_Count count,
int MPI_File_read_ordered_c(MPI_File fh, void *buf, MPI_Count count, MPI_Datatype datatype,
int MPI_File_read_ordered_begin_c(MPI_File fh, void *buf, MPI_Count count, MPI_Datatype datatype)
int MPI_File_read_shared_c(MPI_File fh, void *buf, MPI_Count count, MPI_Datatype datatype,
int MPI_File_write_c(MPI_File fh, const void *buf, MPI_Count count, MPI_Datatype datatype,
[...]

PR #1060, opened 2025-08-04
support MPI-IO large count APIs

wkliao:
This PR address #1020


darshan-runtime/lib/darshan-mpiio.c
```diff
@@ -1899,6 +1899,688 @@ static void mpiio_cleanup()
...
+int DARSHAN_DECL(MPI_File_iread_shared_c)(MPI_File fh, void * buf, MPI_Count count,
+    MPI_Datatype datatype, __D_MPI_REQUEST * request)
+{
+    int ret;
+    double tm1, tm2;
+    MPI_Offset offset;
+
+    MAP_OR_FAIL(PMPI_File_iread_shared_c);
+
+    MPI_File_get_position(fh, &offset);
```
carns, line 2034: This one should be a `MPI_File_get_position_shared()`, right?

darshan-runtime/lib/darshan-mpiio.c
```diff
@@ -1899,6 +1899,688 @@ static void mpiio_cleanup()
...
+int DARSHAN_DECL(MPI_File_iwrite_all_c)(MPI_File fh, const void * buf, MPI_Count count,
+    MPI_Datatype datatype, __D_MPI_REQUEST * request)
+{
+    int ret;
+    double tm1, tm2;
+    MPI_Offset offset;
+
+    MAP_OR_FAIL(PMPI_File_iwrite_all_c);
+
+    MPI_File_get_position(fh, &offset);
```
carns, line 2059: Same here, this seems like it would rely on the shared position.

carns (changes requested):
A couple of notes to address, otherwise it looks good to me.

Kind of wild how long the static options list is now.  At some point we should think about if its even worth maintaining static linking support at this point in time, but we shouldn't worry about it yet.

darshan-runtime/lib/darshan-mpiio.c
wkliao: Thanks for catching this error.

carns (dismissed):
Looks good to me

wkliao:
I added an exhaust test for all 56 MPI file read and write APIs.
At this moment, it only checked field "MPIIO_BYTES_WRITTEN"
(the total amount written) for write from the Darshan log. Similarly,
"MPIIO_BYTES_READ" for read. We can expand this test for other
fields, e.g. offset.

@roblatham00 , Maybe test like this can also be considered for ROMIO, if
it does not have one yet.

carns:
Something is a little weird in the CI setup.  IIUC browsing the github test output, it looks like this is done in the "Install Darshan" step, which is shared by several actions (sorry I may not have the terminology quite right).  It looks like this make check test fails in some of the actions (like the LDMS one and the end-to-end pytest) but it succeeds in the end-to-end regression (the error message is about lack of "slots" in the github environment).  Will the latter action fail if the make check fails there?

I don't know how hard it would be, but maybe the "make check" part of that step should be split out to be separate from darshan installation so that it is easier to inspect separately?

Sorry to ask for more work here, but this test is really helpful; we should follow this example as best practice in the future.

wkliao:
I am now seeing the following errors and have no clue how to fix it.
```
Error: unable to inflate darshan log data.
Error: failed to read name hash from darshan log file.
```
The actual error code returned from inflate() is Z_DATA_ERROR

https://github.com/darshan-hpc/darshan/blob/0ad393396ed556d3f9a42ecb74ac2ed64524fa37/darshan-util/darshan-logutils.c#L1709

wkliao:
I notice these error messages are also the ones shown in #1052.
Not sure whether it is related.
```
Error: unable to inflate darshan log data.
Error: failed to read name hash from darshan log file.
```

After some diggings, I found this error only happens when using OpenMPI
and only on Github action CI. I could not reproduce it on my local machine
running Redhat or compute nodes on cels.anl.gov running Ubuntu.
Fail runs all have the error code `Z_DATA_ERROR`, which means the
Darshan log file is corrupted.

In commit e6a305d, I added a yaml file to build both MPICH and OpenMPI
and use both to run "make check". Only OpenMPI failed and the failure only
happened when running 4 MPI processes. It ran OK for 2.

The next step I tried is to set MPI-IO hint cb_nodes to 1 when configuring
 Darshan (Darshan's default is to set hint cb_nodes to 4). In this case,
"make check" running the test program on 4 MPI processes did not fail.
This makes me think maybe Darshan has an issue with OpenMPI's MPI-IO ...?

darshan-runtime/test/tst_runs.sh
```diff
@@ -0,0 +1,278 @@
+#!/bin/bash
+
+# Exit immediately if a command exits with a non-zero status.
+set -e
```
roblatham00, line 4: 'set -eu' would have caught the typo below

darshan-runtime/test/tst_runs.sh
```diff
@@ -0,0 +1,278 @@
...
+
+TODAY_DATE_PATH=`date "+%Y/%-m/%-d"`
+TST_DARSHAN_LOG_PATH="${TST_DARSHAN_LOG_PATH}/${TODAY_DATE_PATH}"
+mkdir -p ${TST_DARSHAN_LOG_PATH}
+
+if test "x$USERNAME_ENV" = xno ; then
+   USERNAME_ENV=$USER
+fi
+
+DARSGAN_PARSER=../../darshan-util/darshan-parser
```
roblatham00, line 14: that's clearly a typo, right?

darshan-runtime/test/tst_runs.sh
```diff
@@ -0,0 +1,278 @@
...
+          if test "x$opt" = x ; then
+             CMD="${TESTMPIRUN} -n ${NP} ./$exe $TEST_FILE"
+          else
+             CMD="${TESTMPIRUN} -n ${NP} ./$exe -$opt $TEST_FILE"
+          fi
+          # echo "CMD=$CMD"
+          rm -f $TEST_FILE $DARSHAN_LOG_FILE
+          $CMD
+          EXPECT_NBYTE=`stat -c %s $TEST_FILE`
+          nbytes=`$DARSGAN_PARSER ${DARSHAN_LOG_FILE} | grep $DARSGAN_FIELD | cut -f5`
```
roblatham00, line 250: huh.. ok so not a typo? what is this 'DARSGAN' convention?

darshan-runtime/test/tst_mpi_io.c
roblatham00, line 1: this is indeed exhaustive wow.  I don't see any problems with it

darshan-runtime/test/tst_runs.sh
wkliao: It is a typo. Just the same typo is used in lines below.

wkliao:
One question about the default MPI-IO hints set at darshan-runtime configure time,
"romio_no_indep_rw=true;cb_nodes=4"
https://github.com/darshan-hpc/darshan/blob/0fc39110750e9e7247c16106bc7eb88acb0e0c7f/darshan-runtime/configure.ac#L601

Why Darshan sets these default hints? I expect the log files are small after compression.
Setting hints would not matter for most of the cases.

It appears that the OpenMPI (5.0.8 used in my tests) does not take these hints, if using its own
I/O component (now default). OpenMPI-IO hints are listed in
https://docs.open-mpi.org/en/v5.0.6/man-openmpi/man3/MPI_File_open.3.html#hints
Its mpiexec also takes command-line option through `-- mca` to select/set a component
and hint such as `io_ompio_num_aggregators`.

I suspect setting those ROMIO hints is the reason I am seeing the failure of `make check`.
Consider removing them?

roblatham00:
Ah it's because these log files are small that we set the hints.  "no_indep_rw" is how we get "deferred open" (https://wordpress.cels.anl.gov/romio/2003/08/05/deferred-open/ ) so we're asking for only 4 processes to open and write the darshan log file

OpenMPI must ignore hints it does not understand.   sounds like a bug in OpenMPI-IO.   A small write from many processes should be easy for OpenMPI's MPI-IO implementation to handle.

carns:
Ugh.  Ok.  We should be able to create a standalone reproducer (not using Darshan, just a C program that sets hints and makes similar `MPI_File*` calls) in https://github.com/darshan-hpc/darshan/issues/1064.  Once we have done so we should open a bug report with OpenMPI for help.

In the mean time, is there a reliable/safe way to detect OpenMPI in Darshan and alter our default hints?

roblatham00:
sounds like OMPIO doesn't know how to handle "cb_nodes" more than number of processes.   should be an easy fix for them,  but what a headache for us to deal with.     Phil, the reproducer should be prett simple:  take ROMIO's `coll_test` and set `cb_nodes` to nprocs*2

carns:
> sounds like OMPIO doesn't know how to handle "cb_nodes" more than number of processes. should be an easy fix for them, but what a headache for us to deal with. Phil, the reproducer should be prett simple: take ROMIO's `coll_test` and set `cb_nodes` to nprocs*2

My read of https://github.com/darshan-hpc/darshan/pull/1060#issuecomment-3181036369 is that the problem happens with 4 procs and cb_nodes=4?

wkliao:
I have been trying to write a reproducer on a local machine, but could not do so.
The errors only happened on GitHub CI.

When I developed the test program, I simply set the number of MPI processes to 4.
Because it failed in Github CI, I changed it to 2. See line 18 in darshan-runtime/test/tst_runs.sh
I will change it back to 4, so we all can see it failed in the CI report.

(I did add `--oversubscribe` to OpenMPI's mpiexec command line, so I can run more
than 2 processes.)

wkliao:
What is the data partitioning pattern when Darshan writes log data into the file?
Is it appending one process after another based on the MPI rank order?

wkliao:
Like using Darshan to profile Darshan 😄

wkliao:
Inside of darshan-core.c, shouldn't `PMPI_File_open` be called instead of `MPI_File_open`?
Or it does not matter?
https://github.com/darshan-hpc/darshan/blob/0fc39110750e9e7247c16106bc7eb88acb0e0c7f/darshan-runtime/lib/darshan-core.c#L1691

carns:
> Inside of darshan-core.c, shouldn't `PMPI_File_open` be called instead of `MPI_File_open`? Or it does not matter?
> 
> https://github.com/darshan-hpc/darshan/blob/0fc39110750e9e7247c16106bc7eb88acb0e0c7f/darshan-runtime/lib/darshan-core.c#L1691

Good catch. For Darshan itself I don't believe it matters; we don't use PMPI anymore for MPI-IO calls.  We intercept MPI functions the same as we do any other function, with the only variation being that we set up symbol aliases so that the same wrapper will intercept the function whether the caller used the `MPI_*` or `PMPI_*` convention (this is important because some library implementations may use the latter explicitly in bindings for other languages, so we have to catch the `PMPI_*` versions to get any instrumentation at all in those cases).  Maybe for consistency the Darshan code should use all one or the other internally though.

PR #1061, opened 2025-08-05
release checklist: add versioning guide line

wkliao:
It appears that Darshan does not use the libtool ABI version.
Is it because we do not expect user applications directly make
calls to Darshan's APIs? (or there is no public APIs in Darshan?)


carns:
> It appears that Darshan does not use the libtool ABI version. Is it because we do not expect user applications directly make calls to Darshan's APIs? (or there is no public APIs in Darshan?)

I think it is just an omission rather than an explicit decision.

darshan-runtime does not traditionally expose any explicit ABI's to be versioned.

darshan-util does expose an ABI by way of the darshan-logutils library interface.

I don't know how big of a deal it would be to begin properly versioning the ABI.  We probably should, but I don't know that it needs to be a high priority.

carns (approved):

PR #1067, opened 2025-08-17
Tst pthread

wkliao:
branch out of snyder/dxt-extra-info-pthread-id
Test by removing commit of `Add "extra info" field to the records`


wkliao:
Revert commit of `Add "extra info" field to the records` does not resolved the problem. Close this ticket.

PR #1070, opened 2025-08-19
Gihub CI: demonstrate fails when using OpenMPI 5.0.5

wkliao:
After several tests using different versions of OpenMPI, it appears that
5.0.6 fixed the data sieving problem we suspect. This PR demonstrates
it by building Darshan with both 5.0.5 and 5.0.6. Only 5.0.5 failed.

The small test MPI program `mpi_file_write.c` calls only `MPI_File_write()`
once, running on 4 MPI processes. when using OpenMPI 5.0.5, the test
program ran fine, but the contents of generated Darshan log file were corrupted.

According to [OpenMPI 5.0.6 Release note](https://docs.open-mpi.org/en/v5.0.6/release-notes/changelog/v5.0.x.html):
* Detailed Locking Protocol: Modified default file-locking protocols in
  UFS component to ensure data consistency, especially when using
  data-sieving operations, which require broader locking.

Most likely relevant fix in OpenMPI 5.0.6 is https://github.com/open-mpi/ompi/pull/12759


wkliao:
Close this for now. It can be reopened later to test the future release of OpenMPI.

PR #1071, opened 2025-08-21
Set default MPI-IO hint cb_nodes to 1 when using OpenMPI 5.0.5 and priors

wkliao:
OpenMPI 5.0.5 and priors contain a bug that can corrupt the
Darshan log files. The bug is related to file locking protocols
implemented when data sieving is enabled. Setting hint cb_nodes
to 1 can avoid the bug, but may make writing log files slow.
For more information, see PR #1070,

### DIFF
--- a/darshan-runtime/configure.ac
+++ b/darshan-runtime/configure.ac
@@ -701,14 +701,29 @@ if test "x$enable_darshan_runtime" = xyes ; then
       fi
 
       if test "x$GOT_LOG_HINTS" != xyes ; then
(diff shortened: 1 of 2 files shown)


wkliao:
In ad80b29, the default I/O hint has been changed to just `cb_nodes=1`
when OpenMPI is detected.

carns (dismissed):
looks good to me.  I see that some CI actions failed, but I don't know if it is related to these changes or not.

carns:
The `--log-path-by-env=` option specifies a flat directory (it will not traverse a year/month/day hierarchy from there)

PR #1076, opened 2025-09-24
Strictness fix for new compilers

bnicolae:
Newer compilers (e.g. gcc 15+) enforce extern strictness. Use full definitions to pass strictness checks.


wkliao:
Thanks for catching this.

wkliao (approved):

PR #1079, opened 2025-09-30
Setting --with-pnetcdf automatically enables enable_pnetcdf_mod

wkliao:
When Darshan is configured using `--with-pnetcdf=DIR` without
`--enable-pnetcdf-mod`, the pnetcdf module will not be built.
However, in this case, it is clear that the user's intent is to enable pnetcdf.

This PR changes the behavior to automatically sets `--enable-pnetcdf-mod`
when `--with-pnetcdf=DIR` is used at the configure time.

Same for modules HDF5, DAOS, and LDMS.

### DIFF
--- a/darshan-runtime/configure.ac
+++ b/darshan-runtime/configure.ac
@@ -603,13 +603,38 @@ if test "x$enable_darshan_runtime" = xyes ; then
          [], [AC_MSG_ERROR([mdhim requested but headers cannot be found])])
    fi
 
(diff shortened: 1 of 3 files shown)


wkliao:
Hi, @carns 

There are 2 additional ones, `--with-daos=DIR` and `--with-ldms=DIR`.
I wonder if it was designed that way and I am missing something here.

PR #1084, opened 2025-10-04
DOC: add missing configure options to user guides

wkliao:
* `--disable-darshan-util`
* `--disable-darshan-runtime`
* `--enable-pydarshan`
* `--with-zlib=DIR`
* `--without-zlib`
* `--with-bzlib=DIR`
* `--without-bzlib`
* `--with-python-sys-prefix`
* `--with-python_prefix`
* `--with-python_exec_prefix`

Also, modify `darshan-test/automated/build-darshan.sh` to build from the root folder.

### DIFF
--- a/darshan-runtime/doc/darshan-runtime.rst
+++ b/darshan-runtime/doc/darshan-runtime.rst
@@ -79,31 +79,51 @@ Compilation
 
 **Configure and build example (with MPI support)**
 
- .. code-block:: bash
+* When using the release tar ball:
 
-    tar -xvzf darshan-<version-number>.tar.gz
-    cd darshan-<version-number>/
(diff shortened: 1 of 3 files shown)


darshan-runtime/doc/darshan-runtime.rst
```diff
@@ -83,10 +83,8 @@ Compilation
 
     tar -xvzf darshan-<version-number>.tar.gz
     cd darshan-<version-number>/
-    ./prepare.sh
```
carns, line 86: Maybe somewhere we should have a note for people that clone from git that they need to run `./prepare.sh`?  These instructions are correct for the release tarballs.

carns (approved):

PR #1085, opened 2025-10-06
Add pnetcdf lib to --dyn-ld-flags

wkliao:
Add the missing pnetcdf library path, which fixes the linking error below.
```
mpicc -Wall -g -O2 -I/usr/include   hello-mpiio.c   -o hello-mpiio
/usr/bin/ld: /lus/flare/projects/alcf_training/io-libraries/soft/darshan-3.5.0/lib/libdarshan.so: undefined reference to `ncmpi_inq_vardimid'
```


roblatham00 (approved):
Confirmed correct on Aurora

Issue #964, opened 2023-11-03
BUG: darshan-runtime build failing with `--enable-apmpi-coll-sync` config option


shanedsnyder:
This particular configuration isn't used commonly, and it looks like it's tripping some build errors (see https://github.com/spack/spack/issues/40835).

The fix should be relatively simple, probably have just forgotten to apply some changes to interactions with darshan-core that were made at some point (and which have been properly applied to the vanilla APMPI module).

shanedsnyder:
See https://github.com/argonne-lcf/autoperf/pull/17 for a fix in the Autoperf repo.

PR #1087, opened 2025-10-13
Check autoperf submodule git SHA to determine whether to use patched darshan-apmpi.c

wkliao:
Patched darshan-apmpi.c will only be used when submodule autoperf's HEAD SHA is equal or earlier than 8623a0652735785fe87d89514112e43e4f52a78d

related PR #1082 and #1083

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,11 +5,7 @@
 
 ACLOCAL_AMFLAGS = -I maint/config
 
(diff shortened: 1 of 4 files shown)


wkliao:
It appears that @shanedsnyder has found the same issue  in #964 and
provided a different solution to autoperf's APMPI.

PR #1098, opened 2025-11-01
add papi library to DYN_LD_FLAGS

wkliao:
papi library is required when configured with `--enable-apxc-mod`.
This PR add papi's library to flag `--dyn-ld-flags` of `darshan-runtime/darshan-config.in`


wkliao:
Because APXC module is meant for Cray XC systems only and the Darshan
developer team no long has an access to one of such systems, the support of
APXC module will be discarded in the next release (papi is a dependency of APXC module).

PR #1101, opened 2025-11-07
CI: use the latest official release of LDMS 4.5.1

wkliao:
Previously, the LDMS's main branch is used.


wkliao:
FYI. I consulted with Sara Walton at Sandia Lab. and she agreed with the use of an official version of LDMS.

PR #1102, opened 2025-11-10
New feature: collect Pthread IDs into DXT records

wkliao:
This PR enables Darshan to collect pthread IDs (type of an unsigned long) into DXT records.

See #1066 for more discussion. Note that #1066 also mentions an extra annotation, in addition to pthread ID, but this PR only adds pthread IDs, without the extra annotation. Adding the extra annotations will be in a separate PR.

An example of text output from darshan-dxt-parser is shown below.
It has "Pthread-ID" added as the right-most column.
```
# Module    Rank  Wt/Rd  Segment          Offset          Length    Start(s)      End(s)     Pthread-ID
 X_POSIX       2  write        0              16               8      0.0022      0.0022     140498826330816
```
For older versions:
```
# Module    Rank  Wt/Rd  Segment          Offset          Length    Start(s)      End(s)   [OST]     Pthread-ID
 X_POSIX       0  write        0            2048          262144      0.0300      0.0331   [106]     -1

```

### DIFF
--- a/darshan-util/darshan-dxt-logutils.c
+++ b/darshan-util/darshan-dxt-logutils.c
@@ -342,6 +342,42 @@ static void dxt_log_print_mpiio_file_darshan(void *file_rec, char *file_name,
 {
 }
 
+static int ulong_comp(const void *p1, const void *p2)
+{
+    if (*((unsigned long *)p1) > *((unsigned long *)p2))
+        return (1);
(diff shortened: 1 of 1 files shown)


wkliao:
Hi, @GueroudjiAmal 
Thanks for your feedback. We will use the pthread ID returned by `pthread_self()`.

PR #1158, opened 2026-08-07
New feature: use log file path hierarchy

wkliao:
* Applicable to non-MPI jobs only
* Add environment variable DARSHAN_LOGPATH_HIERARCHY. If it is set no matter its value, then this feature is enabled.
* Instead of a flat log file name, this feature uses the following file name hierarchy: USER/PROGRAMNAME/JOBID/HOSTNAME/PID-hashed_name.darshan
* Just a reminder, the default (i.e. when this feature is disabled), the log file path has the form of USER_PROGRAMNAME_JOBID-PID-hashed_name.darshan

### DIFF
--- a/darshan-runtime/lib/darshan-core.c
+++ b/darshan-runtime/lib/darshan-core.c
@@ -1509,6 +1509,55 @@ static void darshan_get_shared_records(struct darshan_core_runtime *core,
 }
 #endif
 
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <errno.h>
+
+static int mkdir_p(const char *base, const char *path, mode_t mode)
+{
+    char *tmp, *p = NULL;
+    int err=0;
+    size_t len;
+
+    len = strlen(base) + strlen(path) + 1;
+    tmp = (char*) malloc(len + 1);
+    snprintf(tmp, len+1, "%s/%s", base, path);
+    tmp[len+1] = '\0';
+
+    /* Remove trailing slash if present */
+    while (tmp[len] == '/') {
+        tmp[len] = '\0';
+        len--;
+    }
+
+    p = tmp + strlen(base) + 2;
+    for (; *p; p++) {
+        if (*p == '/') {
+            *p = 0;
+            err = mkdir(tmp, mode);
+            if (err != 0 && errno != EEXIST) {
+                /* Fail if error other than "already exists" */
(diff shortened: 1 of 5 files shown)


hariteja-jajula:
Tested PR #1158 on Polaris using a 2-node non-MPI job (`7446808`) with 8 independent processes, 4 per node.

The job ran on these two compute nodes:

* `x3208c0s37b1n0.hsn.cm.polaris.alcf.anl.gov`
* `x3208c0s7b0n0.hsn.cm.polaris.alcf.anl.gov` 

With `DARSHAN_LOGPATH_HIERARCHY=1` enabled, Darshan created logs in the expected hierarchical format, for example:

`logs/2026/8/13/hjajula/nonmpi_test/id7446808/x3208c0s37b1n0/<pid>-*.darshan`

and

`logs/2026/8/13/hjajula/nonmpi_test/id7446808/x3208c0s7b0n0/<pid>-*.darshan`

I also ran a separate 2-node MPI sanity test with 8 ranks, and normal MPI logging still produced the expected single aggregated Darshan log.

darshan-runtime/lib/darshan-core.c
```diff
@@ -1593,15 +1642,58 @@ static void darshan_get_logfile_name(
...
+                char hirerarchy_dir[__DARSHAN_PATH_MAX-1];
+
+                if (hname[0] == 0)
+                    (void)gethostname(hname, sizeof(hname));
+
+                dot = strchr(hname, '.');
+                if (dot != NULL)
+                    *dot = '\0'; /* Truncate at the first dot */
+
+                /* folder hierarchy: USER/PROGRAMNAME/JOBID/HOSTNAME/log_file */
```
roblatham00, line 1660: oh... It took me a moment to reason it out but you have HOSTNAME so far down the hierarchy because you're imagining a workflow that's running over multiple nodes.   I initially thought the workflow had lots of processes running on a single node. 

yeah it's fine here:  if it's one node, then it's just one directory level and presumably there are only so many jobs you can run from one workflow on one node.   If it's multiple nodes then the logs are spread out as intended.

roblatham00 (dismissed):

darshan-runtime/lib/darshan-core.c
wkliao: The naming hierarchy also contain job ID. So, the same application running on the same node, but with different job IDs (running at different time) can also help "dilute" the Darshan log files.

roblatham00 (approved):

